PR #13, opened 2026-04-20
fix: update readme & fix build

Imod7:
### Description
Update README to reflect the current state of the repository and added some fixes so that `pnpm build` completes with no errors.


Imod7:
Closing this one in favor of recent commits

PR #20, opened 2026-04-21
feat(host): add host API wrapper for @novasamatech packages

TarikGul:


product-sdk/packages/host/src/truapi.ts
pgherveou, line 1: should we name that truapi instead (even though it depends on @novasamatech/host-api)

product-sdk/packages/host/src/host-api.ts
```diff
@@ -0,0 +1,283 @@
...
+        log.debug("host API unavailable (not in container or SDK not installed)");
+        return null;
+    }
+}
+
+/**
+ * Inject the Spektr wallet extension into `window.injectedWeb3`.
+ *
+ * This makes the host's wallet appear as a browser extension, allowing
+ * compatibility with existing dApps that use the injectedWeb3 API.
```
pgherveou: DQ do we want that? I thought that Utkarsh yesterday mentioned that our concern should only be apps that live in the host

pgherveou (approved):

Imod7 (approved):
I tested with mark3t (replaced 2 packages -crypto and address-) and so far it works, the build completes and the related tests are passing.

filvecchiato (approved):
Overall this looks good! I would also be in favour of switching the naming of hostApi to TrUAPI

product-sdk/packages/host/src/host-api.ts
TarikGul: Yea I actually agree, the original thought is just to have everything exposed here, but in this case its actually not needed based on our requirements. I'll remove it.

TarikGul: Removed

product-sdk/packages/host/src/truapi.ts
TarikGul: Yea lets do it - I'll change the exports as well. I think we should keep the package as host though, but we can rename host-api as truapi

TarikGul: fixed.

eugypalu (approved):
Tested with t3rminal, the build and tests are working

PR #36, opened 2026-04-27
ci: deploy docs to Bulletin on @parity/product-sdk releases

valentinfernandez1:
Adds a workflow that regenerates, builds, and deploys the API docs to Bulletin (product-sdk.dot) on every @parity/product-sdk umbrella release. 

Workflow is triggered after [product-sdk: Release](https://github.com/paritytech/product-sdk/blob/main/.github/workflows/product-sdk-release.yml) workflow.

### DIFF
--- a/.github/workflows/docs-deploy.yml
+++ b/.github/workflows/docs-deploy.yml
@@ -25,7 +25,7 @@ jobs:
             - name: Check for umbrella tag at HEAD
               id: gate
               run: |
-                  tag=$(git tag --points-at HEAD | grep -E '^@parity/product-sdk@[0-9]' | head -n1 || true)
+                  tag=$(git tag --points-at HEAD | grep -E '^@parity/product-sdk@[0-9]+\.[0-9]+\.[0-9]+$' | head -n1 || true)
                   if [ -z "$tag" ]; then
                     echo "skip=true" >> $GITHUB_OUTPUT
                   else
@@ -75,6 +75,4 @@ jobs:
             - name: Deploy to Bulletin Chain
               if: steps.gate.outputs.skip != 'true'
               working-directory: docs
-              env:
-                  DEPLOY_TAG: ${{ steps.gate.outputs.tag }}
               run: npx -y bulletin-deploy ./out product-sdk.dot
--- a/.github/workflows/docs-deploy.yml
+++ b/.github/workflows/docs-deploy.yml
@@ -22,9 +22,6 @@ jobs:
               with:
                   fetch-depth: 0
 
-            # Release runs even when no changesets were consumed (no-op path)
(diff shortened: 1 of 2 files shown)


.github/workflows/docs-deploy.yml
```diff
@@ -0,0 +1,80 @@
...
+              if: steps.gate.outputs.skip != 'true'
+              working-directory: docs
+              run: pnpm docs:generate
+
+            - name: Build static site
+              if: steps.gate.outputs.skip != 'true'
+              working-directory: docs
+              run: pnpm build
+
+            - name: Deploy to Bulletin Chain
```
TarikGul, line 75: Aren't we missing the BULLETIN_MNEMONIC here?

.github/workflows/docs-deploy.yml
```diff
@@ -0,0 +1,80 @@
...
+              if: steps.gate.outputs.skip != 'true'
+              working-directory: docs
+              env:
+                  DEPLOY_TAG: ${{ steps.gate.outputs.tag }}
```
TarikGul: What is the purpose of the DEPLOY_TAG

.github/workflows/docs-deploy.yml
```diff
@@ -0,0 +1,80 @@
...
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v4
+              with:
+                  fetch-depth: 0
+
+            - name: Check for umbrella tag at HEAD
+              id: gate
+              run: |
+                  tag=$(git tag --points-at HEAD | grep -E '^@parity/product-sdk@[0-9]' | head -n1 || true)
```
TarikGul: Little nit but we might want to consider tighter regex: `grep -E '^@parity/product-sdk@[0-9]+\.[0-9]+\.[0-9]+'` this way it applies to just valid semver versioning.

TarikGul (commented):
Looks good overall, just some nits and questions, also might be good to have some type of failure notification.

.github/workflows/docs-deploy.yml
valentinfernandez1: The Mnemonic is used for the DotNS owner, however for now we are just deploying to paseo so there is no need to add the MNEMONIC as the `bulletin-deploy` CLI handles it directly on paseo.

.github/workflows/docs-deploy.yml
valentinfernandez1: Good catch I had actually missread the [bulletin-deploy readme](https://github.com/paritytech/bulletin-deploy#tagging-test-and-benchmark-runs). 

I though that providing the tag allowed it to appear on their dashboards but its actually the opposite.

TarikGul (approved):
🚀

PR #38, opened 2026-04-27
docs: write @packageDocumentation and TSDoc across SDK packages

valentinfernandez1:
## Summary

- Adds `@packageDocumentation` blocks to the 11 packages that were missing one, so the API reference overview table reads as peers rather than a mix of described and undescribed packages.
- Documents type aliases and interfaces that were previously missing descriptions across `address`, `contracts`, `statement-store`, `storage`, `signer`, `logger`, `crypto`, and `host`.
- Tightens the docs generator: function pages now only render the **Parameters** and **Returns** sections when there's a description to show, and parameters render as a bullet list (`- name: description`) instead of a table that mostly duplicated the signature line.

### DIFF
--- a/docs/scripts/lib/render-symbol.ts
+++ b/docs/scripts/lib/render-symbol.ts
@@ -40,16 +40,15 @@ function collapseWhitespace(s: string): string {
   return s.replace(/\s+/g, " ").trim();
 }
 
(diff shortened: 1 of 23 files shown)


TarikGul (approved):
👍

PR #39, opened 2026-04-29
feat(bench): add bundle-size benchmark + CI gate

TarikGul:
## Summary
                                                                                                                                                                                                                                                    
  Adds a bundle-size benchmark that runs on every PR touching `product-sdk/` and posts a sticky comment with the size impact per package.                                                                                                           
                                                                                                                                                                                                                                                    
  For each `@parity/product-sdk-*` package and every umbrella subpath, we measure:                                                                                                                                                                  
  - **Ship**: raw + gzip + brotli of the dist file (what npm sends).
  - **Bundled**: esbuild bundle of `import * as M` (consumer ceiling).                                                                                                                                                                              
  - **Shaken**: esbuild bundle of one named import (tree-shaken cost).                                                                                                                                                                              
                                                                                                                                                                                                                                                    
  The shaken/bundled ratio surfaces tree-shake regressions directly. The first run already flags `@parity/product-sdk-storage` at 100% retention and a few others under 80% — follow-up work.                                                       
                                                                                                                                                                                                                                                    
  ## What's gated                                                                                                                                                                                                                                   
                                                            
  CI compares PR head vs base and fails the job on >20% or >15KB bundled growth. Warns at >10% / >5KB. Deps-dominated entries (>100KB baseline) gate on percentage only.

### DIFF
--- a/product-sdk/BUNDLE_SIZE.md
+++ b/product-sdk/BUNDLE_SIZE.md
@@ -51,6 +51,8 @@ pnpm bench:compare
 
 The committed `bundle-size.json` represents the SDK's current measured size. **Update it** when you intentionally land a change that grows or shrinks bundles — adding a feature, swapping a dep, ripping out dead code.
 
+Regenerate using **Node 24** to match CI — V8/zlib output can shift slightly across Node majors, so a baseline captured on a different version will show phantom drift in PR comparisons run on `main`.
+
 ```bash
 pnpm build
 pnpm bench
--- a/product-sdk/BUNDLE_SIZE.md
+++ b/product-sdk/BUNDLE_SIZE.md
@@ -1,0 +1,97 @@
+# Bundle Size
+
+Every PR that touches `product-sdk/**` runs the bundle-size benchmark and posts a sticky comment with the impact per package. The goal is to keep the SDK small and to catch tree-shaking regressions early.
+
+## What gets measured
+
+For every `@parity/product-sdk-*` package and every subpath export of `@parity/product-sdk` and `@parity/product-sdk-descriptors`, three numbers are recorded:
+
+| Metric | How it's measured | What it tells you |
+|---|---|---|
+| **Ship** | Raw + gzip + brotli of `dist/<entry>.js` | What npm sends per file |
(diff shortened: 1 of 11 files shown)


TarikGul:
One con is this does a check per push, might be some improvements to make in that sense

filvecchiato (approved):
🚀

product-sdk/bundle-size.json
```diff
@@ -0,0 +1,657 @@
+{
+  "version": 1,
+  "generatedAt": "2026-04-29T15:31:32.236Z",
+  "node": "v24.11.0",
```
Imod7, line 4: Maybe its worth regenerating this file using the same node version as the CI? In the `product-sdk-bundle-size.yml` it uses node 22.

product-sdk/BUNDLE_SIZE.md
```diff
@@ -0,0 +1,97 @@
...
+| 🔴 fail | ≥20% or ≥15 KB | Job fails |
+
+For deps-dominated entries (>100 KB baseline), only the percentage applies — a +5 KB swing in a 1 MB bundle is noise. For small entries (<100 KB), absolute byte budgets kick in to catch slow drift.
+
+## Local commands
+
+```bash
+# Measure once and write bundle-size.json + bundle-size.md
+pnpm bench
+
```
Imod7, line 41: If the above comment is accepted then it would be nice to update this line so that anyone who runs `pnpm bench`, they run `nvm use 22` (node version 22 is also aligned with the rest of the repo also) so that the json file is aligned with the CI.

Imod7 (approved):

product-sdk/bundle-size.json
TarikGul: This is a good point instead I'll bump the CI to do latest, to keep things from being static.

TarikGul: Actually not having it static could be bad so lets move it to node 24

TarikGul: Fixed

product-sdk/BUNDLE_SIZE.md
TarikGul: fixed

PR #40, opened 2026-04-30
feat(bulletin): wrap @parity/bulletin-sdk for chunking + chain verification 

TarikGul:
## Wrap `@parity/bulletin-sdk` in `@parity/product-sdk-bulletin` for chunking, DAG-PB, and on-chain verification                                                                                                                                  
                                                                                                                                                                                                                                                    
  Replace the hand-rolled upload machinery in `@parity/product-sdk-bulletin` with a thin wrapper around upstream `@parity/bulletin-sdk@^0.3.0`. Consumers get native chunking (>2 MiB files), DAG-PB manifests, progress events, and a fluent       
  store-builder API for free. Adds two new helpers — `createLazySigner` (signer-on-demand for clients constructed before account selection) and `verifyOnChain` (CID-on-chain verification via `TransactionStorage.Transactions`).
                                                                                                                                                                                                                                                    
  Reads stay container-only and route through the host's preimage subscription, matching PR #26's stance — `@parity/product-sdk-bulletin` does not fetch from public IPFS gateways.                                                                 
  
  `pnpm-lock.yaml` and `bundle-size.json` look noisy in the diff because of the merged-in novasama 0.7 / PAPI 2.x bump that already landed on main — that work is not part of this PR's review surface.                                             
                                                            
  ## What changes for consumers of `@parity/product-sdk-bulletin`                                                                                                                                                                                   
                                                            
  | Before | After |                                                                                                                                                                                                                                
  | --- | --- |                                             
  | `BulletinClient.create("paseo")` | `BulletinClient.create({ environment: "paseo", signer })` — signer required |
  | `BulletinClient.from(api)` | `BulletinClient.from(inner, api)` — pass an `AsyncBulletinClient` |                                                                                                                                                
  | `bulletin.upload(data, signer?)` | `await bulletin.store(data).send()` (fluent builder) |                                                                                                                                                       
  | `bulletin.batchUpload([...])` | Loop `for (const item of items) await bulletin.store(item.data).send()` |                                                                                                                                       
  | `result.kind === "preimage" \| "transaction"` | `result: StoreResult` from upstream (`{ cid?, size, blockNumber?, extrinsicIndex?, chunks? }`) |                                                                                                
  | `computeCid(data)` (sync) | `await calculateCid(data)` (async — Web Crypto) |                                                                                                                                                                   
                                                                                                                                                                                                                                                    
  For an app consumer of just `@parity/product-sdk`'s headline API (`createApp`, `app.bulletin.upload(data)`, `app.bulletin.fetch(cid)`), nothing in source code needs to change — the umbrella facade is unchanged. The breakage above lands on    
  consumers using `@parity/product-sdk-bulletin` directly.                                                                                                                                                                                          
                                                                                                                                                                                                                                                    
  ## What's new                                                                                                                                                                                                                                     
  
  ### `BulletinClient` (rewritten)                                                                                                                                                                                                                  
                                                            
  Wraps `AsyncBulletinClient` from `@parity/bulletin-sdk`. Mirrors upstream's fluent builders (`.store(data).withChunkSize(...).withCallback(...).send()`, `.authorizeAccount`, `.renew`, `.estimateAuthorization`) and adds our own `fetchBytes`,  
  `fetchJson`, `checkAuthorization`, `verifyOnChain`, `destroy`. Two construction shapes: `BulletinClient.create({ environment, signer })` (env shorthand using `getChainAPI`) or `BulletinClient.create({ genesisHash, descriptor, signer })` (BYOD
   form).                                                                                                                                                                                                                                           
                                                            
  ### `createLazySigner(getter)` — `lazy-signer.ts` (new)                                                                                                                                                                                           
  
  `AsyncBulletinClient` takes a fixed `PolkadotSigner` at construction. Apps usually build the bulletin client at startup, before any account is selected. `createLazySigner(() => signerManager.getSigner())` returns a `PolkadotSigner` whose     
  `publicKey` / `signTx` / `signBytes` resolve through the getter on every call — picks up account changes, throws clearly when no signer is available. `createApp` uses this so `app.bulletin` works end-to-end without forcing consumers into a
  "connect before construct" dance.                                                                                                                                                                                                                 
                                                            
  ### `verifyOnChain(api, cid, { block, index? })` — `verify.ts` (new)

  The bulletin chain stores metadata only (`{ chunk_root, content_hash, hashing, cid_codec, size, block_chunks }`), not bytes. The original issue's "chain-storage byte read" path doesn't exist. What does exist is on-chain verification: parse   
  the CID's digest + multihash code, look it up in `TransactionStorage.Transactions[block]`, match against `content_hash` + `hashing`. Returns `{ block, index, size, blockChunks }` on match, `null` otherwise. Common pattern: pass `blockNumber`
  from a `store(...).send()` receipt to confirm the upload landed where expected.                                                                                                                                                                   
                                                            
  Replaces the previous `readFromChainStorage` placeholder that was misleadingly named — it would have always returned `null` because the chain doesn't expose CID-keyed bytes.

  ### `BulletinChain.paseo` — `networks.ts` (new)                                                                                                                                                                                                   
   
  Network preset bundling genesis hash + PAPI descriptor. Easy to extend when polkadot/kusama bulletin networks go live.                                                                                                                            
                                                            
  ### `query.ts` (rewritten)                                                                                                                                                                                                                        
                                                            
  Reads route through the host's preimage subscription via `resolveQueryStrategy()` → `getPreimageManager().lookup()`. Container-only by design (matching PR #26): outside a Polkadot Browser / Desktop host, `fetchBytes` throws                   
  `BulletinHostUnavailableError` rather than silently falling back to a public gateway. DAG-PB chunked content is reassembled transparently — `executeQuery` parses the manifest with `UnixFsDagBuilder`, fetches each child chunk through the host
  in parallel, and concatenates in manifest order. Pass `{ noReassemble: true }` to inspect the raw manifest bytes instead.                                                                                                                         
                                                            
  ### `cid.ts` (slimmed)                                                                                                                                                                                                                            
   
  Kept the two genuinely-useful helpers from the old code: `hashToCid(hexHash, hashCode?, codec?)` and `cidToPreimageKey(cidString)`. `computeCid` is gone — use upstream's `calculateCid` (async, returns a `CID` object). Multi-hash / multi-codec
   support stays for converting on-chain hashes back to CIDs.
                                                                                                                                                                                                                                                    
  ### `errors.ts` (rewritten)                               

  Two error families coexist: upstream `BulletinError` + `ErrorCode` (re-exported) cover upload/store/authorization failures with `code` / `retryable` / `recoveryHint`; a new `ProductBulletinError` base class covers our read-side errors —      
  `BulletinHostUnavailableError`, `BulletinLookupTimeoutError`, `BulletinLookupInterruptedError`, `BulletinCidError`, `BulletinAuthorizationError`. Catch upstream errors with `instanceof BulletinError`, ours with `instanceof
  ProductBulletinError`.                                                                                                                                                                                                                            
                                                            
  ## Risk

  Medium. The public API of `@parity/product-sdk-bulletin` has structural breaking changes (`BulletinClient` construction, upload/store call shape). For app consumers using `@parity/product-sdk`'s `createApp` facade, no source-level migration  
  is needed. For consumers using the bulletin package directly, the migration is the table at the top — straightforward but real.
                                                                                                                                                                                                                                                    
  The host-preimage e2e tests in `examples/bulletin-demo` are currently `describe.skip` because the test harness needs to be updated to seed preimages through the host's subscription-based `preimageManager` (the wrapped SDK now fetches via     
  `getPreimageManager().lookup`, not the older synchronous test seam). Worth a follow-up to rebuild the harness so we have green-flag confirmation for the new path.
                                                                                                                                                                                                                                                    
  ## Follow-ups (not in this PR)                                                                                                                                                                                                                    
   
  - Rebuild `examples/bulletin-demo` e2e harness against the subscription-based `preimageManager`. The current skips are explicit placeholders.                                                                                                     
  - Extend the demo with chunking + progress + `verifyOnChain` showcases (currently exercises only the basic upload+fetch path).
  - `skills/product-sdk-bulletin/{SKILL,references/bulletin-api}.md` predate this PR's upstream-SDK migration entirely — they describe an old API surface (`upload(data, signer?)`, `cidExists`, standalone `fetchBytes`/`fetchJson`, etc.) that no 
  longer matches the package. Needs a full rewrite, separate PR.                                                                                                                                                                                    
  - `@polkadot-labs/hdkd-helpers` is no longer pulled by anything in our tree — could be removed from devDeps in a cleanup pass.


TarikGul:
rel: https://github.com/paritytech/polkadot-bulletin-chain/pull/472

product-sdk/examples/bulletin-demo/e2e/boot.spec.ts
```diff
@@ -4,18 +4,13 @@ import { waitForAppReady } from "./helpers";
...
- *   - resolveUploadStrategy() → preimage kind
- *   - resolveQueryStrategy() → host-lookup kind
+ * SKIPPED: asserts on `upload-strategy` / `query-strategy` UI fields and the
+ * "preimage"/"host-lookup" log lines that came from the host-preimage code
+ * path. After the migration to AsyncBulletinClient those DOM nodes and code
+ * paths are gone. A trimmed boot smoke test (signer connects + bulletin
+ * client constructs) can be reinstated alongside the rest of the suite.
  */
-test.describe("@parity/product-sdk-bulletin via Host API — boot", () => {
+test.describe.skip("@parity/product-sdk-bulletin via Host API — boot", () => {
```
pgherveou: should we just 🔪  these tests or is there a plan to un-skip it later ?

product-sdk/packages/bulletin/src/cid.ts
```diff
@@ -133,6 +69,9 @@ export function hashToCid(
                 `got ${hexHash.length} chars`,
         );
     }
+    if (!/^0x[0-9a-fA-F]{64}$/.test(hexHash)) {
```
valentinfernandez1, line 72: Small nit but this could also fail if the hex is too short/long so the error is not necessarily correct, I would do something like

```typescript
throw new BulletinCidError(
        `Invalid hash format: expected 0x-prefixed 32-byte hex string, got: ${hexHash}`
    );
```

product-sdk/packages/bulletin/src/lazy-signer.ts
```diff
@@ -0,0 +1,90 @@
...
+    };
+
+    return {
+        get publicKey(): Uint8Array {
+            return resolve().publicKey;
+        },
+        signTx: async (...args) => resolve().signTx(...args),
+        signBytes: async (data) => resolve().signBytes(data),
+    } as PolkadotSigner;
+}
```
valentinfernandez1: Would avoid casting here instead I think this is cleaner

```suggestion
const lazy: PolkadotSigner = {
    get publicKey() {
        return resolve().publicKey;
    },
    signTx: (...args: Parameters<PolkadotSigner["signTx"]>) =>
        resolve().signTx(...args),
    signBytes: (...args: Parameters<PolkadotSigner["signBytes"]>) =>
        resolve().signBytes(...args),
};
return lazy;
```

product-sdk/packages/bulletin/src/lazy-signer.ts
```diff
@@ -0,0 +1,90 @@
...
+            expect(() => lazy.publicKey).toThrow("No signer available");
+        });
+
+        test("publicKey resolves through getter when signer is available", () => {
+            const inner = makeMockSigner("alice");
+            const lazy = createLazySigner(() => inner);
+            expect(lazy.publicKey).toBe(inner.publicKey);
+        });
+
+        test("signTx throws when getter returns null", async () => {
```
valentinfernandez1, line 68: There is no test for signTx happy path

product-sdk/packages/bulletin/src/client.ts
```diff
@@ -1,255 +1,291 @@
...
     }
 
-    /**
-     * Upload multiple items sequentially.
-     *
-     * @param items  - Array of items to upload, each with data and a label.
-     * @param signer - Optional signer. When omitted, auto-resolved.
-     * @param options - Batch upload options (timeout, progress callback).
-     */
-    async batchUpload(
```
valentinfernandez1, line 114: With this new implementation what would be the alternative method for batchUpload? I'm probably missing something.

valentinfernandez1 (commented):
Looks good overall just a few comments and nitpicks

product-sdk/packages/bulletin/src/cid.ts
TarikGul: Fixed.

product-sdk/packages/bulletin/src/query.ts
```diff
@@ -1,82 +1,64 @@
...
-import { resolveQueryStrategy, type QueryStrategy } from "./resolve-query.js";
-import type { QueryOptions } from "./types.js";
+import { BulletinGatewayUnavailableError } from "./errors.js";
+import { fetchBytes as fetchFromGateway } from "./gateway.js";
+import type { BulletinApi, QueryOptions } from "./types.js";
 
 const log = createLogger("bulletin");
 
 /**
- * Fetch raw bytes for a CID using the host preimage lookup.
```
Imod7, line 9: tiny q: I see the host-preimage path was removed on purpose. Is upstream @parity/bulletin-sdk expected to add it back later, or is gateway-only the new permanent model?

product-sdk/packages/bulletin/src/lazy-signer.ts
TarikGul: Nice suggestion, fixed.

product-sdk/packages/bulletin/src/lazy-signer.ts
TarikGul: Added

product-sdk/packages/bulletin/src/client.ts
TarikGul: The migration is: loop over `.store(data).send()` calls, since upstream `@parity/bulletin-sdk` doesn't provide a batch helper. The bulletin chain requires sequential transaction submission for nonce ordering anyway, so this is what batchUpload was internally, a for loop with progress reporting.

TarikGul: Might be worth adding it back as a public helper function

valentinfernandez1 (approved):
🚀

Imod7 (approved):

Imod7:
Also maybe worth adding the renaming of `computeCid` to `calculateCid` in the changes table?

product-sdk/packages/bulletin/src/query.ts
TarikGul: My honest opinion actually is remove the gateway and go back to host preimage retrieval.

The gateway is a bit hacky.

filvecchiato (approved):

product-sdk/examples/bulletin-demo/e2e/boot.spec.ts
TarikGul: Yea removed it :)

PR #45, opened 2026-05-05
chore(deps): bump novasama 0.6→0.7 + polkadot-api 1.x→2.x

TarikGul:
## Bump novasama 0.6 → 0.7 + polkadot-api 1.x → 2.x
                                                                                                                                                                                                                                                    
  Aligns the workspace with the latest published `triangle-js-sdks` release line. This was scoped as a routine version bump in `pnpm-workspace.yaml`, but novasama 0.7 crosses the `polkadot-api 1.x → 2.x` boundary *and* includes a structural    
  rewrite of `@novasamatech/sdk-statement`'s subscription API, so the change ended up touching three of our packages.                                                                                                                               
                                                                                                                                                                                                                                                    
  ## Version changes                                        

  | Package | Before | After |
  | --- | --- | --- |
  | `polkadot-api` (catalog) | `^1.23.3` | `^2.0.2` |                                                                                                                                                                                               
  | `@novasamatech/product-sdk` (catalog) | `^0.6.17` | `^0.7.5` |
  | `@novasamatech/sdk-statement` (catalog) | `^0.5.0` | `^0.6.0` |                                                                                                                                                                                 
  | `@novasamatech/host-api` (catalog) | `^0.7.0` | `^0.7.5` |                                                                                                                                                                                      
                                                                                                                                                                                                                                                    
  Eight per-package `package.json`s also switched their `polkadot-api: ^1.9.x` direct pins to `polkadot-api: catalog:` so the catalog bump actually propagates — without this, two `polkadot-api` versions would coexist in the tree.               
                                                            
  ## Breaking changes consumers will see                                                                                                                                                                                                            
                                                            
  ### `@parity/product-sdk-host`                                                                                                                                                                                                                    
                                                            
  - **`HostStatementStore.subscribe` signature changed.** Was `subscribe(topics: Uint8Array[], callback: (statements: unknown[]) => void)`, now `subscribe(filter: StatementTopicFilter, callback: (page: StatementsPage) => void)`. Callers pass a 
  structured filter (`{ matchAll: Topic[] } | { matchAny: Topic[] }`) and the callback receives pages of statements (`{ statements, isComplete }`) instead of raw arrays.
  - **`StatementProof` variants renamed and added to.** Was `Sr25519 | Ed25519 | Secp256k1Ecdsa | EcdsaRecoverable`, now `Sr25519 | Ed25519 | Ecdsa | OnChain`. The `Ecdsa` variant replaces `Secp256k1Ecdsa`; `EcdsaRecoverable` is gone; `OnChain`
   is new (chain-attestation-based proof referencing a block + event).                                                                                                                                                                              
  - **New exported types.** `StatementTopicFilter`, `StatementsPage`, `HostSubscription` are now exported from `@parity/product-sdk-host`.
  - **`JsonRpcProvider` import location.** Internally moved from `polkadot-api/ws-provider/web` (subpath gone in PAPI 2.x) to top-level `polkadot-api`. No external impact unless consumers import the same path themselves — in which case they    
  should update to `import type { JsonRpcProvider } from "polkadot-api"`.                                                                                                                                                                           
                                                                                                                                                                                                                                                    
  ### `@parity/product-sdk-statement-store`                                                                                                                                                                                                         
                                                            
  - **Subscription delivery is now page-based.** The `HostTransport.subscribe` callback fires once per page from the host, not per batch. Inside the transport this is hidden — `StatementTransport.subscribe`'s consumer callback signature        
  `(statements: Statement[]) => void` is unchanged — but the per-fire batch sizes may differ from before.
  - **No more `Secp256k1Ecdsa` / `EcdsaRecoverable` proofs.** If your code branches on proof variants, you'll need to handle `Ecdsa` and `OnChain` instead. Statements signed with the old variants are no longer producible by the upstream SDK.   
                                                                                                                                                                                                                                                    
  ### `@parity/product-sdk-bulletin`
                                                                                                                                                                                                                                                    
  - **`Binary.fromBytes` no longer needed when calling `tx.TransactionStorage.store`.** PAPI 2.x's typed `tx` accepts `Uint8Array` directly. We dropped the wrapper internally; if external code did the same, those `Binary.fromBytes(...)` calls  
  become no-ops at best and broken imports at worst (the function was removed from the `Binary` namespace in PAPI 2.x — it's now `{ toText, toHex, toOpaque, fromText, fromHex, fromOpaque }`, no `fromBytes`).
                                                                                                                                                                                                                                                    
  ### Workspace-wide (PAPI 2.x)                             

  - **`polkadot-api/ws-provider/web` and `/node` subpaths are gone.** PAPI 2.x consolidates these into `polkadot-api/ws`. Imports targeting the old subpaths break with `Cannot find module`.                                                       
  - **`Binary` namespace shape changed** as noted above. Most consumers were using `Binary.fromBytes()` which was removed.
  - **PAPI internal types changed.** `JsonRpcProvider`'s `onMessage` callback now receives `JsonRpcMessage<any>` instead of `string`, and `isResponse`/`isRequest` are now exported from `@polkadot-api/json-rpc-provider`. Consumer impact is      
  limited because we re-export `JsonRpcProvider` from `polkadot-api` directly.                                                                                                                                                                      
                                                                                                                                                                                                                                                    
  ## Tricky bit: the `json-rpc-provider` override                                                                                                                                                                                                   
                                                            
  `@polkadot-api/json-rpc-provider-proxy@0.4.0` (transitively pulled in via novasama 0.7.5 → polkadot-api 2.0.2) imports `isResponse` from `@polkadot-api/json-rpc-provider` — but it declares the dependency as a `devDependency`, not a regular   
  one. Without that runtime declaration, pnpm hoists whatever happens to be in the tree, which lands on `0.0.1` (pulled in by `@substrate/connect@0.8.11` → `@polkadot/rpc-provider` → ...) — a version that pre-dates `isResponse`. Vite's bundler
  then chokes when it actually tries to resolve the import.                                                                                                                                                                                         
                                                            
  The fix is a `pnpm.overrides` block in the root `package.json` pinning `@polkadot-api/json-rpc-provider` to `^0.2.0` everywhere. This forces dedup against the new version. If/when the upstream proxy package fixes its packaging metadata, the override can be removed.

### DIFF
--- a/product-sdk/bundle-size.json
+++ b/product-sdk/bundle-size.json
@@ -1,6 +1,6 @@
 {
   "version": 1,
-  "generatedAt": "2026-04-29T15:31:32.236Z",
+  "generatedAt": "2026-05-05T18:09:53.544Z",
   "node": "v24.11.0",
   "esbuild": "0.25.12",
   "packages": {
@@ -14,16 +14,16 @@
             "brotli": 278
           },
           "bundled": {
-            "raw": 4476259,
-            "gzip": 1468511,
-            "brotli": 675888
+            "raw": 4456890,
+            "gzip": 1467733,
+            "brotli": 675165
           },
           "shaken": {
-            "raw": 4346714,
-            "gzip": 1426165,
-            "brotli": 647087
+            "raw": 4338195,
+            "gzip": 1426844,
+            "brotli": 645741
           },
-          "shakeRatio": 0.9711
+          "shakeRatio": 0.9734
         },
         "./address": {
           "file": "packages/sdk/dist/address/index.js",
@@ -52,14 +52,14 @@
             "brotli": 91
           },
           "bundled": {
-            "raw": 4346183,
-            "gzip": 1427968,
-            "brotli": 649911
+            "raw": 4339527,
+            "gzip": 1426396,
+            "brotli": 647347
           },
           "shaken": {
             "raw": 12668,
             "gzip": 4739,
-            "brotli": 4336
+            "brotli": 4334
           },
           "shakeRatio": 0.0029
         },
@@ -71,16 +71,16 @@
             "brotli": 90
           },
           "bundled": {
-            "raw": 4254377,
-            "gzip": 1395798,
-            "brotli": 621908
+            "raw": 4247935,
+            "gzip": 1394295,
+            "brotli": 620754
           },
           "shaken": {
-            "raw": 218474,
-            "gzip": 73966,
-            "brotli": 62627
+            "raw": 213452,
+            "gzip": 73219,
+            "brotli": 64198
           },
-          "shakeRatio": 0.0514
+          "shakeRatio": 0.0502
         },
         "./contracts": {
           "file": "packages/sdk/dist/contracts/index.js",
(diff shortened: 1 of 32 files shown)


filvecchiato (approved):
Looks good!

PR #47, opened 2026-05-06
docs: keep host URL params on client-side navigation

valentinfernandez1:
## Summary

Fixes "Invalid sandbox URL — Missing required URL param `contentBackend`" when navigating between sections of the deployed docs site on dot.li

image
When the docs site is deployed to dot.li, every internal `<Link>` click eventually causes the sandbox to fail with:

> Invalid sandbox URL — Missing required URL param `contentBackend`. The host did not specify a backend — reload from dot.li.

It happens reliably after a few in-app navigations, or immediately on any reload that follows an internal click.

## Why

dot.li loads each app inside an iframe at `<cid>.app.dot.li` and the sandbox bootstrap validates required URL params (`contentBackend`, `chainBackend`, `skipArchiveCache`, ...) on every page load. From dot.li's bootstrap:

```ts
// The sandbox lives on cid.app.dot.li and cannot read the host's
// localStorage, so every user-chosen axis MUST arrive via URL param.
const urlParams = new URL(window.location.href).searchParams;
const parsed = validateSandboxParams(urlParams);
if (!parsed.ok) {
  showError("Invalid sandbox URL", parsed.reason);
  return;
}
```

Next.js App Router (with `output: "export"`) does soft navigation via `history.pushState`, which rewrites the path but drops the query string. The address bar goes from `…?contentBackend=…&chainBackend=…` to just `/api/sdk/`. Any subsequent reload (manual, browser history navigation, dev tools refresh, anchor that triggers a bootstrap, etc.) re-runs the validator against a now-paramless URL and fails.

This is a contract every sandbox-hosted app has to satisfy. It is not specific to Next.js. Any app using client-side routing (Vite + React Router, Vue Router, SvelteKit, etc.) hits the same wall once deployed to dot.li.

### DIFF
--- a/docs/app/_components/preserve-host-params.tsx
+++ b/docs/app/_components/preserve-host-params.tsx
@@ -1,0 +1,61 @@
+"use client";
+
+import { Suspense, useEffect } from "react";
+import { usePathname, useSearchParams } from "next/navigation";
+
+// dot.li (and similar sandbox hosts) load each app inside an iframe at
+// `<cid>.app.dot.li` and re-validate URL params (`contentBackend`,
+// `chainBackend`, `skipArchiveCache`, ...) on every page load. Next.js
+// soft navigation rewrites the URL via history.pushState and drops query
+// params by default, which trips the host validator the next time anything
+// re-bootstraps the iframe.
+//
+// Strategy: on first navigation effect after mount, capture whatever
+// params the host put in the URL. After every subsequent navigation,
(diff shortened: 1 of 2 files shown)


valentinfernandez1:
Will open an issue on dotli as this is a product side fix but every product could encounter this issue eventually to a fix should be implemented on the host side.

TarikGul (approved):
👍

PR #49, opened 2026-05-06
feat(bulletin): add top-level authorizeAccount helper

Imod7:
### Description
- Adds top-level authorizeAccount(api, who, transactions, bytes, signer, options?) to @parity/product-sdk-bulletin, mirroring the call shape of the existing top-level upload(api, data, signer, options?).
- Supports both observed real-world patterns via a single viaSudo?: boolean option - direct call (mark3t, hackm3) and Sudo.sudo(...)-wrapped (sourc3s, sh33ts, task-rabbit).
- Pairs with the existing checkAuthorization for a "check, grant if insufficient, then upload" flow that no longer requires consumers to drop to raw api.tx.TransactionStorage.authorize_account(...).

### Why
Surveyed 9 Parity products. **5 of them independently reimplement `authorize_account`** in their own codebases — ~80 LOC of duplicated chain-extrinsic plumbing across separate repos, each one a place where a future runtime change to the call signature would silently break.

| Product | Pattern | LOC saved |
|---|---|---|
| mark3t | Direct (no sudo) | ~10 |
| sourc3s | Sudo-wrapped | ~30 |
| sh33ts | Sudo-wrapped | ~10 |
| task-rabbit | Sudo-wrapped, parallel | ~10 |
| hackm3 | Direct (manual encoding) | ~20 |

### API

```ts
export interface AuthorizeAccountOptions {
    viaSudo?: boolean;       // default false; wrap in Sudo.sudo(...)
    waitFor?: WaitFor;       // "best-block" | "finalized"
    timeoutMs?: number;
    onStatus?: (s: TxStatus) => void;
}

// Top-level — mirrors upload(api, data, signer, options?)
export async function authorizeAccount(
    api: BulletinApi,
    who: string,
    transactions: number,
    bytes: bigint,
    signer: PolkadotSigner,
    options?: AuthorizeAccountOptions,
): Promise<{ blockHash: string }>;
```
Users who already hold a BulletinClient get a builder-based client.authorizeAccount(who, transactions, bytes): AuthCallBuilder from main's separate addition (delegates to upstream's AsyncBulletinClient); the top-level helper here is the path with viaSudo support, additive-quota safety, and a one-shot {blockHash } return.

### Behavior decisions worth surfacing

**Sudo-pallet guard**

If `viaSudo: true` is requested against a chain without Sudo (production Polkadot/Kusama removed it via governance), throw `ProductBulletinError` before submission with a clear message. Instead of letting consumers see a low-level `TypeError: Cannot read properties of undefined` (reading 'sudo').

**No `withRetry` and why**

The sibling upload helper wraps `submitAndWatch` in `withRetry`. We deliberately do not here.

Verified polkadot-bulletin-chain/pallets/transaction-storage/src/lib.rs:
```
AuthorizationScope::Account(_) => {
    authorization.extent.bytes_allowance = authorization.extent.bytes_allowance
        .saturating_add(bytes_allowance);   // ← saturating_add, NOT overwrite
    authorization.extent.transactions_allowance = authorization.extent.transactions_allowance
        .saturating_add(transactions_allowance);
},
```

`authorize_account` is additive, not idempotent. A silent retry after a successful-but-lost-acknowledgment submission would double-grant quota. Callers needing retry can wrap this helper and verify post-state via `checkAuthorization` first - they then explicitly own the additive risk.

JSDoc carries a prominent ## Additive semantics warning so future maintainers don't reintroduce the wrapper without thinking through the consequence.

### Test plan
 - pnpm -F @parity/product-sdk-bulletin test → 70/70 passing
   - authorization.ts (13 total: 7 existing + 6 new)
       - Direct path: correct call params, no Sudo invocation, returns block hash
    - Sudo path: wraps in Sudo.sudo({ call: decodedCall }), returns block hash from sudo extrinsic
    - Sudo-missing guard: throws ProductBulletinError, does not proceed to submit
- pnpm -F @parity/product-sdk-bulletin build → ESM 18.16 KB, DTS 25.26 KB
- pnpm check (Biome lint + format) → clean
- pnpm -r --workspace-concurrency=1 --filter "./packages/*" build → all packages green

Submission-error propagation is intentionally not re-tested here - owned by @parity/product-sdk-tx's suite.

### Merge note

Originally proposed both a top-level helper and a BulletinClient.authorizeAccount class-method delegator. After merging main, which introduced its own builder-based `BulletinClient.authorizeAccount(...)`: `AuthCallBuilder` (a thin delegator over upstream's AsyncBulletinClient, matching the file's established pattern alongside store, authorizePreimage, renew), our class-method addition became a redundant duplicate. Dropped it; kept the top-level helper.

The 5 surveyed beneficiary products all reach for the top-level form.

### Out of scope / follow-ups

- Tightening BulletinApi from `TypedApi<any>` to a typed descriptor: would remove the as any cast used here and in upload.ts/checkAuthorization. Workspace-wide change,
separate effort.
- Broader SDK convention question (functional-first vs. dual-form for new helpers): main's class method follows the upstream-builder pattern; this PR's top-level helper follows the functional shape. Worth opening as a separate discussion if the SDK team wants a decision that applies to future helpers.


filvecchiato (approved):
looks good!

PR #55, opened 2026-05-07
refactor: Docs auto discover packages from the workspace

valentinfernandez1:
While working on the storage refactor in #54 I noticed that the current way that we where generating the docs with fixed paths was not ideal.

This PR future proofs the process by allowing the docs to auto discover the available packages allowing the generator to pick up on package name changes or new packages.


Imod7 (approved):
LGTM!

PR #58, opened 2026-05-07
feat: contracts pvm loader

### LINKED ISSUE #57: Add cargo-pvm-contract support and remove ink! from @parity/product-sdk-contracts
## Problem

`@parity/product-sdk-contracts` wraps `@polkadot-api/sdk-ink` and consumes ABIs through CDM-generated `cdm.json`. Two gaps:

- Contracts authored with `cargo-pvm-contract` (which emits `target/<name>.release.{abi.json,polkavm}`) cannot be consumed without going through CDM.
- ink! is no longer a target for the product-sdk, but the package's public API (`InkSdk`, `InkContract`, `createInkSdk` examples in skills/JSDoc) and runtime (`sdk-ink`) still surface it.

### Artefact loader — `@parity/product-sdk-contracts/pvm`

New subpath export.

- `parsePvmContractAbi(unknown): AbiEntry[]` — sync, browser-safe; accepts parsed JSON, JSON string, `Uint8Array`, or `{ abi }` wrapper.
- `loadPvmContractAbi(path): Promise<AbiEntry[]>` — async, Node-only; lazy-imports `node:fs/promises`.
- `loadPvmContractArtifacts(basePath): Promise<{ abi, bytecode }>` — reads both artefacts.

Output drops into existing `createContract` / `createContractFromClient` / `ContractManager` unchanged — the emitted ABI is already Solidity-compatible and consumed end-to-end by the existing codegen.

### Runtime swap — replace `sdk-ink` with pallet-revive-direct

`wrap.ts` consumes exactly two sdk-ink methods: `inkContract.query()` and `inkContract.send()`. Both replaced by a pallet-revive-aware encoder calling `ReviveApi.call` (dry-run) and `Revive.call` (extrinsic) directly.

- **Solidity codec via `viem`.** `encodeFunctionData` / `decodeFunctionResult` from `viem` handle selector + head/tail encoding. `viem` is already a transitive dependency of `@polkadot-api/sdk-ink`, so the effective bundle delta is removing the sdk-ink wrapper + the `@polkadot-api/ink-contracts` peer (~4 MB ink metadata) while keeping the codec piece that was already in the tree. Promotes `viem` to a direct dependency at the catalog version.
- `@polkadot-api/sdk-ink` removed from `packages/contracts/package.json` and the `pnpm-workspace.yaml` catalog.
- Address handling reuses `@parity/product-sdk-address` (`deriveH160`, `ss58ToH160`, `h160ToSs58`) — no new utilities needed. CDM addresses are already H160; SS58 origin → H160 mapping uses the existing `0xee` padding helper.

### Symbol rename — drop ink! from public API (hard cut)

`InkSdk` → `ContractRuntime`, `InkContract` → `ContractHandle`, `inkSdk` field/parameter → `runtime`. Sweep `createInkSdk` JSDoc in `chain-client/src/{types,presets,clients}.ts` and `tx/src/account-mapping.ts`. Sweep `"ink!"` / `"Ink!/PolkaVM"` → `"PolkaVM/Solidity"` in skills + reference docs. No deprecation aliases.

### Codegen

Extend `generateContractTypes` to accept artefact paths in addition to `{ library, abi }`.

## Out of scope

- Replacing CDM. CDM stays as the build/deploy/registry orchestrator.
- Writing a hand-rolled Solidity ABI codec. Considered and rejected: `viem` is already in the transitive tree via sdk-ink, so adopting it directly is bundle-neutral-to-positive and avoids correctness risk on selector/head-tail encoding under deadline.

## Acceptance

- An example consumes `cargo pvm-contract build` artefacts end-to-end against Asset Hub Paseo without importing `@polkadot-api/sdk-ink` from product code.
- `grep -rE 'ink!|InkSdk|createInkSdk' packages/ skills/` returns zero hits (excluding `CHANGELOG.md`).
- `@parity/product-sdk-contracts` main-entry bundle size drops by approximately the size of the removed sdk-ink runtime + `@polkadot-api/ink-contracts` peer.

decrypto21:
Closes #57.
cc: @UtkarshBhardwaj007 

This PR adds support for `cargo-pvm-contract` build artefacts to `@parity/product-sdk-contracts` and removes the `@polkadot-api/sdk-ink` dependency entirely. The package now drives `pallet-revive` directly via PAPI typed APIs and uses `viem` for the Solidity ABI codec.

- **Pre-change**: contracts package was bundled around ink! / `pallet-contracts`, pulling ~750 KB gzip of `@polkadot-api/sdk-ink` runtime + ink! metadata into every consumer build.
- **Post-change**: zero ink! references in product code; consumer bundle drops to ~73 KB gzip (~90 % reduction); contracts can be loaded straight from `cargo pvm-contract build` output without a `cdm.json` manifest.

## Root cause

`pallet-contracts` (ink!) is being phased out on Asset Hub in favour of `pallet-revive` (PolkaVM / Solidity-compatible). The new toolchain -`cargo-pvm-contract` - emits Solidity-style ABI JSON + a `.polkavm` blob, not ink! metadata. `@polkadot-api/sdk-ink`:

1. cannot consume `cargo-pvm-contract` artefacts,
2. ships ink!-specific runtime that consumers no longer need, and
3. anchors us to a deprecated pallet.

## Changes made
1. **`feat(contracts): add cargo-pvm-contract artifact loader`** - new `./pvm` subpath exporting `loadPvmContractAbi`, `loadPvmContractCode`, `loadPvmContractArtifacts`, `parsePvmContractAbi`. Browser-safe (sync) + Node-only (async, lazy `node:fs` import) variants.
2. **`feat(contracts): add pvm-contracts example, fs tests, and skill docs`** - `examples/pvm-contracts-example/` with a real `counter.release.abi.json` fixture; fs round-trip tests; updated `skills/product-sdk-contracts/`.
3. **`feat(contracts): swap sdk-ink for pallet-revive direct + viem codec; rename ink symbols`** - the core swap. Drops `@polkadot-api/sdk-ink`; introduces `createContractRuntime(typedApi)` + `ContractRuntime`; `ReviveApi.call` for dry-runs, `Revive.call` for extrinsics; `viem.encodeFunctionData / decodeFunctionResult` for codec. Sweeps every `ink!` / `InkSdk` / `createInkSdk` reference out of `packages/`, `examples/`, and `skills/`.
4. **`feat(contracts): codegen artefact-path support`** -`generateContractTypes` now accepts `{ library, abi }` *or* `{ library, abiPath }` via new `resolveContractTypeInputs()` helper, so consumers can codegen straight from `cargo-pvm-contract` build output.
5. **`feat(contracts): pre-flight dry-run failure + account-mapping helper`** - `.tx()` path now throws `ContractDryRunFailedError` if the pre-flight `ReviveApi.call` reports failure (no more paying gas on chain-rejected txs). New `ensureContractAccountMapped(runtime, address, signer)` helper for the one-time `Revive.map_account()` prerequisite.
6. **`feat(contracts): route ReviveApi.call dry-run through unsafe API to bypass compat-token drift`** - first attempt at fixing the e2e compat-token failure. Added createContractRuntimeFromClient as the PolkadotClient-aware factory, but the actual routing fix turned out to be incomplete (see commit 7).
7. **`fix(contracts): drop class-based Binary/FixedSizeBinary; pass HexString + Uint8Array to PAPI 2.x codecs`** - root cause was that substrate-bindings 0.12's class API (Binary, FixedSizeBinary) isn't recognised by PAPI 2.x's compat check. The contracts package was the only holdout that hadn't migrated when polkadot-api was bumped 1.x→2.x. Removed @polkadot-api/substrate-bindings from the package's dependencies; types now use HexString / Uint8Array throughout, matching what PAPI 2.x codecs actually accept.
8. **`fix(contracts): map Solidity bytes/bytesN to HexString/SizedHex`** - codegen output aligned with the new runtime contract. Integration tests pin PAPI's expected arg shape (`0x` `dest`, `Uint8Array` calldata) so a future regression to class-based wrappers fails loudly in unit tests instead of silently in CI.

9. **`test(contracts-demo): live-chain query e2e tests`** - `getReportCount` (uint256), `getAllDates` (string[]), `getCID` (string) against the deployed t3rminal contract on Paseo Asset Hub. Read-only, deterministic (zero-address shop), no signing, no chain-state dependency.

10. **`fix(contracts): address bundle bench, loadPvmContractCode, account-mapping wiring, 0.3.0 bump`** 

11. **`feat(contracts): port .prepare() onto the revive runtime`** -  `.prepare()` was ink-based. Re-implemented against the revive runtime via a shared `buildReviveCall` helper that .tx() and .prepare() both compose: encode calldata → conditionally dry-run to size weight + storage → build `Revive.call` extrinsic. `.prepare()` returns the `SubmittableTransaction` directly (consumable as a `BatchableCall` by `batchSubmitAndWatch` since PAPI submittables carry `.decodedCall`). Five new unit tests pin: both-overrides-skips-dry-run, partial-overrides-still-dry-run, no-signer-required, dry-run-failure-before-extrinsic, end-to-end batch round-trip via `batchSubmitAndWatch`.

12. **`chore(bench): floor percentage rule at 10 KB baseline; tolerate base-bench errors`** - Two adjustments to `scripts/bench-bundle.mjs`: the 20% percentage gate now only applies above a 10 KB baseline (below that, absolute byte budgets govern alone - a single new helper on a 1.5 KB subpath is +50% by construction); and "base errored, head succeeded" entries surface as `now measurable` informational rows rather than reading the `null → N KB` jump as a regression.

13. **`fix(contracts): assemble multi-output return values into the named object shape codegen advertises`** - Review fix. viem hands back `[T1, T2]` for multi-output methods while `generateMethodResponseType` advertises `{n1: T1; n2: T2}`. `decodeReturn` now assembles the named object from viem's array using the ABI output names, falling back to `_0`, `_1`, … to match codegen's naming policy. Three new unit tests pin all four output shapes (single, multi-named, multi-unnamed, tuple-passthrough). Encode side double-checked across six shapes (multi-arg primitive, single tuple, mixed positional+tuple, nested tuples, tuple arrays, bytesN) - viem accepts both `{...}` and `[...]` for tuple slots at every depth, so no encode fix needed.

14. **`docs(contracts): scrub README ink references; fix chain-client raw-client examples; drop changeset`** - Review fixes. Contracts README scrubbed of every ink mention; `ContractManager.fromClient` doc table corrected (sync factory, new `descriptor` param). Three JSDoc examples in `chain-client/{clients,presets,types}.ts` fixed to use `createContractRuntimeFromClient(client.raw.assetHub, paseo_asset_hub)` instead of `createContractRuntime(client.assetHub)` - comment said "Raw client", now actually uses it. Changeset removed.

## Public API diff

```diff
- import { createInkSdk, type InkSdk } from "@polkadot-api/sdk-ink";
- const inkSdk = createInkSdk(typedApi);

+ // Production: build a runtime straight from a raw PolkadotClient.
+ // ContractManager.fromClient handles typed-API + unsafe-API wiring under
+ // the hood, and getRuntime() exposes it for helpers that need a runtime
+ import {
+     ContractManager,
+     ensureContractAccountMapped,
+     ContractDryRunFailedError,
+ } from "@parity/product-sdk-contracts";
+ const manager = ContractManager.fromClient(cdm, client, paseo_asset_hub, { signerManager });
+ await ensureContractAccountMapped(manager.getRuntime(), address, signer); // once at boot

  // Codegen output for Solidity bytes / bytesN aligns with PAPI 2.x:
- // bytes   → Binary
- // bytesN  → FixedSizeBinary<N>
+ // bytes   → HexString
+ // bytesN  → SizedHex<N>
+ // Re-run `cdm install` after upgrading to regenerate user-facing types.

  // Node-only loaders now live behind a subpath so the main entry stays
  // browser-safe (no fs/path/os imports leaking into the runtime bundle):
- import { loadPvmContractAbi } from "@parity/product-sdk-contracts";
+ import {
+     parsePvmContractAbi,
+     loadPvmContractAbi,
+     loadPvmContractCode,
+     loadPvmContractArtifacts,
+ } from "@parity/product-sdk-contracts/pvm";
```

## Breaking changes
- ContractManager.fromClient(...) is now synchronous. Was async on main (lazy-imported sdk-ink).
- ContractManager.fromClient(cdm, client, descriptor, options?) now requires a chain descriptor as the third argument (e.g. paseo_asset_hub).
- Exported types (ReviveCallTx, ReviveTypedApi, ReviveDryRunCall, ReviveDryRunResult) switched binary fields from class-based FixedSizeBinary<N> / Binary to HexString / Uint8Array. Code that compiled against the old shapes would have failed at runtime against PAPI 2.x's compat check; the new shapes match what PAPI 2.x actually accepts.
- Codegen output for Solidity bytes / bytesN changed to HexString / SizedHex<N>. Re-run cdm install after upgrading.
- Node-only artefact loaders live behind /pvm; build-time codegen lives behind /codegen. 
- createInkSdk / InkSdk are removed. Projects still on pallet-contracts / ink! must stay on contracts 0.2.x.

## Testing
681 unit tests passing across 14 packages, 62 test files. 
Contracts package: 87 tests / 5 files (up from 78 at the start of the PR - +9 net new across the boundary integration tests, error-path coverage, override-bypass behaviour, and pvm artefact round-trips).


product-sdk/packages/contracts/src/wrap.ts
```diff
@@ -104,41 +79,168 @@ function resolveSigner(
...
+        functionName: methodName,
+        args,
+    });
+}
+
+/**
+ * Decode a successful query's return data via the Solidity ABI codec.
+ * Returns `undefined` for void methods.
+ */
+function decodeReturn(abi: AbiEntry[], methodName: string, returnData: Uint8Array): unknown {
```
charlesHetterich, line 127: I believe there is a mismatch between this decode function & `contracts/src/codegen.ts`'s `generateMethodResponseType`.

In `generateMethodResponseType` we map a list of `[{name, type}, {name, type}, ...]` to `{name1: type1, name2: type2, ...}`, so this decode should follow the same format so that the data we get is in the expected shape.

I would double check if this pattern mismatch also exists on `encodeCalldata`

decrypto21: Thanks! Fixed in [f245742](https://github.com/paritytech/product-sdk/pull/58/commits/f245742db6f7747fd4998baf12a7530133ab68fe).

Decode: confirmed with viem. viem hands back [T1, T2] for multi-output methods while generateMethodResponseType advertises {n1: T1; n2: T2}. decodeReturn now assembles the named object from viem's array using the ABI output names, falling back to _0, _1, … to match generateMethodResponseType's naming policy.

Encode: double checked. No encode fix needed.

TarikGul:
Looks like the contracts README still has references to Ink which need to be cleaned up.

product-sdk/.changeset/contracts-pvm-loader.md
```diff
@@ -0,0 +1,23 @@
+---
```
TarikGul: We can remove this changeset since it will run the release on the PR being merged, I am going to create a better workflow soon so this doesnt cause issues in the future.

product-sdk/packages/chain-client/src/clients.ts
```diff
@@ -56,9 +56,9 @@ function configFingerprint(chains: Record<string, ChainDefinition>): string {
  * const account = await client.assetHub.query.System.Account.getValue(addr);
  * const fee = await client.bulletin.query.TransactionStorage.ByteFee.getValue();
  *
- * // Raw client for advanced use (e.g., InkSdk for contracts)
- * import { createInkSdk } from "@polkadot-api/sdk-ink";
- * const inkSdk = createInkSdk(client.raw.assetHub, { atBest: true });
+ * // Raw client for advanced use (e.g., a ContractRuntime for pallet-revive contracts)
+ * import { createContractRuntime } from "@parity/product-sdk-contracts";
+ * const runtime = createContractRuntime(client.assetHub);
```
TarikGul: The raw client would be client.raw.assetHub. That being said we mention Raw but instead use `client.assetHub`.

product-sdk/packages/chain-client/src/presets.ts
```diff
@@ -172,9 +172,9 @@ export type PresetChains<E extends Environment> = PresetDescriptors[E];
  * const account = await client.assetHub.query.System.Account.getValue(addr);
  * const fee = await client.bulletin.query.TransactionStorage.ByteFee.getValue();
  *
- * // Raw client for advanced use (e.g., InkSdk for contracts)
- * import { createInkSdk } from "@polkadot-api/sdk-ink";
- * const inkSdk = createInkSdk(client.raw.assetHub, { atBest: true });
+ * // Raw client for advanced use (e.g., a ContractRuntime for pallet-revive contracts)
+ * import { createContractRuntime } from "@parity/product-sdk-contracts";
+ * const runtime = createContractRuntime(client.assetHub);
```
TarikGul: Same issue with the client vs client.raw here.

product-sdk/packages/chain-client/src/types.ts
```diff
@@ -52,15 +52,15 @@ export interface ChainClientConfig<
  * // Typed API access — fully typed from your descriptors
  * const account = await client.assetHub.query.System.Account.getValue(addr);
  *
- * // Raw client for advanced use (e.g., InkSdk for contracts)
- * import { createInkSdk } from "@polkadot-api/sdk-ink";
- * const inkSdk = createInkSdk(client.raw.assetHub, { atBest: true });
+ * // Raw client for advanced use (e.g., a ContractRuntime for pallet-revive contracts)
+ * import { createContractRuntime } from "@parity/product-sdk-contracts";
+ * const runtime = createContractRuntime(client.assetHub);
```
TarikGul: Same thing with the raw client here as well.

product-sdk/.changeset/contracts-pvm-loader.md
decrypto21: Resolved with [1a22b87](https://github.com/paritytech/product-sdk/pull/58/commits/1a22b87148d10513a7731cfe94ceb37b5ca42f2f)

product-sdk/packages/chain-client/src/presets.ts
decrypto21: Resolved with [1a22b87](https://github.com/paritytech/product-sdk/pull/58/commits/1a22b87148d10513a7731cfe94ceb37b5ca42f2f)

product-sdk/packages/chain-client/src/types.ts
decrypto21: Resolved with [1a22b87](https://github.com/paritytech/product-sdk/pull/58/commits/1a22b87148d10513a7731cfe94ceb37b5ca42f2f)

product-sdk/packages/chain-client/src/clients.ts
decrypto21: Resolved with [1a22b87](https://github.com/paritytech/product-sdk/pull/58/commits/1a22b87148d10513a7731cfe94ceb37b5ca42f2f)

TarikGul (approved):

PR #60, opened 2026-05-08
docs: verify build on PRs

valentinfernandez1:
Adds a PR check that builds the docs, so we catch breakages before release instead of during deploy.


TarikGul (approved):
I dont mind for now it building the sdk in its own path, I think its okay for now since its really small.

I can follow this up with a fix in the terminal package to get it to build.

PR #62, opened 2026-05-08
fix(terminal): route signTx through session.signPayload to avoid BadProof

TarikGul:
# fix(terminal): route `signTx` through `session.signPayload` to avoid `BadProof`

  ## Summary                                                                                                                                                                                                                                        
   
  `@parity/product-sdk-terminal`'s session signer routed both transaction signing and raw-message signing through `session.signRaw`. That's the wallet's raw-message path, which always applies the `<Bytes>...</Bytes>` anti-phishing envelope     
  before signing — appropriate for arbitrary user data, but wrong for SCALE-encoded extrinsics. The chain expects a signature over the unwrapped extrinsic, so it rejected every transaction with `BadProof`.
                                                                                                                                                                                                                                                    
  The fix routes transaction signing through `session.signPayload` instead. The single-callback `getPolkadotSigner` is replaced with the two-callback `getPolkadotSignerFromPjs` to enable the split.                                               
   
  The implementation mirrors the proven pattern in `@novasamatech/product-sdk`'s in-host signer (`packages/product-sdk/src/accounts.ts`) — same adapter, same callback split, same hex normalization helper.                                                                                                                                                                                             
                       
  ## What changed                                                                                                                                                                                                                                   
                       
  | Before | After |
  | --- | --- |
  | `getPolkadotSigner(accountId, "Sr25519", oneCallback)` | `getPolkadotSignerFromPjs(addressHex, signPayload, signRaw)` |
  | Single callback, unconditional `session.signRaw` | Two callbacks; `signTx` → `session.signPayload`, `signBytes` → `session.signRaw` |                                                                                                           
  | Tx signing produced `BadProof` on every chain | Tx signing produces a valid signature; raw-byte signing keeps the `<Bytes>` wrap as intended |                                                                                                  
                                                                                                                                                                                                                                                    
  **Public API is unchanged.** `createSessionSigner(session, adapter)` and `createSessionSignerForAccount(session, ref)` keep their signatures. Internal routing only.                                                                              
                                                                                                                                                                                                                                                    
  ## Verification                                                                                                                                                                                                                                   
                                                            
  - **PAPI source** (`@polkadot-api/pjs-signer/from-pjs-account.ts`): confirms `signTx` reaches `signPayload` and `signBytes` reaches `signRaw`. The PJS adapter accepts a `0x`-prefixed hex AccountId; we pass `0x` + hex of                       
  `session.remoteAccount.accountId` so `publicKey` stays identical to what we exposed before.
  - **host-papp source** (`packages/host-papp/src/sso/sessionManager/userSession.ts`): confirms `signPayload` and `signRaw` route to different SCALE enum variants (`SignRequest('Payload', …)` vs `SignRequest('Raw', …)`), and only the `Raw`     
  variant gets dispatched to mobile's `<Bytes>`-wrapping interactor.                                                                                                                                                                                
  - **host-api protocol** (`docs/design/host-api-protocol.md`): two distinct RPCs (`host_sign_payload`, `host_sign_raw`), wrap applied only to the latter.
  - **Cross-checked against `@novasamatech/product-sdk`** in-host signer — same adapter, same routing, same `asHex`/`assetId` handling. Our implementation matches the working browser-side path.                                                   
                                                                                                                                                                                                                                                    
  **End-to-end tx signing roundtrip is NOT in CI** — no way to exercise a real phone. Manual smoke test (`packages/terminal/manual-tests/qr-pair-and-sign.mjs`) currently only covers `signBytes`. **Validation against a paseo chain is required   
  before publishing 0.1.1.** The changeset is parked in `local-docs/pending-changesets/` so this PR's merge does NOT trigger a release.                                                                                                             
                                                                                                                                                                                                                                                    
  ## Bundle size                                            

  `dist/index.js` grew 10.7 → 14.4 KB (+3.7 KB) — metadata-decoder helpers needed by `getPolkadotSignerFromPjs` for extrinsic version detection and signed-extension iteration. `polkadot-api/pjs-signer` itself is correctly externalized.

### DIFF
--- a/product-sdk/packages/terminal/src/signer.ts
+++ b/product-sdk/packages/terminal/src/signer.ts
@@ -51,14 +51,135 @@ export interface ProductAccountRef {
 }
 
 /**
- * PAPI's PJS payload exposes hex fields as plain `string` (the typedef is
- * `HexString = string` with no enforcement). The mappers always emit
- * `0x`-prefixed values, but we defensively prepend the prefix if missing —
- * matches the pattern in `@novasamatech/product-sdk`'s in-host signer.
+ * Defensively prepend `0x` if missing.
+ *
+ * PAPI's PJS payload types every hex field as `HexString = string` with no
+ * runtime check. In practice the signed-extension mappers always emit
+ * `0x`-prefixed values, so this should never actually prepend — but the
+ * type contract doesn't guarantee it, and host-papp's SCALE codec rejects
+ * unprefixed input. Mirrors the `asHex` helper in `@novasamatech/product-sdk`'s
+ * in-host signer for the same reason.
  */
 function asHex(v: string): `0x${string}` {
     return v.startsWith("0x") ? (v as `0x${string}`) : (`0x${v}` as `0x${string}`);
 }
+
+/**
+ * Build the `signPayload` callback PAPI calls for transaction signing.
+ *
+ * Translates PAPI's `SignerPayloadJSON` into host-papp's `SigningPayloadRequest`
+ * and routes it to `session.signPayload` — the mobile wallet's JSON-payload
+ * interactor, which signs the SCALE-encoded extrinsic directly with no
+ * `<Bytes>` envelope. This is the path that the chain accepts.
+ *
+ * Exported only via `import.meta.vitest` so unit tests can exercise the
+ * callback wiring directly without standing up a full PolkadotSigner.
+ */
+function makeSignPayloadCallback(session: UserSession, productAccountId: [string, number]) {
+    return async (
+        payload: PjsSignerPayloadJSON,
+    ): Promise<{
+        signature: `0x${string}`;
+        signedTransaction?: `0x${string}`;
+    }> => {
+        const result = await session.signPayload({
+            productAccountId,
+            blockHash: asHex(payload.blockHash),
+            blockNumber: asHex(payload.blockNumber),
+            era: asHex(payload.era),
+            genesisHash: asHex(payload.genesisHash),
+            method: asHex(payload.method),
+            nonce: asHex(payload.nonce),
+            specVersion: asHex(payload.specVersion),
+            tip: asHex(payload.tip),
+            transactionVersion: asHex(payload.transactionVersion),
+            signedExtensions: payload.signedExtensions,
+            version: payload.version,
+            // PJS types `assetId` as `number | object`. In practice the
+            // ChargeAssetTxPayment mapper always emits a hex string when
+            // present — we trust the mapper rather than runtime-checking,
+            // matching the reference in `@novasamatech/product-sdk`.
+            assetId:
+                payload.assetId !== undefined
+                    ? (payload.assetId as never as `0x${string}`)
+                    : undefined,
+            metadataHash: payload.metadataHash ? asHex(payload.metadataHash) : undefined,
+            mode: payload.mode,
+            withSignedTransaction: payload.withSignedTransaction,
+        });
+
+        if (result.isErr()) {
+            throw new Error(`Mobile signing rejected: ${result.error.message}`);
+        }
+
+        return {
+            signature: toHex(result.value.signature) as `0x${string}`,
+            signedTransaction: result.value.signedTransaction
+                ? (toHex(result.value.signedTransaction) as `0x${string}`)
+                : undefined,
+        };
+    };
+}
+
+/**
+ * Build the `signRaw` callback PAPI calls for arbitrary-byte signing
+ * (`signBytes`). Routes to `session.signRaw` — the mobile wallet's raw
+ * interactor, which applies the standard `<Bytes>...</Bytes>` anti-phishing
+ * envelope before signing. Correct for arbitrary user data.
+ *
+ * Exported only via `import.meta.vitest` for direct unit tests.
+ */
+function makeSignRawCallback(session: UserSession, productAccountId: [string, number]) {
+    return async (
+        payload: PjsSignRawPayload,
+    ): Promise<{
+        id: number;
+        signature: `0x${string}`;
+    }> => {
+        const result = await session.signRaw({
+            productAccountId,
+            data: { tag: "Bytes" as const, value: fromHex(payload.data) },
+        });
+
+        if (result.isErr()) {
+            throw new Error(`Mobile signing rejected: ${result.error.message}`);
+        }
+
+        return {
+            id: 0,
+            signature: toHex(result.value.signature) as `0x${string}`,
+        };
+    };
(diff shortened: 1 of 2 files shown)


pgherveou:
## Review summary

**Verdict:** Useful but Incomplete

The conceptual fix is right. PAPI's `getPolkadotSigner` collapses `signTx` and `signBytes` onto a single callback, and routing both through `session.signRaw` triggers the mobile `<Bytes>...</Bytes>` envelope on SCALE-encoded extrinsics, which the chain rightfully rejects as `BadProof`. Switching to `getPolkadotSignerFromPjs` to split the callbacks (tx → `session.signPayload`, raw → `session.signRaw`) is the correct shape, and the wire field names match the installed `@novasamatech/host-papp@0.7.7` codecs (`productAccountId: [string, number]`, not `address`).

What it doesn't do: actually unit-test the path it's fixing. The `tx signing path` `describe` block does not exercise `signPayload`. Combined with the author's own note that "End-to-end tx signing roundtrip is NOT in CI", the regression-critical path remains uncovered. A future PR could quietly revert to a unified callback and no test would fail.

## Findings

| # | Severity | File / Location | Description |
|---|----------|------------------|-------------|
| 1 | **High** | `packages/terminal/src/signer.ts` (≈L271–339, `tx signing path` describe) | The two tests in this block don't test the new `signPayload` integration. `signPayload callback returns hex signature in PJS shape` only asserts `signer.publicKey instanceof Uint8Array`. `signPayload routes to session.signPayload with productAccountId` actually calls `signer.signBytes(...)` and asserts `signPayload.mock.calls.length === 0` — i.e. it asserts signPayload is NOT called. The bug being fixed has no unit-test guard; the existing "regression guard for BadProof bug" test only proves `signBytes → signRaw` still works. |
| 2 | Medium | `packages/terminal/src/signer.ts` L311–338 | Test name `signPayload routes to session.signPayload with productAccountId` is misleading — the body asserts the opposite (signPayload not called). It duplicates the earlier `signBytes does NOT call session.signPayload` regression-guard test. Rename or replace with a real signPayload-callback test. |
| 3 | Medium | `packages/terminal/src/signer.ts` L291–309 | Test name `signPayload callback returns hex signature in PJS shape` doesn't match its body; the inline comment even admits the test doesn't actually verify the return shape. Either invoke the `signPayload` callback (e.g. via PAPI's `signTx` with stub metadata) or remove the test. |
| 4 | Medium | `packages/terminal/src/signer.ts` L97–100 | `assetId: (payload.assetId as unknown as \`0x${string}\`)` is an unsafe double-cast. `SignerPayloadJSON.assetId` is typed `number \| object`; if the `ChargeAssetTxPayment` mapper ever emits anything other than a `0x...` string, scale-ts will throw at encode time inside the SCALE codec. The cited reference in `@novasamatech/product-sdk` does not normalize — it spreads `{...payload}` and trusts the mapper. Prefer either dropping the cast (let TS validate) or guarding with a runtime hex check that throws a clear error. |
| 5 | Low | PR description / repo | PR description says "The changeset is parked in `local-docs/pending-changesets/`". `local-docs` is gitignored, so the file isn't visible to reviewers and won't be picked up by `changesets/cli`. That appears intentional (delays the release), but it should be documented in `RELEASES.md` / contributing docs so future maintainers don't unintentionally skip a real changeset step. |
| 6 | Info | `packages/terminal/README.md` L207 | Note references "Versions of this package prior to `0.1.1`". Currently published is `0.1.0`; the wording reads correctly only once `0.1.1` actually ships. Fine as-is if the parked-changeset workflow is intentional. |
| 7 | Info | `packages/terminal/src/signer.ts` L50–58 | `asHex` is essentially defensive padding; `@polkadot-api/utils#toHex` already returns `0x...` and PAPI's PJS inputs are already hex strings. Harmless, but a quick comment that this is a defensive no-op (or just dropping it) would simplify the read. |

## Cross-file consistency checks

- **UniFFI / host-mobile bindings:** N/A — only `packages/terminal` touched.
- **Parallel implementations:** PR mirrors `@novasamatech/product-sdk` `accounts.ts` in shape; the only divergence is the `assetId` cast (#4) and per-field `asHex` normalization (#7). Worth aligning on the reference's "trust the mapper" pattern.
- **Constants / wire shapes:** `productAccountId: [string, number]` matches the installed `@novasamatech/host-papp@0.7.7` `SigningPayloadRequestCodec` and `SigningRawRequestCodec` types. ✓

## Suggested changes before merge

1. Add a real test for the `signPayload` callback. Two options:
   - Refactor `buildSessionSigner` to expose the two callbacks (or a thin wrapper) for testability, and assert directly that calling the signPayload callback with a synthetic `SignerPayloadJSON` invokes `session.signPayload` with the right SCALE shape and hex-encodes the returned signature.
   - Or drive PAPI's `signTx` with a tiny v4-extrinsic metadata fixture (heavier, but pins the integration end-to-end without a real chain).
2. Rename or remove the two misleading tests in the `tx signing path` describe block.
3. Drop the `as unknown as \`0x${string}\`` cast on `assetId` and either pass it through (matching the reference) or runtime-check it.

Manual paseo validation before publishing `0.1.1` (already called out in the PR description) remains the right gate for the end-to-end happy path.

pgherveou (changes requested):
Conceptual fix is right, but the tests in the new tx signing path describe block don't exercise signPayload — see review comment. The central regression-prevention path is uncovered. Also flagging the assetId cast as a runtime risk.

product-sdk/packages/terminal/src/signer.ts
```diff
@@ -163,23 +275,83 @@ if (import.meta.vitest) {
...
+            // wires signRaw and signPayload independently — if signBytes
+            // hits signRaw, that's structurally enough to demonstrate
+            // that the two callbacks are routed to different host-papp
+            // methods. The full signTx path is covered by the manual smoke
+            // test that exercises against a real chain.
+            const session = makeSession({
+                signRaw: async () => ok({ signature: new Uint8Array([1]) }),
+            });
+            const signer = createSessionSigner(session, fakeAdapter("my-app"));
+            await signer.signBytes(new Uint8Array([1, 2, 3]));
```
UtkarshBhardwaj007: This is still using signBytes.

UtkarshBhardwaj007 (approved):
LGTM, one comment on tests.

TarikGul:
@pgherveou Addressed the comments above. 1, 2, 3, 4, 7 - 5, 6, wont be necessary to change.

product-sdk/packages/terminal/src/signer.ts
TarikGul: That test was old and removed now :)

PR #63, opened 2026-05-08
feat(contracts): add .prepare() for batching + cover cdm.json flow

TarikGul:
# fix(contracts): port `.prepare()` and cover the cdm.json flow                                                                                                                                                                                   
                                                                                                                                                                                                                                                  
  ## Summary                                                                                                                                                                                                                                        
   
  Two pieces flagged as missing from `@parity/product-sdk-contracts` vs the upstream source. After investigation:                                                                                                                                   
                                                            
  - **`.prepare()` was actually missing.** Ported from `polkadot-apps@4b60d19`.                                                                                                                                                                     
  - **`ContractManager.getContract("@org/name")` already worked** — but the package shipped without a README or any `manager.ts` tests, so the feature was effectively invisible. Added both.
                                                                                                                                                                                                                                                    
  ## Changes                                                
                                                                                                                                                                                                                                                    
  - **`.prepare()`** on every wrapped ABI method — returns a `BatchableCall` consumable by `batchSubmitAndWatch`. Doesn't require a signer; the batch's signer is the dispatched origin at submission. New `PrepareOptions` type. `BatchableCall`   
  re-exported from `@parity/product-sdk-tx`.
  - **8 `ContractManager` tests** in `manager.ts` covering construction from a real-world cdm.json, `getContract` resolution, `getAddress`, `ContractNotFoundError`, auto-target-selection, explicit `targetHash`, empty-targets error,             
  `setDefaults`. Closes a pre-existing test gap (manager.ts had zero tests in 0.2.0).                                                                                                                                                               
  - **`packages/contracts/README.md`** — new file. Primary `cdm.json` flow front-and-center, full `ContractManager` API, signer/origin resolution hierarchy, batching with `.prepare()`, manual `createContract` escape hatch, `cdm.json` schema
  reference, type generation, errors table.                                                                                                                                                                                                         
  - **`CdmJsonContract.metadataCid` made optional** — real-world cdm.json files often omit it, runtime never reads it.

### DIFF
--- a/product-sdk/packages/contracts/README.md
+++ b/product-sdk/packages/contracts/README.md
@@ -1,0 +1,224 @@
+# @parity/product-sdk-contracts
+
+Typed contract interactions on Polkadot Asset Hub. Resolve deployed contracts from a `cdm.json` manifest, get fully-typed handles for `query`, `tx`, and batched `prepare` calls — all backed by the ink SDK.
+
+## Install
+
+```bash
+pnpm add @parity/product-sdk-contracts
+```
+
+## Quick start (with cdm.json)
+
+The `cdm.json` flow is the primary path. A `cdm.json` manifest in your project root pins each contract to an address + ABI per target chain; `ContractManager.fromClient(cdm, api)` resolves them at runtime.
+
+```ts
+import { createChainClient } from "@parity/product-sdk-chain-client";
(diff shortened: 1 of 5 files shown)


charlesHetterich (approved):
LGTM

I wasn't able to test that transactions work e2e in desktop because of unrelated account/signing issues, but type resolution from cdm.json worked perfectly for me, and prepare looks good.

Issue #68, opened 2026-05-11
Support Previewnet (and Paseo Nextv2) as an environment


ottovlotto:
Paseo next v1 is being deprecated and we need to move to preview net and later to paseo next v2 (preview net is priority for now) -

- This needs to be supported as an environment in the chain-client
- (Optionally) need descriptors for this environment as well.
- Need RPCs for the previewnet in bulletin, statement store etc.

TarikGul:
This is all updated, closing.

PR #71, opened 2026-05-11
feat(chain-client): wire previewnet and split bulletin/individuality descriptors per environment 

TarikGul:
## Summary                                                

  Wires `previewnet` as a known preset environment in `@parity/product-sdk-chain-client` so consumers can call `getChainAPI("previewnet")` against the new zombienet deployment at `previewnet.substrate.dev` (Paseo runtime). Replaces Paseo Next  
  v1 as the recommended test target ahead of the triangle retreat presentation and Thursday's demo with Gav.
                                                                                                                                                                                                                                                    
  In the same wave, restructures `@parity/product-sdk-descriptors` so bulletin and individuality resolve per-environment — mirroring the asset-hub pattern — so `descriptor.genesis` matches the live chain instance the consumer connects to.      
  
  Refs #68.                                                                                                                                                                                                                                         
                                                            
  ## What's in this PR

  **`@parity/product-sdk-chain-client`**
  - `types.ts`: add `"previewnet"` to the protocol-level `Environment` union.
  - `presets.ts`:                                                                                                                                                                                                                                   
    - add `"previewnet"` to the preset `Environment` union and `AVAILABLE_ENVIRONMENTS`;
    - add `rpcs.previewnet` block with asset-hub, bulletin, and individuality endpoints;                                                                                                                                                            
    - restructure `loadDescriptors` so every chain (not just asset-hub) resolves per-environment via a single loader map;                                                                                                                           
    - replace `AssetHubDescriptors` with `PresetDescriptors`, a per-environment shape covering all three chains;                                                                                                                                    
    - tests assert distinct genesis hashes per chain per environment;                                                                                                                                                                               
  - `clients.ts`: correct misleading JSDoc that claimed a direct-WebSocket fallback — the SDK is container-only.                                                                                                                                    
                                                                                                                                                                                                                                                    
  **`@parity/product-sdk-descriptors`**                                                                                                                                                                                                             
  - New per-environment subpackages: `previewnet-asset-hub`, `paseo-bulletin`, `previewnet-bulletin`, `paseo-individuality`, `previewnet-individuality`. Each generated against the live chain with its own genesis hash and `.scale` metadata blob.
  - Drop the shared `bulletin` and `individuality` subpackages (replaced by the per-environment variants).                                                                                                                                          
  - `package.json` exports + `scripts/build.sh` chain list updated to match.
                                                                                                                                                                                                                                                    
  **`@parity/product-sdk-bulletin`**                        
  - `BulletinChain.previewnet` preset with the live previewnet bulletin genesis hash.                                                                                                                                                               
  - `BulletinChain` entries now reference per-environment descriptors instead of a single shared one; `BulletinNetwork.descriptor` widens from a single type to a union.                                                                            
  - Tests assert distinct descriptors and distinct genesis hashes across environments.                                                                                                                                                              
                                                                                                                                                                                                                                                    
  **`@parity/product-sdk-host`**                                                                                                                                                                                                                    
  - `chains.ts`: add `previewnet: ["wss://previewnet.substrate.dev/bulletin"]` to `BULLETIN_RPCS`; add corresponding test.                                                                                                                          
                                                                                                                                                                                                                                                    
  **Skill docs + JSDoc examples**
  - 7 skill files updated to surface previewnet in environment lists, warnings, and `Environment` type unions.                                                                                                                                      
  - BYOD examples migrated from the removed `/bulletin` and `/individuality` paths to the new per-environment paths (`paseo-bulletin`, `paseo-individuality`).                                                                                      
  - App-builder `chains.md` reference doc gains chain-property sections for previewnet asset-hub, bulletin, and individuality.                                                                                                                      
                                                                                                                                                                                                                                                    
  ## Side fix: stale paseo-individuality genesis                                                                                                                                                                                                    
                                                                                                                                                                                                                                                    
  When regenerating `paseo-individuality` against the live `wss://paseo-people-next-rpc.polkadot.io`, the genesis hash came back as `0xa22a2424...` — different from the `0xd01475...` baked into the removed shared `individuality` descriptor. The
   paseo people-next chain was redeployed (v1 → v2) at some point and the old descriptor had stale metadata. The new descriptor reflects the current live chain.
                                                                                                                                                                                                                                                    
  ## Breaking changes (BYOD path)                           

  Direct importers of the removed shared descriptors must migrate:                                                                                                                                                                                  
  
  - `@parity/product-sdk-descriptors/bulletin` → `@parity/product-sdk-descriptors/paseo-bulletin` (or `/previewnet-bulletin`)                                                                                                                       
  - `@parity/product-sdk-descriptors/individuality` → `@parity/product-sdk-descriptors/paseo-individuality` (or `/previewnet-individuality`)
  - Named exports rename: `bulletin` → `paseo_bulletin`, `individuality` → `paseo_individuality`, etc.                                                                                                                                              
                                                                                                                                                                                                                                                    
  Consumers using `getChainAPI(env)` are not affected at the call site — the typed return shape just becomes more precise. No internal callers needed migration; all workspace tests still pass.                                                    
                                                                                                                                                                                                                                                    
  ## Statement-store note                                                                                                                                                                                                                           
                                                            
  No SDK changes needed for previewnet's statement-store. `packages/statement-store/src/transport.ts` always routes through the host container via `getStatementStore()` (`packages/host/src/container.ts:97-104`), which uses the host's native    
  `remote_statement_store_*` binary protocol rather than picking its own WebSocket URL. The `wss://previewnet.substrate.dev/people` statement_* endpoint is configured in the mobile dev app builds, not here.
                                                                                                                                                                                                                                                    
  ## Endpoints wired                                        

  | Chain | URL |
  |---|---|
  | Asset Hub | `wss://previewnet.substrate.dev/asset-hub` |
  | Bulletin | `wss://previewnet.substrate.dev/bulletin` |                                                                                                                                                                                          
  | Individuality (People) | `wss://previewnet.substrate.dev/people` |


product-sdk/packages/chain-client/src/presets.ts
```diff
@@ -68,8 +75,11 @@ async function loadDescriptors(env: Environment) {
             ? ahMod.polkadot_asset_hub
             : "kusama_asset_hub" in ahMod
               ? ahMod.kusama_asset_hub
-              : (ahMod as typeof import("@parity/product-sdk-descriptors/paseo-asset-hub"))
-                    .paseo_asset_hub;
+              : "paseo_asset_hub" in ahMod
+                ? ahMod.paseo_asset_hub
+                : (
+                      ahMod as typeof import("@parity/product-sdk-descriptors/previewnet-asset-hub")
```
TarikGul: descriptors will need to go before this can be merged

product-sdk/packages/chain-client/src/presets.ts
```diff
@@ -199,6 +216,14 @@ if (import.meta.vitest) {
         expect(descriptors.individuality.genesis).toBe(GENESIS.individuality);
     });
 
+    test("loadDescriptors returns descriptors for previewnet", async () => {
+        const descriptors = await loadDescriptors("previewnet");
+        expect(descriptors).toBeDefined();
```
UtkarshBhardwaj007, line 281: Should have some genesis hash assertions

UtkarshBhardwaj007 (approved):
LGTM, needs follow ups as listed already in PR description but just some that come to mind - BulletinClient needs a `previewNet` config as well; Skills, readme etc. need updates

product-sdk/packages/chain-client/src/presets.ts
TarikGul: fixed, just needed the descriptors

charlesHetterich:
`BulletinChain` object in the `Bulletin` package should be updated to reference the new Preview Net environment (And check if there are other places that reference chain environments & need to be updated for the new environment)

TarikGul:
> `BulletinChain` object in the `Bulletin` package should be updated to reference the new Preview Net environment (And check if there are other places that reference chain environments & need to be updated for the new environment)

Updated

TarikGul:
This is ready to merge, we can ignore the failing test as its just the descriptors blowing up the bundle but its not actually included in the shipped bundle so its not a worry. Just running a quuick smoke test locally

product-sdk/packages/chain-client/src/presets.ts
```diff
@@ -8,14 +8,15 @@ import type { ChainClient } from "./types.js";
 import type { polkadot_asset_hub as PolkadotAssetHubDef } from "@parity/product-sdk-descriptors/polkadot-asset-hub";
 import type { kusama_asset_hub as KusamaAssetHubDef } from "@parity/product-sdk-descriptors/kusama-asset-hub";
 import type { paseo_asset_hub as PaseoAssetHubDef } from "@parity/product-sdk-descriptors/paseo-asset-hub";
+import type { previewnet_asset_hub as PreviewnetAssetHubDef } from "@parity/product-sdk-descriptors/previewnet-asset-hub";
 import type { bulletin as BulletinDef } from "@parity/product-sdk-descriptors/bulletin";
 import type { individuality as IndividualityDef } from "@parity/product-sdk-descriptors/individuality";
```
charlesHetterich: Do we need a similar separation for `preivew-*` / `paseo-*` for bulletin & individuality similar to what we have for asset hub?

Afaik, it seems that in the current setup `getChainAPI` will get the correct asset hub per environment, but will always give paseo's individuality/bulletin (including paseo's genesis hash for these chains)

charlesHetterich (approved):
LGTM outside of my last comment which still needs to be addressed.

In general bulletin & individuality should resolve in the same way, dynamically based on the given environment, Asset Hub currently resolves.

This would involve adding the correct preview net descriptors for bulletin & individuality, and then follow the same "wiring" scheme in code that Asset Hub follows, adding these genesis hash's to the test cases, etc.

product-sdk/packages/chain-client/src/presets.ts
TarikGul: Yea I am on board with the seperation. It allows for a more consistent architecture and will probably benefit us long term. I will update now.

TarikGul:
> LGTM outside of my last comment which still needs to be addressed.
> 
> In general bulletin & individuality should resolve in the same way, dynamically based on the given environment, Asset Hub currently resolves.
> 
> This would involve adding the correct preview net descriptors for bulletin & individuality, and then follow the same "wiring" scheme in code that Asset Hub follows, adding these genesis hash's to the test cases, etc.

Updated, its going to be a breaking change but I think its worth the future benefit.

PR #79, opened 2026-05-12
feat(contracts)!: expose value on failed query results

ReinhardHatko:
## Summary

`QueryResult` dropped the failure payload — callers got `{ success: false, value: undefined }` with no signal for why the dry-run failed. `QueryResult<T>` is now a discriminated union whose failure branch carries the dry-run's raw payload as `value: unknown` (typically a tagged enum like `{ type: "Reverted", value: ... }` or `{ type: "Trapped", ... }`), so callers can narrow on shape and react.

## What changed

- **`types.ts`**: `QueryResult<T>` is now a discriminated union:
  ```ts
  type QueryResult<T> =
      | { success: true; value: T; gasRequired: bigint }
      | { success: false; value: unknown; gasRequired?: bigint };
  ```
  Success branch gains a guaranteed `gasRequired: bigint` (was optional). Failure branch carries `value: unknown` with the raw dry-run payload.
- **`wrap.ts`**: pass `result.value` through on failure instead of replacing it with `undefined`.

## Breaking changes

Type-level only. **Runtime behavior on the success path is unchanged.**

| Pattern | Before | After |
|---|---|---|
| `r.value` without narrowing on `r.success` | Typed `T`; `undefined` at runtime on failure (silent bug) | Typed `T \| unknown` — TS flags the unnarrowed read |
| `if (r.success)` discrimination | No narrowing — `r.value` always `T` | True discriminated union — `r.value` narrows per branch |
| Constructing a `QueryResult<T>` literal | `gasRequired` optional everywhere | `gasRequired: bigint` required on the success branch |
| `declare module … { interface QueryResult … }` | Allowed (interface) | Blocked (type alias) |

### Migration

If your code reads `r.value` without first checking `if (r.success)`, add the narrowing. Code that was already narrowing keeps working unchanged.

```ts
// Before — compiled, but `r.value` was `undefined` at runtime on failure:
const r = await contract.query.foo();
processResponse(r.value);

// After:
const r = await contract.query.foo();
if (r.success) {
    processResponse(r.value);
} else {
    handleFailure(r.value); // typed as `unknown` — narrow on shape
}
```

## Version bump

`@parity/product-sdk-contracts` `0.2.2` → `0.3.0` (pre-1.0 minor; Changesets convention for breaking-but-pre-stable). Changeset included.

## Test plan

- [x] `pnpm --filter "@parity/product-sdk-contracts" build`
- [x] `pnpm --filter "@parity/product-sdk-contracts" test`
- [ ] CI green
- [ ] Smoke-test a known-failing query against a real chain to confirm the failure payload lands intact end-to-end


TarikGul:
In order to get this in, i did the smoke test and cleaned up the PR, we are good to go, I'll merge this then fix https://github.com/paritytech/product-sdk/issues/80 and get a release out.

TarikGul (approved):

PR #81, opened 2026-05-13
fix: terminal papi native signer

### LINKED ISSUE #80: product-sdk-terminal cannot sign Paseo v2 `Revive.map_account` because `AsPgas` is not representable through the PJS signing bridge
`@parity/product-sdk-terminal` currently builds its `PolkadotSigner` through `polkadot-api/pjs-signer` so PAPI transaction signing can be forwarded to `host-papp` / mobile as a Polkadot-JS-style `signPayload` request.

On Paseo Next v2, `Revive.map_account()` includes the `AsPgas` signed extension. `@polkadot-api/pjs-signer` does not know how to map `AsPgas` into a PJS payload, so signing fails locally in the CLI before the request reaches mobile.

This blocks `playground-cli dot init` from mapping the `playground.dot` product account on Asset Hub, which then blocks product-account-origin contract interactions.

## Reproduction

Using `playground-cli` after updating to the current product-sdk/CDM releases:

- `@parity/product-sdk-terminal@0.2.0`
- `@parity/product-sdk-contracts@0.4.0`
- `@parity/product-sdk-descriptors@0.4.0`
- `@parity/product-sdk-tx@0.2.2`
- `polkadot-api@2.1.2`
- `@polkadot-api/pjs-signer@0.7.2`
- `@novasamatech/host-papp@0.7.8-2`

Run:

```sh
dot init
```

Observed:

```txt
Attestation for 5FKsrh7LhhLDAZ46MathqA1foUprRqXDdGJpvcEhM9Jm7jbv successfully passed.

✓ logged in     5EtCcBcNx8gxq5XetBRcTBW1baH4J87pnuThJcbneJgDWRaL

account

✕ asset hub mapping approve on your Polkadot mobile app…
    PJS does not support this signed-extension: AsPgas

[tx] Transaction subscription error {
  error: "PJS does not support this signed-extension: AsPgas",
}
```

Important distinction: QR login and attestation succeed. The failure is not in personhood/session establishment. It happens in the later Asset Hub account-mapping transaction.

## Code Path

In `playground-cli`, `dot init` does:

1. Create a product account signer:

```ts
createSessionSignerForAccount(session, {
  productId: "playground.dot",
  derivationIndex: 0,
});
```

2. Check whether that product account is mapped on Asset Hub.

3. If unmapped, submit:

```ts
api.tx.Revive.map_account()
```

through `@parity/product-sdk-tx`'s `ensureAccountMapped(...)`.

The selected account/signing model appears correct: this is the `playground.dot` product account, not Alice and not a legacy account.

The failure occurs before the mobile app receives a signing request:

```js
// @parity/product-sdk-terminal/dist/index.js
return getPolkadotSignerFromPjs(
  accountIdHex,
  makeSignPayloadCallback(session, productAccountId),
  makeSignRawCallback(session, productAccountId)
);
```

Then:

```js
// @polkadot-api/pjs-signer/dist/from-pjs-account.js
if (!pjsSignedExtensionsMappers[identifier]) {
  if (signedExtension.value.length === 0 && signedExtension.additionalSigned.length === 0)
    return;
  throw new Error(`PJS does not support this signed-extension: ${identifier}`);
}
```

`pjs-signed-extensions-mappers.js` supports common extensions such as `CheckGenesis`, `CheckNonce`, `CheckMortality`, `CheckSpecVersion`, `CheckTxVersion`, `ChargeTransactionPayment`, `ChargeAssetTxPayment`, and `CheckMetadataHash`, but not `AsPgas`.

## Why This Matters

Paseo Next v2 / the current Product chain runtime uses PGAS-related signed extensions. Product accounts need to submit Asset Hub transactions such as `Revive.map_account()` before they can interact with contracts.

Because the terminal signer bridges PAPI signing into a PJS-shaped payload, arbitrary/current runtime signed extensions cannot pass through unless the PJS bridge explicitly understands them.

## Expected Behavior

`createSessionSignerForAccount(...)` should be able to sign Paseo v2 Asset Hub transactions that include `AsPgas`, or expose a signing path that does not require translating PAPI signed-extension data into the limited PJS payload shape.

## Possible Fix Direction

Preferred: avoid the PJS signing bridge for terminal/mobile product-account signing. Pass a PAPI-native signing request shape, or another protocol shape that preserves arbitrary signed-extension payloads, through `host-papp` / mobile.

Narrower: add `AsPgas` support to the bridge, if the current mobile `signPayload` schema can represent everything required by `AsPgas`. This may not be enough if the PJS-shaped payload cannot carry the extension semantics.

decrypto21:
Closes #80 (`PJS does not support this signed-extension: AsPgas` in `dot init`).

Swaps `getPolkadotSignerFromPjs` for PAPI-native `getPolkadotSigner` in `packages/terminal/src/signer.ts`. Transaction signing now routes through `session.signRaw({ data: { tag: "Payload", value: hex(toSign) } })` - an opaque-bytes wire route that already exists in `@novasamatech/host-papp` 0.7.7. polkadot-api assembles `callData ‖ extras ‖ additionalSigneds` for every signed extension declared by the runtime, and the wallet just signs the resulting bytes. Any extension carried by the chain - including ones polkadot-api's PJS adapter doesn't have a built-in mapper for (`AsPgas` on Paseo Next v2, future runtime extensions) - survives end-to-end.

## Root cause

Two stacked limitations on the previous path:

1. **`@polkadot-api/pjs-signer`** hard-throws on any signed extension without a built-in mapper. The mapper table covers eight extensions only (`CheckGenesis`, `CheckNonce`, `CheckTxVersion`, `CheckMortality`, `CheckSpecVersion`, `ChargeAssetTxPayment`, `ChargeTransactionPayment`, `CheckMetadataHash`). `AsPgas` isn't in that list.
2. **`host-papp`'s `SigningPayloadRequest` codec** has fixed slots (`era`, `nonce`, `tip`, `assetId`, `metadataHash`, `mode`, …) and `signedExtensions: Vector(str)` - names only, no payload bytes. Even if PJS were patched to pass `AsPgas` through, the wire format itself can't carry the extension's `extra` and `additionalSigned` bytes.

The `Payload`-tag route on `session.signRaw` is a tagged-bytes channel the upstream `@novasamatech/product-sdk` already uses for non-`bytes`-typed raw requests, so the wallet code path is exercised in the field.

## Changes

- **`packages/terminal/src/signer.ts`** - replaces PJS-bridged construction with `getPolkadotSigner(publicKey, "Sr25519", sign)`. `signBytes` is overridden post-construction to keep the `Bytes`-tag route (so the wallet's `<Bytes>...</Bytes>` anti-phishing wrap applies exactly once to arbitrary-data signing, not to tx signing).
- **`packages/terminal/README.md`** - updates the Signing section to document the new wire path and the AsPgas fix.

Two tiny helpers (`makeTxSignCallback`, `makeRawBytesSignCallback`) replace the previous PJS field-translation glue and are unit-tested directly. Net diff: +150 / −288 lines (refactor reduces the file because the PJS field-translation glue is gone).

## What this does *not* fix (explicit scope)

This PR fixes the terminal/QR-paired flow only (`createSessionSigner` / `createSessionSignerForAccount`). The in-host signer path - `packages/signer/src/providers/host.ts` calling `accountsProvider.getProductAccountSigner()` from `@novasamatech/product-sdk` - stays PJS-bridged and will still throw `PJS does not support this signed-extension: <name>` for arbitrary extensions until two things happen upstream:

1. [`paritytech/triangle-js-sdks#168`](https://github.com/paritytech/triangle-js-sdks/pull/168) merges - adds a `signerType: 'createTransaction'` option to `getProductAccountSigner` that builds a PAPI-native signer routing through the new `host_create_transaction` RPC.
2. We bump `@novasamatech/product-sdk` to the release containing that change and flip the option in `packages/signer/src/providers/host.ts`.

The upstream PR is currently OPEN, mergeable, awaiting review. The follow-up here is effectively one line per call site once it ships.

**Consumer impact in the meantime:** apps using `createSessionSigner` (CLI/QR-paired) are unblocked by this PR. Apps reaching mobile signing via the in-host `host` provider still hit the original error on AsPgas-bearing extrinsics. `playground-cli` uses `createSessionSignerForAccount`, so `dot init` is covered.

## Validation

- `pnpm check` (biome lint + format, 205 files) - clean
- `pnpm build` (all packages + examples) - pass
- `pnpm test` - 725 tests across 14 packages, 0 failures
  - 13 in `packages/terminal/src/signer.ts` covering the new wire path and regression guards against the legacy `signPayload` call
- `pnpm exec tsc --noEmit` (terminal package) - clean
- Bundle size: **no change** (terminal stays at 4.6 KB gzip)

## Test plan

- [x] `pnpm check` (biome) — clean
- [x] `pnpm build` (all packages + examples) - pass
- [x] `pnpm test` — 725/725 across 14 packages
- [x] Bundle size — no change (terminal: 4.6 KB gzip)
- [x] Local e2e roundtrip against Paseo Next v2 Asset Hub

### DIFF
--- a/product-sdk/packages/terminal/README.md
+++ b/product-sdk/packages/terminal/README.md
@@ -201,10 +201,10 @@ const sessions = await waitForSessions(adapter);
 
 After login and attestation, the paired wallet can sign both transactions and raw messages via the statement store. The `PolkadotSigner` returned by `createSessionSigner` routes each path to the right host-papp method:
 
-- **Transactions** (`signTx` from polkadot-api's perspective — what `submitAndWatch`, `signSubmitAndWatch`, contract method calls, etc. invoke) go through `session.signPayload`. Mobile signs the SCALE-encoded extrinsic payload directly, with no envelope wrapping. The chain accepts the resulting signature.
-- **Raw bytes** (`signBytes`) go through `session.signRaw`. Mobile applies the standard `<Bytes>...</Bytes>` anti-phishing wrap before signing — appropriate for arbitrary data, the same behavior `signRaw` has across all Polkadot wallets.
+- **Transactions** (`signTx` from polkadot-api's perspective — what `submitAndWatch`, `signSubmitAndWatch`, contract method calls, etc. invoke) go through `session.signRaw` with the `Payload` tag. polkadot-api assembles the full SCALE-encoded signing payload from runtime metadata — `callData ‖ extras ‖ additionalSigneds` for every signed extension the chain declares — and hands the bytes to the wallet as an opaque hex blob. The wallet signs the payload as-is, with no envelope wrapping. Any signed extension declared by the runtime (including extensions polkadot-api doesn't know about, e.g. `AsPgas` on Paseo Next v2) survives end-to-end because the wallet doesn't inspect the bytes — it just signs them.
+- **Raw bytes** (`signBytes`) go through `session.signRaw` with the `Bytes` tag. Mobile applies the standard `<Bytes>...</Bytes>` anti-phishing wrap before signing — appropriate for arbitrary data, the same behavior `signRaw` has across all Polkadot wallets.
 
-> **Note on a previous bug.** Versions of this package prior to `0.1.1` routed *all* signing through `signRaw`, which caused mobile to wrap SCALE-encoded extrinsic payloads with `<Bytes>...</Bytes>` and produced signatures the chain rejected with `BadProof`. If you hit `BadProof` on every transaction, upgrade.
+> **Note on previous bugs.** Versions prior to `0.1.1` routed *all* signing through `signRaw` with no tag distinction, producing `BadProof`-rejected signatures. Versions `0.1.1` through `0.2.x` then used `polkadot-api/pjs-signer`, which threw `PJS does not support this signed-extension: <name>` on any extension outside its eight built-in mappers (notably `AsPgas`). The current path bypasses both pitfalls.
 
 ## Notes
 
--- a/product-sdk/packages/terminal/src/signer.ts
+++ b/product-sdk/packages/terminal/src/signer.ts
@@ -1,17 +1,23 @@
 /**
  * Create a PolkadotSigner from a QR-paired session.
  *
- * Bridges the host-papp session to polkadot-api's `PolkadotSigner` interface
- * via `getPolkadotSignerFromPjs`, routing **transaction signing** through
- * `session.signPayload` (no `<Bytes>...</Bytes>` envelope — produces a
- * signature over the actual extrinsic payload) and **raw-message signing**
- * through `session.signRaw` (mobile applies the `<Bytes>...</Bytes>`
- * anti-phishing wrap, as expected for arbitrary data).
+ * Builds the `PolkadotSigner` via PAPI-native `getPolkadotSigner` and routes
+ * **transaction signing** through `session.signRaw` with the `Payload` tag
+ * — a tagged-bytes wire route in host-papp that signs the SCALE-encoded
+ * extrinsic payload as-is, with no `<Bytes>...</Bytes>` anti-phishing
+ * envelope. PAPI assembles the full payload (`callData ‖ extras ‖
+ * additionalSigneds`) from runtime metadata, so every signed extension
+ * declared by the chain — including extensions not known to PAPI's PJS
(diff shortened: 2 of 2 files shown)


UtkarshBhardwaj007 (approved):
LGTM

TarikGul (approved):

PR #84, opened 2026-05-13
feat(host,signer): typed permission ergonomics + onConnect lifecycle hook

ReinhardHatko:
## Summary

Every dapp calling `hostApi.requestResourceAllocation` or `hostApi.permission` writes the same boilerplate: re-declare the request/response unions inline because host-api only ships SCALE codecs, cast the request `as never` because the inferred type loses precision through `UnwrapVersionedResult`, and hand-roll the once-per-connect bookkeeping that gates the call on the right signer transition. We hit this pattern twice in `playground-app` alone in two separate sessions;.

This PR collapses that boilerplate into the SDK across three additive layers.

### 1. Type aliases (`@parity/product-sdk-host`)

Derive TS aliases from the `@novasamatech/host-api` codecs via `CodecType<typeof X>` and re-export them, plus tag-only narrows:

- `AllocatableResource`, `AllocatableResourceTag`
- `AllocationOutcome`, `AllocationOutcomeTag`
- `RemotePermission`, `RemotePermissionTag`

Consumers no longer redeclare these inline or import `scale-ts` directly. Schema drift in upstream codecs now surfaces as a TypeScript error at this boundary instead of silently passing through `as never` casts.

### 2. Wrappers (`@parity/product-sdk-host`)

```ts
export async function requestProductPermissions(
    resources: AllocatableResource[],
): Promise<Result<AllocationOutcome[], string>>;

export async function requestPermission(
    permission: RemotePermission,
): Promise<Result<boolean, string>>;
```

Both wrappers build the `v1` versioned envelope, call the underlying `hostApi.*` method, unwrap the response, and return a typed lightweight `Result`. Unrecognized response version tags surface a typed error rather than hard-coding `"v1"`. Falls back to a clean `{ ok: false, error: "Host API unavailable" }` when the host is missing.

Also exports `formatHostError(error)` — moved out of `@parity/product-sdk-signer`'s internals (which now imports it from host) so both packages share one implementation.

### 3. `onConnect` lifecycle hook (`@parity/product-sdk-signer`)

```ts
new SignerManager({
    onConnect: async (account, { requestPermissions, signal }) => {
        const result = await requestPermissions([
            { tag: "SmartContractAllowance", value: account.derivationIndex },
            { tag: "AutoSigning", value: undefined },
        ]);
        if (!result.ok || result.value.some((o) => o.tag !== "Allocated")) {
            captureWarning("partial permissions", result);
        }
    },
});
```

Fires **exactly when** `state.status` flips into `"connected"` with a selected account — not on the many other state mutations (`selectAccount`, account-list updates, etc.) that happen while already connected. Refires after auto-reconnect, so a fresh host session re-runs the callback. `ctx.signal` aborts if the user disconnects or destroys the manager mid-flight. Errors thrown from `onConnect` are logged but don't break the connected state.

Replaces ~50 lines of transition-gated subscription, once-per-session bookkeeping, and HMR cleanup that every product app was writing by hand.

## Consumer impact

Before:

```ts
// Re-declared because host-api only ships the codec
type AllocationOutcomeTag = "Allocated" | "Rejected" | "NotAvailable";
type AllocationResponse = { value: Array<{ tag: AllocationOutcomeTag; value: undefined }> };

const request = enumValue("v1", [
    { tag: "SmartContractAllowance", value: 0 },
    { tag: "AutoSigning", value: undefined },
]) as never; // ← typed shape isn't exported

await hostApi.requestResourceAllocation(request).match(
    (response: AllocationResponse) => { /* walk response.value[i].tag */ },
    (err: unknown) => { /* GenericError, not typed */ },
);

// Plus ~25 lines of transition-gated subscribe() bookkeeping to ensure
// this fires once per connect, resets on disconnect, and cleans up HMR.
```

After:

```ts
new SignerManager({
    onConnect: async (account, { requestPermissions }) => {
        const result = await requestPermissions([
            { tag: "SmartContractAllowance", value: 0 },
            { tag: "AutoSigning", value: undefined },
        ]);
        if (!result.ok) return warn(result.error);
        const [smartContract, autoSigning] = result.value;
        if (smartContract.tag !== "Allocated") warn("contract allowance denied");
    },
});
```

The full `playground-app` workaround (~50 lines) collapses to the block above.

## Documentation

- `SignerManager` class JSDoc now has a full lifecycle block: ASCII diagram of `disconnected → connecting → connected → auto-reconnect`, callback ordering (subscribe fires *before* onConnect's side effects), and a `disconnect()` vs `destroy()` comparison.
- `subscribe()`, `disconnect()`, and `destroy()` each gained JSDoc explaining their firing/reset/terminality semantics.

## Demo

[`examples/signer-demo`](examples/signer-demo/) now wires the `onConnect` hook and adds a **"Subscribe vs onConnect"** panel showing three live counters (subscribe fires / status transitions / onConnect fires) plus a running list of each status flip. Clicking around makes the contract visible: subscribe ticks fast, transitions and onConnect tick only on connect/reconnect.

New e2e spec `e2e/onconnect.spec.ts` locks two invariants:

1. `onConnect` fires once per connect transition and refires after reconnect.
2. `selectAccount` bumps the subscribe count but **not** the transition or onConnect counts.

## Test infra

- **`@parity/host-api-test-sdk` catalog bumped** from `^0.7.3` to `^0.7.5` — picks up the `host_request_resource_allocation` handler the demo's new `onConnect` calls. Without it the e2e test for `ctx.requestPermissions` hangs until timeout because the older test SDK silently drops unknown methods.
- **`e2e/permission.spec.ts` unskipped** and rewritten to assert the auto-request denial path: with the host in `reject-all` mode and `ChainSubmit` revoked, the host records the denial (via `getPermissionLog()`) and `SignerManager.connect()` tolerates it. The original `signRaw`-rejection assertion is parked — test SDK 0.7.5 declares `setEnforcePermissions` but doesn't wire it (the flag is set, nothing reads it). Bring back that assertion when the SDK lands real enforcement.

## Version bump

Pre-1.0 minor for both packages:

- `@parity/product-sdk-host` `0.2.1` → `0.3.0`
- `@parity/product-sdk-signer` `0.2.2` → `0.3.0`

Additive only — no breaking changes. Changeset included.

## Test plan

- [x] `pnpm --filter "@parity/product-sdk-host" build && pnpm --filter "@parity/product-sdk-host" test` (32 tests)
- [x] `pnpm --filter "@parity/product-sdk-signer" build && pnpm --filter "@parity/product-sdk-signer" test` (82 tests including 9 new `onConnect` lifecycle tests)
- [x] `pnpm --filter "@parity/product-sdk-signer-demo" build`
- [x] `pnpm --filter "@parity/product-sdk-signer-demo" test:e2e` — 9/9 (previously 8 pass + 1 skip; the permission spec is unskipped in this PR)
- [ ] CI green
- [ ] Smoke-test the `onConnect` flow inside a real product container (Polkadot Desktop/Mobile) to confirm `ctx.requestPermissions` round-trips to the production host without the test-SDK indirection


TarikGul:
I think you can pull from main and fix the conflicts and this should be about good to go.

TarikGul:
Did a few changes to get this PR inline with main, (claude description):

## TL;DR                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                    
  Main shipped parallel work in #82 (`requestResourceAllocation` + `AllocatableResource` / `AllocationOutcome` aliases) and #92 (host-api-test-sdk catalog at `^0.8.2`) while this PR was open. The fix-up keeps main's implementation wherever the 
  two overlap, drops the duplicates from this branch, and slims the PR down to its uniquely-additive surface.                                                                                                                                       
                                                                                                             
  ## What stays from this PR                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                    
  - **`RemotePermission` + tag-only aliases** (`AllocatableResourceTag`, `AllocationOutcomeTag`, `RemotePermissionTag`) — additive over main's existing exports, now live in `truapi.ts` next to them.
  - **`requestPermission(permission)`** wrapper for `hostApi.permission` — slimmed to match main's `requestResourceAllocation` shape (`Promise<boolean>`, throws on failure).                                                                       
  - **`SignerManager` `onConnect` lifecycle hook** — unchanged in behavior; 9 lifecycle tests still pass.                                                                    
  - **Demo wiring** — the "Subscribe vs onConnect" panel + e2e spec stay.                                                                                                                                                                           
  - **`permission.spec.ts` rewrite** — already uses `ChainSubmit` + `getPermissionLog()`, both present in main's `host-api-test-sdk@0.8.2`.                                                                                                         
                                                                                                                                                                                                                                                    
  ## What this PR no longer adds (now from main)                                                                                                                                                                                                    
                                                                                                                                                                                                                                                    
  - `AllocatableResource` / `AllocationOutcome` type aliases — already in main's `truapi.ts`.                                                                                                                                                       
  - `requestResourceAllocation` wrapper — already in main, throws-style.                                                                                                                                                                            
  - `formatHostError` helper — main keeps its private `formatError` in the signer host provider; the new file is gone.
  - `requestProductPermissions` (the `Result<T, string>` style wrapper) — dropped in favor of main's throws-style `requestResourceAllocation`. The `onConnect` ctx now exposes `requestResourceAllocation` (not `requestPermissions`).              
                                                                                                                                                                                                                                                    
  ## Other corrections in the fix-up                                                                                                                                                                                                                
                                                                                                                                                                                                                                                    
  - **Format check:** `pnpm check` now clean. Earlier "CI green" claim was stale on two files.                                                                                                                                                      
  - **JSDoc + changeset examples:** the `account.derivationIndex` (non-existent field) and `account.publicKey.length` (meaningless filler) references are gone — examples now use `_account` to make it clear the account isn't load-bearing for the
   demo case.                                                                                                                                                                                                                                       
  - **Umbrella bump:** added `"@parity/product-sdk": minor` to the changeset per `pending-changesets/README.md`.                                                                                                                                    
  - **Changeset location:** moved from `.changeset/` to `pending-changesets/` per the staging convention.

TarikGul (approved):

PR #89, opened 2026-05-14
feat: terminal request resource allocation

### LINKED ISSUE #88: RFC-10 product-SDK coverage: wrap `statement_create_proof_authorized` (TruAPI) + `request_resource_allocation` (AP) Body
[RFC-10 - W3S Allowance Management in TrUAPI](https://github.com/paritytech/truapi/blob/main/docs/rfcs/0010-allowance.md) §Compatibility introduces three new product-facing surfaces. #82 covered one; the other two are still unwrapped, plus the Host-side responsibilities the RFC assigns are unimplemented for the terminal-CLI case.

## RFC-10 coverage today

| Surface / responsibility | Where | Status |
|---|---|---|
| `host_request_resource_allocation` (TruAPI) | `@parity/product-sdk-host` | ✅ #82 |
| `statement_create_proof_authorized` (TruAPI) | `@parity/product-sdk-host` | ❌ no wrapper |
| `request_resource_allocation` (Accounts Protocol) | `@parity/product-sdk-terminal/host` | ❌ no wrapper |
| `preimage_submit` / `statement_create_proof` (unchanged) | `@parity/product-sdk-host` | ✅ existing |
| §Stakeholders Host triad: AP client + cache + signer | terminal package | ❌ none of the three |

In a CLI using `@parity/product-sdk-terminal`, the QR-paired mobile wallet is the Account Holder and the CLI plays the Host role. RFC-10 §Stakeholders assigns three jobs to the Host: implement the AP client, cache received slot keys, and sign on the product's behalf. All three are missing for this architecture today.

## Proposed scope

1. **`createProofAuthorized` in `@parity/product-sdk-host`** - typed wrapper for the new SSS TruAPI call. Pairs with #82.
2. **New `./host` subpath in `@parity/product-sdk-terminal`** for the host-runner facet:
   - `requestResourceAllocation(session, adapter, resources, options?)` - AP wrapper with on-disk cache and auto-picked `onExisting` per RFC §"How the Host picks OnExistingAllowancePolicy".
   - `getCachedAllocation(adapter, resource)` - read-only lookup.
   - `createSlotAccountSigner(adapter, resource)` - local sr25519 `PolkadotSigner` backed by the cached slot key. Closes the third §Stakeholders responsibility.
   - `ensureSlotAccountSigner(session, adapter, resource)` - RFC §"Implicit allocation" semantics at the signer level (the RFC's stated trigger `preimage_submit`/`statement_create_proof_authorized` doesn't exist in the CLI architecture, so this helper substitutes).
3. Expose `storageDir` on `TerminalAdapter` so the cache lands next to sessions.

## Upstream readiness

- Mobile handlers shipped: Android [polkadot-app-android-v2#563](https://github.com/paritytech/polkadot-app-android-v2/pull/563), iOS [polkadot-app-ios-v2#759](https://github.com/paritytech/polkadot-app-ios-v2/pull/759). Both return `NotAvailable` for `AutoSigning` today.
- Chain-side authorization plumbing: [polkadot-sdk#11692](https://github.com/paritytech/polkadot-sdk/pull/11692), [polkadot-bulletin-chain#394](https://github.com/paritytech/polkadot-bulletin-chain/pull/394) landed.
- Slot renewal: [individuality#931](https://github.com/paritytech/individuality/pull/931) open. First-time allocation works end-to-end today.

## What this issue does NOT cover

- **Revocation** - RFC-10 §Drawbacks: *"No revocation in v1"*. Deferred to a future RFC.

## Acceptance

- [ ] `createProofAuthorized` exported from `@parity/product-sdk-host` with JSDoc + unit test
- [ ] All four functions above exported from `@parity/product-sdk-terminal/host` with JSDoc + unit tests
- [ ] Per-`appId` cache file with atomic writes and `0o600` permissions
- [ ] Auto-picked `onExisting` rule covering Bulletin / SSS / SC / AutoSigning
- [ ] All types derived from `UserSession['requestResourceAllocation']` so upstream codec renames break compile, not runtime
- [ ] Terminal package README documents the `./host` subpath

decrypto21:
Closes #88

## Description

Closes the unwrapped [RFC-10](https://github.com/paritytech/truapi/blob/main/docs/rfcs/0010-allowance.md) product-SDK surfaces and implements the Host-side responsibilities RFC-10 §Stakeholders assigns for the terminal-CLI architecture. Related to #82.

| RFC-10 surface or responsibility | Where | Status |
|---|---|---|
| `host_request_resource_allocation` TruAPI | `@parity/product-sdk-host` | #82 |
| `statement_create_proof_authorized` TruAPI | `@parity/product-sdk-host` | done |
| `request_resource_allocation` AP | `@parity/product-sdk-terminal/host` | done |
| `preimage_submit`, `statement_create_proof` (unchanged) | unchanged | unchanged |
| §Stakeholders: implement AP client | `requestResourceAllocation` | done |
| §Stakeholders: cache allowance private keys | per-`appId` JSON, atomic writes, `0o600` | done |
| §Stakeholders: use them to sign on the product's behalf | `createSlotAccountSigner` (local sr25519) | done |
| §"Implicit allocation" for the CLI architecture | `ensureSlotAccountSigner` | done |
| §"Renewal is not a product concern" | AH's job - verified absent | by design |
| §Drawbacks: "No revocation in v1" | deferred per the RFC | by design |

### `@parity/product-sdk-host`: `createProofAuthorized`

Typed wrapper around the new `statement_create_proof_authorized` TruAPI call. Same style as #82.

```ts
import { createProofAuthorized, getStatementStore } from "@parity/product-sdk-host";

const proof = await createProofAuthorized({
    proof: undefined,
    decryptionKey: undefined,
    expiry: undefined,
    channel: undefined,
    topics: [],
    data: payload,
});
const store = await getStatementStore();
await store?.submit({ ...statement, proof });
```

### `@parity/product-sdk-terminal`: new `./host` subpath

For CLIs paired via QR with a mobile wallet. The CLI plays the Host role; this subpath is the Host-runner facet.

```ts
import {
    ensureSlotAccountSigner,
    requestResourceAllocation,
    getCachedAllocation,
    createSlotAccountSigner,
    type AllocatableResource,
    type ApAllocationOutcome,
    type CachedAllocation,
} from "@parity/product-sdk-terminal/host";

// One call: ensure allowance, get a local signer. Prompts the wallet on
// first use; subsequent calls are wire-free.
const signer = await ensureSlotAccountSigner(session, adapter, {
    tag: "BulletInAllowance",
    value: undefined,
});
await bulletinTx.submitAndWatch(signer);
```

- `requestResourceAllocation` - AP wrapper. Auto-picks `onExisting`: `Ignore` unless every requested resource is a slot-table variant (Bulletin/SSS) AND already cached, then `Increase`. Override via `options.onExisting`.
- `getCachedAllocation` - read-only lookup; no wire.
- `createSlotAccountSigner` - local sr25519 `PolkadotSigner` backed by the cached slot key. Handles both 32-byte mini-secret and 64-byte expanded forms; throws on other lengths.
- `ensureSlotAccountSigner` - cache hit → signer; miss → allocate, then signer.
- Cache lives at `{adapter.storageDir}/{appId}_AllowanceKeys.json`, atomic write (temp + rename), `0o600`.
- `TerminalAdapter` exposes `readonly storageDir?: string` so the cache lands next to sessions.
- Terminal package README updated with the `./host` API reference.

### Caveats
- **`AutoSigning` returns `NotAvailable`** on both Android ([#563](https://github.com/paritytech/polkadot-app-android-v2/pull/563)) and iOS ([#759](https://github.com/paritytech/polkadot-app-ios-v2/pull/759)) as of 2026-05. The codec accepts the variant; mobile clients don't yet grant it.
- **Slot renewal at expiry** depends on [individuality#931](https://github.com/paritytech/individuality/pull/931) (open). First-time allocation works end-to-end today.
- **`slotAccountKey` wire encoding** isn't pinned by the RFC or host-papp codec. The signer auto-detects 32-byte (mini-secret → derive) vs 64-byte (expanded → direct sign). Any other length throws; a regression test pins the throw.

### Commits

```
986c27b feat(host): expose createProofAuthorized as typed wrapper
56fa065 feat(terminal): expose storageDir on TerminalAdapter
be4ada0 feat(terminal): add allowance-key cache for the host-runner facet
fb23fba feat(terminal): add createSlotAccountSigner for cached allowance keys
85863e8 feat(terminal): add ./host subpath with requestResourceAllocation + ensureSlotAccountSigner
e994e9e docs(terminal): document the ./host subpath
a6f7082 chore: change docs references
4e95e0d fix(terminal): address review feedback - lock scope, variant alignment, error unwrap, single loadCache
```

## Test plan

- [x] Workspace build clean (`pnpm -r build`)
- [x] Workspace unit tests pass - **794 tests across 67 files**, zero failures
  - `@parity/product-sdk-host`: 19 (was 18 + 1 new `createProofAuthorized` test)
  - `@parity/product-sdk-terminal`: 131 (was 65 at PR start, +56 net)
- [x] Biome clean (`pnpm check` - 208 files)
- [x] `./host` subpath resolves end-to-end via `node t.mjs` after symlinking the package
- [x] Co-located tests via `import.meta.vitest` (matches the existing terminal-package convention; same as `adapter.ts`, `signer.ts`, etc.)
- [x] Pre-existing chain-client-demo / contracts-demo / tx-demo e2e failures confirmed environmental (unreachable `wss://paseo-asset-hub-next-rpc.polkadot.io`)

## Design decisions

- **AP-shaped surface, not TruAPI-shaped**: the CLI is the Host, not a Product, so `Allocated` outcomes correctly carry private key material rather than the stripped form #82 returns for in-container products.
- **Cache as required infrastructure**: per RFC-10 §"How the Host picks OnExistingAllowancePolicy", the `Ignore` vs `Increase` decision is cache-driven. Without persistence the `Increase` branch is unreachable, so the cache ships with the wrapper rather than as a follow-up.
- **`0o600` cache file mode** is defense-in-depth, not RFC-mandated - the RFC's threat model explicitly trusts the Host. Documented as such in code comments.
- **`ensureSlotAccountSigner` substitutes for RFC §"Implicit allocation"** in the CLI architecture: the RFC's stated trigger (`preimage_submit` / `statement_create_proof_authorized` TruAPI calls) doesn't exist in a CLI, so the helper folds the implicit-allocation step into signer construction instead.

### DIFF
--- a/product-sdk/packages/terminal/src/adapter.ts
+++ b/product-sdk/packages/terminal/src/adapter.ts
@@ -60,8 +60,8 @@ export type TerminalAdapter = PappAdapter & {
      * The on-disk storage directory used for sessions and (when the
      * host-runner facet is in use) for the allowance-key cache the
      * `./host` subpath maintains. `undefined` when the default
-     * `~/.polkadot-apps/` location is in use; set explicitly when the
-     * caller passed `storageDir` to {@link createTerminalAdapter}.
+     * is in use; set explicitly when the caller passed `storageDir` to
+     * {@link createTerminalAdapter}.
      */
     readonly storageDir?: string;
     /**
--- a/product-sdk/packages/terminal/src/adapter.ts
+++ b/product-sdk/packages/terminal/src/adapter.ts
@@ -57,6 +57,14 @@ export type TerminalAdapter = PappAdapter & {
     /** The `appId` passed to {@link createTerminalAdapter}. Useful for {@link createSessionSigner}. */
     readonly appId: string;
     /**
+     * The on-disk storage directory used for sessions and (when the
+     * host-runner facet is in use) for the allowance-key cache the
+     * `./host` subpath maintains. `undefined` when the default
+     * `~/.polkadot-apps/` location is in use; set explicitly when the
+     * caller passed `storageDir` to {@link createTerminalAdapter}.
+     */
+    readonly storageDir?: string;
+    /**
      * Disconnect the WebSocket and release resources.
      *
      * @remarks
@@ -112,6 +120,7 @@ export function createTerminalAdapter(options: TerminalAdapterOptions): Terminal
     return {
         ...adapter,
         appId: options.appId,
+        storageDir: options.storageDir,
         destroy(): Promise<void> {
             if (destroyPromise) return destroyPromise;
             destroyPromise = teardown(adapter.sessions, trackedLazyClient);
--- a/product-sdk/packages/terminal/src/host-cache.ts
+++ b/product-sdk/packages/terminal/src/host-cache.ts
@@ -1,0 +1,486 @@
+/**
+ * Persistent allowance-key cache. One JSON file per `appId`, 0o600.
+ *
+ * Cache key is the variant tag, except `SmartContractAllowance::{dest}`
+ * which disambiguates per-derivation-index PGAS pre-warming.
+ *
+ * @internal
+ */
+import { createLogger } from "@parity/product-sdk-logger";
+import { toHex } from "@polkadot-api/utils";
+import { mkdir, readFile, rename, writeFile } from "node:fs/promises";
+import { homedir } from "node:os";
+import { dirname, join } from "node:path";
+
+import type { AllocatableResource, ApAllocationOutcome } from "./host.js";
+
+const log = createLogger("terminal");
+
+const DEFAULT_STORAGE_DIR = join(homedir(), ".polkadot-apps");
+const CACHE_FILE_MODE = 0o600;
+
+/** One cached allowance entry. Hex strings are 0x-prefixed. */
+export type CachedAllocation =
+    | { tag: "BulletInAllowance"; slotAccountKey: string }
+    | { tag: "StatementStoreAllowance"; slotAccountKey: string }
+    | { tag: "SmartContractAllowance"; dest: number }
+    | {
+          tag: "AutoSigning";
+          productDerivationSecret: string;
+          productRootPrivateKey: string;
+      };
+
+interface AllowanceCacheV1 {
+    version: 1;
+    entries: Record<string, CachedAllocation>;
+}
+
+function sanitizeAppId(appId: string): string {
+    return appId.replace(/[^a-zA-Z0-9_.-]/g, "_");
+}
+
+function cachePath(appId: string, storageDir?: string): string {
+    return join(storageDir ?? DEFAULT_STORAGE_DIR, `${sanitizeAppId(appId)}_AllowanceKeys.json`);
+}
+
+function emptyCache(): AllowanceCacheV1 {
+    return { version: 1, entries: {} };
+}
+
+export async function loadCache(appId: string, storageDir?: string): Promise<AllowanceCacheV1> {
+    const path = cachePath(appId, storageDir);
+    let raw: string;
+    try {
+        raw = await readFile(path, "utf-8");
+    } catch (e: unknown) {
+        if ((e as NodeJS.ErrnoException)?.code === "ENOENT") return emptyCache();
+        throw e;
+    }
+    try {
+        const parsed = JSON.parse(raw) as AllowanceCacheV1;
+        if (
+            parsed?.version !== 1 ||
+            typeof parsed.entries !== "object" ||
+            parsed.entries === null
+        ) {
+            log.warn("allowance cache schema mismatch; starting fresh", { appId, path });
+            return emptyCache();
+        }
+        return parsed;
+    } catch (e) {
+        log.warn("allowance cache parse failed; starting fresh", { appId, path, error: String(e) });
+        return emptyCache();
+    }
+}
+
+export async function saveCache(
+    appId: string,
+    cache: AllowanceCacheV1,
+    storageDir?: string,
+): Promise<void> {
+    const path = cachePath(appId, storageDir);
+    await mkdir(dirname(path), { recursive: true });
+    // Temp + rename so a mid-write crash can't leave a half-written file.
+    // Concurrent Hosts: the Account Holder serializes per (user, product,
+    // resource) and returns the same key to all callers, so racing writes
+    // converge on identical bytes.
+    const tmp = `${path}.tmp`;
+    await writeFile(tmp, JSON.stringify(cache, null, 2), { mode: CACHE_FILE_MODE });
+    await rename(tmp, path);
+}
+
+export function cacheKey(resource: AllocatableResource): string {
+    if (resource.tag === "SmartContractAllowance") {
+        return `${resource.tag}::${resource.value}`;
+    }
+    return resource.tag;
+}
+
+/**
+ * Pick the wire `onExisting` policy from current cache state. `Ignore`
+ * unless every requested resource is a slot-table variant (Bulletin /
+ * SSS) AND already cached, in which case `Increase`. SC and AutoSigning
+ * aren't slot-additive so they always force `Ignore`.
+ */
+export function pickOnExistingPolicy(
+    cache: AllowanceCacheV1,
+    resources: AllocatableResource[],
+): "Ignore" | "Increase" {
+    if (resources.length === 0) return "Ignore";
+    const allSlotTableAndCached = resources.every((r) => {
+        if (r.tag !== "BulletInAllowance" && r.tag !== "StatementStoreAllowance") return false;
+        return Object.hasOwn(cache.entries, cacheKey(r));
+    });
+    return allSlotTableAndCached ? "Increase" : "Ignore";
+}
+
+/**
+ * Merge `Allocated` outcomes into the cache. Non-`Allocated` outcomes
+ * leave the cache unchanged.
+ */
+export function mergeOutcomes(
+    cache: AllowanceCacheV1,
+    requested: AllocatableResource[],
+    outcomes: ApAllocationOutcome[],
+): AllowanceCacheV1 {
+    const entries = { ...cache.entries };
+    const n = Math.min(requested.length, outcomes.length);
+    for (let i = 0; i < n; i++) {
+        const req = requested[i];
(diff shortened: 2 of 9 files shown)


product-sdk/packages/terminal/src/adapter.ts
```diff
@@ -56,6 +56,14 @@ export interface TerminalAdapterOptions {
 export type TerminalAdapter = PappAdapter & {
     /** The `appId` passed to {@link createTerminalAdapter}. Useful for {@link createSessionSigner}. */
     readonly appId: string;
+    /**
+     * The on-disk storage directory used for sessions and (when the
+     * host-runner facet is in use) for the allowance-key cache the
+     * `./host` subpath maintains. `undefined` when the default
+     * `~/.polkadot-apps/` location is in use; set explicitly when the
```
TarikGul: polkadot-apps ?

product-sdk/packages/terminal/src/host.ts
```diff
@@ -0,0 +1,760 @@
...
+        resources,
+        onExisting,
+    });
+
+    if (result.isErr()) {
+        throw new Error(`requestResourceAllocation failed: ${result.error.message}`);
+    }
+
+    const next = mergeOutcomes(cache, resources, result.value);
+    if (next !== cache) {
```
Imod7: This looks like it is always going to be true so `saveCache` will always run. Can you double check what `mergeOutcomes` returns in `host-cache.ts` ? If this issue is true then we can also add a test so this does not regress.

product-sdk/packages/terminal/src/host-cache.ts
```diff
@@ -0,0 +1,486 @@
...
+
+export async function saveCache(
+    appId: string,
+    cache: AllowanceCacheV1,
+    storageDir?: string,
+): Promise<void> {
+    const path = cachePath(appId, storageDir);
+    await mkdir(dirname(path), { recursive: true });
+    // Temp + rename so a mid-write crash can't leave a half-written file.
+    // Concurrent Hosts: the Account Holder serializes per (user, product,
```
Imod7, line 84: The comment here covers one specific race type: **multiple separate CLI processes** (or multiple host instances) asking the wallet for the same resource at the same time.

What the comment doesn't cover (and the code does not guard either) is another type of race: **one process, parallel calls for different resources**. So, one CLI process issuing parallel calls for different resources. The wallet returns different keys: each call's in-memory cache snapshot is missing the other's key; the last writer wins.

product-sdk/packages/terminal/src/host-cache.ts
```diff
@@ -0,0 +1,486 @@
...
+ * Merge `Allocated` outcomes into the cache. Non-`Allocated` outcomes
+ * leave the cache unchanged.
+ */
+export function mergeOutcomes(
+    cache: AllowanceCacheV1,
+    requested: AllocatableResource[],
+    outcomes: ApAllocationOutcome[],
+): AllowanceCacheV1 {
+    const entries = { ...cache.entries };
+    const n = Math.min(requested.length, outcomes.length);
```
Imod7: Here we silently truncate length mismatches. If the wire returns a shorter or longer outcomes array than requested, today we silently process the prefix. The RFC says length-matched: if the contract ever breaks, we would want a warning rather than silent truncation.

product-sdk/packages/terminal/src/adapter.ts
decrypto21: Resolved

product-sdk/packages/terminal/src/host.ts
```diff
@@ -0,0 +1,760 @@
...
+ *
+ * @example
+ * ```ts
+ * const signer = await ensureSlotAccountSigner(session, adapter, {
+ *   tag: "BulletInAllowance", value: undefined,
+ * });
+ * await someBulletinTx.submitAndWatch(signer);
+ * ```
+ */
+export async function ensureSlotAccountSigner(
```
Imod7, line 150: small nit: in this function we call `loadCache` 3 times. Can we optimize maybe and have just one call and the result can be re used?

Imod7:
Also another tiny nit: this PR's description and the [issue](https://github.com/paritytech/product-sdk/issues/88) it links to, both point to the RFC that lives in triangle-js-sdk repo [here](https://github.com/paritytech/triangle-js-sdks/blob/valentunn-rfc-0010/docs/rfcs/0010-allowance.md). Maybe it is better we point to the RFC under truapi repo? so this [one](https://github.com/paritytech/truapi/blob/main/docs/rfcs/0010-allowance.md)?

product-sdk/packages/terminal/src/host.ts
decrypto21: Resolved in [4e95e0d](https://github.com/paritytech/product-sdk/pull/89/commits/4e95e0de927ee36f0c196b540e6a5872e8e855ff)

product-sdk/packages/terminal/src/host-cache.ts
decrypto21: Resolved in [4e95e0d](https://github.com/paritytech/product-sdk/pull/89/commits/4e95e0de927ee36f0c196b540e6a5872e8e855ff)

product-sdk/packages/terminal/src/host-cache.ts
decrypto21: Resolved in [4e95e0d](https://github.com/paritytech/product-sdk/pull/89/commits/4e95e0de927ee36f0c196b540e6a5872e8e855ff)

product-sdk/packages/terminal/src/host.ts
decrypto21: Resolved in [4e95e0d](https://github.com/paritytech/product-sdk/pull/89/commits/4e95e0de927ee36f0c196b540e6a5872e8e855ff)

Imod7 (approved):

TarikGul (approved):
🚀

PR #99, opened 2026-05-18
ci(descriptors): daily drift detector for @parity/product-sdk-descriptors

### LINKED ISSUE #98: ci: scheduled descriptor-drift check to catch stale @polkadot-apps/descriptors early
## Context

`@polkadot-apps/chain-client` (which we use to talk to Paseo Asset Hub) imports pre-generated PAPI type bindings from `@polkadot-apps/descriptors`. Those bindings are a frozen snapshot of the chain's metadata at the moment that package was last regenerated and published.

When the chain runtime is upgraded — which happens often on Paseo right now, especially anything touching `pallet-revive` — the bundled descriptors go stale. PAPI then either errors with `Incompatible runtime entry RuntimeCall(...)` or, worse, silently mis-decodes a subscription so it never emits an event (the failure mode that bit us in paritytech/polkadot-apps#113 / paritytech/product-sdk#53).

Right now we discover this via E2E tests failing after a merge — late, noisy, and after time has been spent on misdiagnosis. We want an earlier, cheaper signal.

## Proposed work

Add a scheduled GitHub Action that detects descriptor drift before tests do.

Two layers, in order of value:

### Layer 1 (cheap, high value): chain-vs-bundled codeHash diff

A daily job that:

1. Connects to Paseo Asset Hub via PAPI light-client (no funder, no contracts, no signer needed)
2. Reads the chain's current runtime `codeHash` (or `system.lastRuntimeUpgrade.specVersion`)
3. Reads what `@polkadot-apps/descriptors` was generated against (either via a metadata accessor on the package if one exists, or by spec-version lookup)
4. If they differ → open or update a tracking issue with: current chain codeHash, bundled descriptor's codeHash, link to `@polkadot-apps/descriptors` npm page, and a one-line hint ("when SDK republishes, bump the dep and re-run E2E to validate")
5. If a newer `@polkadot-apps/descriptors` is already on npm vs what's in `pnpm-lock.yaml` → also flag it

Single tracking issue (de-duplicated, updated in place via `gh issue edit`) so we don't accumulate noise.

### Layer 2 (lower priority, only if Layer 1 isn't enough): live smoke test

A daily job that does a minimal write tx through the iframe path against Paseo Asset Hub. If `signSubmitAndWatch` doesn't emit a terminal status within ~30s, alert.

This is closer to what the E2E suite already covers, so only worth adding if the cheap check above misses cases (e.g., compatibility issues that don't show up as a codeHash mismatch).

## Why playground-app rather than polkadot-apps

Long-term, the SDK side should have automated descriptor regen + republish on chain runtime drift (worth raising upstream as a separate ask). But playground-app is the consumer with the actual stake in the breakage — better to have our own watchdog than wait for the SDK team's process to catch up. Also makes us first-discoverers, which we are anyway.

## Acceptance criteria

- [ ] `.github/workflows/descriptor-drift.yml` runs daily on schedule + `workflow_dispatch`
- [ ] On drift, files (or updates in place) a single tracking issue with chain codeHash + npm descriptor version
- [ ] No false-positives on stable runtime weeks (i.e. doesn't fire when there's no drift)
- [ ] README addition / `e2e/README.md` blurb documenting the check, so engineers seeing the auto-issue understand what to do

## Out of scope

- Auto-bumping the dep. Manual bump + re-run E2E is fine; the goal is *early signal*, not full automation.
- Catching breakage in `bulletin` or `individuality` chain descriptors — same pattern would apply, but the failure mode is less load-bearing for our current write paths. Add later if it bites.

## Priority

Low / V2. The E2E suite is currently catching breakage adequately; this is about catching it *earlier* and reducing diagnosis cost. Worth doing before the run-up to the Web3 Summit Developer Lab event (June 2026), so we're not paying diagnosis tax at the wrong moment.

## Related

- paritytech/product-sdk#53 — descriptor republish for the current incident
- paritytech/polkadot-apps#113 — the original retry-layer issue (now superseded by paritytech/product-sdk#53)

TarikGul:
closes: https://github.com/paritytech/product-sdk/issues/98

## Summary                                                                                                                                                                                                                                        
   
  Adds a daily scheduled GitHub Action that detects when bundled PAPI descriptors have drifted from the live runtime of any chain `@parity/product-sdk-descriptors` ships bindings for. Today we discover this through E2E failures — late, noisy, and after time has been spent on misdiagnosis. This is the cheap, early signal we asked for.

  ## What changed                                                                                                                                                                                                                                   
                                          
  - **`.github/workflows/product-sdk-descriptors-drift.yml`** — new daily workflow (`0 8 * * *` UTC + `workflow_dispatch`). For each chain under `packages/descriptors/chains/*/`:                                                                  
    - drops the cached `.scale` metadata blob,                                                                                                                                                                                                      
    - runs `npx papi update <key> --config .papi/polkadot-api.json --skip-codegen` with a 60s per-chain timeout,
    - extracts the live `codeHash` and `genesis` and compares against what's pinned,                                                                                                                                                                
    - restores the working tree so subsequent steps see a clean checkout.
                                                                                                                                                                                                                                                    
    Every chain is probed independently — no inference from shared-runtime assumptions. A flaky RPC for one chain doesn't poison the report for the others.                                                                                         
                                                                                                                                                                                                                                                    
  - **Tracking issue, deduplicated by label.** On drift (or partial unreachability), the workflow upserts a single open issue labeled `descriptors-drift` with title `chore(descriptors): runtime drift detected`. The body has a per-chain status  
  table, a collapsed JSON detail block, and step-by-step regen-and-PR instructions. On a fully-clean run the issue is auto-closed.                                                                                                                  
                                                                                                                                                                                                                                                    
  - **Failure-mode handling.** Full network outage (all chains unreachable) emits a `::warning::` and skips the issue update so a single bad-network day doesn't flap. Partial unreachability is reported in the issue alongside drift — a chain    
  unreachable for a week deserves visibility.               
                                                                                                                                                                                                                                                    
  - **`packages/descriptors/README.md`** — new "Detecting Drift" section explaining the mechanism and the regen-and-PR steps.

### DIFF
--- a/.github/workflows/product-sdk-descriptors-drift.yml
+++ b/.github/workflows/product-sdk-descriptors-drift.yml
@@ -1,0 +1,229 @@
+# Detects when @parity/product-sdk-descriptors has drifted from the live
+# runtime of any chain it bundles. PAPI's bundled type bindings are a frozen
+# snapshot at regen time, so when a runtime upgrades the bindings can either
+# error with "Incompatible runtime entry" or silently mis-decode a
+# subscription. This job catches drift before E2E does.
+#
+# Mechanism: `papi update --skip-codegen` connects to each chain's RPC,
+# fetches the current metadata + codeHash, and rewrites
+# `.papi/polkadot-api.json` in place. A diff on `codeHash` (and `genesis`)
+# vs. what's checked in is the drift signal.
+name: "product-sdk: Descriptors drift"
+
+on:
+    schedule:
+        - cron: "0 8 * * *"
+    workflow_dispatch:
+
+concurrency:
+    group: descriptors-drift
+    cancel-in-progress: false
+
+defaults:
+    run:
+        working-directory: product-sdk
+
+permissions:
+    contents: read
+    issues: write
+
+jobs:
+    check:
+        name: "Check descriptor drift"
+        runs-on: ubuntu-latest
+        timeout-minutes: 15
+        steps:
+            - uses: actions/checkout@v4
+
+            - uses: pnpm/action-setup@v4
+              with:
+                  version: 10
+
+            - uses: actions/setup-node@v4
+              with:
+                  node-version: "22"
+                  cache: "pnpm"
+                  cache-dependency-path: product-sdk/pnpm-lock.yaml
+
+            - run: pnpm install --frozen-lockfile
+
+            # For each chain dir under packages/descriptors/chains/, drop the
+            # cached .scale so `papi update` is forced to hit the network,
+            # then diff the rewritten polkadot-api.json against HEAD.
+            #
+            # Per-chain failures are isolated — a 60s wallclock timeout per
+            # chain means a flaky RPC doesn't poison the whole report.
+            #
+            # Output: drift-report.json with one entry per chain:
+            #   { chain, key, pinnedCodeHash, liveCodeHash, pinnedGenesis,
+            #     liveGenesis, status: "clean" | "drift" | "unreachable" }
+            - name: Probe every chain
+              id: probe
+              run: |
+                  set -euo pipefail
+                  cd packages/descriptors
+                  : > drift-report.jsonl
+                  for chain_dir in chains/*/; do
+                    chain=$(basename "$chain_dir")
+                    cfg="$chain_dir.papi/polkadot-api.json"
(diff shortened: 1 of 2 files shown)


eugypalu (approved):
LGTM

PR #113, opened 2026-05-20
fix: contracts revert flag

### LINKED ISSUE #111: `product-sdk-contracts` decodes reverted contract calls as successful return data
`@parity/product-sdk-contracts` appears to ignore nonzero `ReviveApi.call` return flags.

If a contract call reverts, `ReviveApi.call` can return:

```json
{
  "result": {
    "success": true,
    "value": {
      "flags": 1,
      "data": "0x556e617574686f72697a6564"
    }
  }
}
```

Here `flags: 1` indicates a contract-level revert, and the data is the revert payload (`Unauthorized` as UTF-8). The SDK currently treats this as a successful call and tries to decode `data` as the method's normal ABI return value, which can throw a misleading viem decode error such as:

```text
AbiDecodingDataSizeTooSmallError: Data size of 12 bytes is too small for given parameters.
```

- `.query(...)` checks `!dryRun.result.success`, but not `dryRun.result.value.flags`
- `.tx(...)` / `.prepare(...)` preflight does the same in `buildReviveCall`

**Expected behavior**: if `result.success === true` but `result.value.flags` is nonzero, treat the call as reverted instead of decoding it as normal return data. The SDK should return/throw a contract revert result that preserves the raw revert bytes, and ideally a decoded UTF-8 message when applicable.

(In the case of typed errors from ABI I suppose we would want a way to give the decoded error instead of the raw bytes)

decrypto21:
## Summary

Fixes #111. `@parity/product-sdk-contracts` was ignoring the `REVERT` bit on `ReviveApi.call`'s success payload, so a contract-level revert looked like a normal return to the SDK:

- `.query()` handed the revert bytes to viem's ABI decoder, which usually threw `AbiDecodingDataSizeTooSmallError` rather than reporting the revert.
- `.tx()` / `.prepare()` built and submitted the extrinsic, letting users pay gas on a call the chain had already simulated as reverting.

Both code paths now check `(flags & REVERT_FLAG) !== 0` after the existing `success: false` branch and surface the revert as a structured result.

## What changed

- **`.query()`** returns `{ success: false, value: { type: "ContractRevertedWithPayload", data, reason?, decoded? }, gasRequired }` on a revert. The discriminant is intentionally distinct from `pallet-revive`'s existing bare `{ type: "ContractReverted" }` dispatch-error variant (which can also land on `QueryResult.value` on failure) so callers can narrow on `value.type` without ambiguity between the two failure modes.
- **`.tx()` / `.prepare()`** throw a new `ContractRevertedError` from the same pre-flight dry-run that already throws `ContractDryRunFailedError`. The transaction is never submitted. *Caveat:* if a caller passes both `gasLimit` and `storageDepositLimit` overrides, the dry-run (and thus the REVERT pre-check) is skipped entirely — preserving the pre-existing behavior that override-everything callers accept the gas cost of a potentially-reverting tx.
- **Revert decoding** tries viem's `decodeErrorResult` against the contract's ABI first — recovers `Error(string)`, `Panic(uint256)`, and custom ABI errors. Well-known Solidity panic codes (`0x01` assertion, `0x11` overflow, `0x12` div-by-zero, …) are mapped to readable strings — error messages read `Panic: arithmetic overflow` instead of `Panic(0x11)`. Falls back to a strict UTF-8 decode for raw `revert(bytes)` payloads (the issue's `0x556e617574686f72697a6564` → `"Unauthorized"`). Raw bytes are always exposed via `data` regardless.
- **Error message ergonomics.** When the decoded error is the canonical `Error(string)`, `ContractRevertedError.message` prints the bare reason (`Contract reverted in "transfer": Unauthorized`) rather than the redundant `Error("Unauthorized")` form. Custom errors keep the `Name(args)` form.
- **Flag check is masked, not `!== 0`.** Only bit 0 of `ReturnFlags` is the REVERT signal; other bits are reserved. A bare `!== 0` would false-positive a successful return if `pallet-revive` ever sets a non-revert bit.

## New exports

- `ContractRevertedError` (extends `ContractError`) — `methodName` / `data: HexString` / `reason?: string` / `decoded?: { errorName, args }`
- `ContractRevertInfo` type — the `value` shape on a reverted query result (`type: "ContractRevertedWithPayload"`)
- `DecodedContractRevert` type — viem-decoded error info

## Tests

120 tests pass in `@parity/product-sdk-contracts` — up from 103 on `main`, **+17 new tests:**
- `decodeRevert` (7): `Error(string)`, `Panic(uint256)` for both well-known and unknown codes, ABI-defined custom errors, raw UTF-8 (the issue's exact payload), opaque non-UTF-8 bytes, and empty payload
- `wrapContract — REVERT flag handling` (6): `.query()` returning the tagged value (with the issue's payload and end-to-end through an empty-payload revert), `.tx()` and `.prepare()` throwing `ContractRevertedError`, `Error(string)` payload via `.query()`, and a regression guard for `flags === 0` staying on the happy path
- `ContractRevertedError` message builder (4): top-level and nested bigint args (`[{ owed: 42n }]`), `Error(string)`-prefers-bare-reason

## Local CI sweep

- `pnpm check` clean (biome)
- `pnpm build` clean across all 14 packages (ESM + DTS)
- `pnpm test` — 842 tests passing across the workspace

## Migration

Callers that wrap `.query()` in `if (result.success)` already work — the failure branch just gains a richer `value`. Existing `.tx()` / `.prepare()` call sites that catch `ContractDryRunFailedError` should add a `ContractRevertedError` branch (or catch the shared `ContractError` base) to distinguish a contract-level revert from a dispatch error.

```ts
try {
  await contract.transfer.tx(to, amount);
} catch (err) {
  if (err instanceof ContractRevertedError) {
    // err.reason / err.decoded / err.data
  } else if (err instanceof ContractDryRunFailedError) {
    // dispatch error (Module / AccountNotMapped / etc.)
  }
}
```

## Test plan
-  CI passes (lint / build / test / bundle-size / e2e / docs-verify)
-  117 tests pass

### DIFF
--- a/product-sdk/packages/contracts/src/errors.ts
+++ b/product-sdk/packages/contracts/src/errors.ts
@@ -1,3 +1,5 @@
+import type { HexString } from "polkadot-api";
+
 /** Base class for all contract errors. Use `instanceof ContractError` to catch any contract-related error. */
 export class ContractError extends Error {
     constructor(message: string, options?: ErrorOptions) {
@@ -55,6 +57,63 @@ export class ContractDryRunFailedError extends ContractError {
     }
 }
 
+/** viem-decoded standard or ABI-defined contract error. */
+export interface DecodedContractRevert {
+    errorName: string;
+    args: readonly unknown[] | undefined;
+}
+
+/**
+ * Tagged-enum value surfaced on `QueryResult.value` when a contract reverts
+ * via the REVERT flag. The discriminant is intentionally distinct from
+ * `pallet-revive`'s bare `{ type: "ContractReverted" }` dispatch-error variant,
+ * which is the other path that can populate `QueryResult.value` on failure.
+ */
+export interface ContractRevertInfo {
+    type: "ContractRevertedWithPayload";
+    data: HexString;
+    reason?: string;
+    decoded?: DecodedContractRevert;
+}
+
+// Top-level bigints stringify unquoted (`42`); bigints inside an object or
+// array stringify as JSON strings (`"42"`) because that's the only way a
+// JSON replacer can emit them. Tolerated since this string is only ever
+// read by humans in an error message, not parsed.
+function stringifyArg(arg: unknown): string {
+    if (typeof arg === "bigint") return arg.toString();
+    return JSON.stringify(arg, (_, v) => (typeof v === "bigint" ? v.toString() : v));
+}
+
+/** A contract call returned with the `REVERT` flag set on a dispatched-OK call. */
+export class ContractRevertedError extends ContractError {
+    readonly methodName: string;
+    readonly data: HexString;
+    readonly reason?: string;
+    readonly decoded?: DecodedContractRevert;
+
+    constructor(
+        methodName: string,
+        data: HexString,
+        info?: { reason?: string; decoded?: DecodedContractRevert },
+    ) {
+        // `Error(string)` is the canonical "revert with a message" path -
+        // print the bare reason rather than `Error("...")` for readability.
+        const suffix =
+            info?.decoded && info.decoded.errorName !== "Error"
+                ? `${info.decoded.errorName}(${(info.decoded.args ?? []).map(stringifyArg).join(", ")})`
+                : (info?.reason ?? data);
+        super(
+            `Contract reverted in "${methodName}": ${suffix}. The transaction was not submitted.`,
+        );
+        this.name = "ContractRevertedError";
+        this.methodName = methodName;
+        this.data = data;
+        this.reason = info?.reason;
+        this.decoded = info?.decoded;
+    }
+}
+
 if (import.meta.vitest) {
     const { test, expect, describe } = import.meta.vitest;
 
@@ -69,6 +128,9 @@ if (import.meta.vitest) {
             expect(new ContractSignerMissingError()).toBeInstanceOf(ContractError);
             expect(new ContractNotFoundError("@a/b", "abc")).toBeInstanceOf(ContractError);
             expect(new ContractDryRunFailedError("foo", "x")).toBeInstanceOf(ContractError);
+            expect(new ContractRevertedError("foo", "0x" as HexString)).toBeInstanceOf(
+                ContractError,
+            );
         });
     });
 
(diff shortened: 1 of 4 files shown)


product-sdk/packages/contracts/src/errors.ts
```diff
@@ -108,4 +170,55 @@ if (import.meta.vitest) {
...
+            expect(err.name).toBe("ContractRevertedError");
+            expect(err.methodName).toBe("transfer");
+            expect(err.data).toBe(data);
+            expect(err.reason).toBe("Unauthorized");
+            expect(err.decoded?.errorName).toBe("Error");
+            expect(err.message).toContain("transfer");
+            expect(err.message).toContain("Unauthorized");
+            expect(err.message).toContain("not submitted");
+            // Error(string) prefers the bare reason over `Error("...")` form.
+            expect(err.message).not.toContain('Error("');
```
pgherveou: just assert what the message is instead of making 4 toContain

product-sdk/packages/contracts/src/wrap.ts
```diff
@@ -101,6 +111,67 @@ function hexToBytes(hex: HexString): Uint8Array {
...
 }
 
+/** Convert a `Uint8Array` to a `0x`-prefixed hex string. */
+function bytesToHex(bytes: Uint8Array): HexString {
+    let hex = "0x";
+    for (let i = 0; i < bytes.byteLength; i++) {
+        hex += bytes[i].toString(16).padStart(2, "0");
+    }
+    return hex as HexString;
+}
```
pgherveou: can't this be imported from existing deps already ?

product-sdk/packages/contracts/src/wrap.ts
```diff
@@ -101,6 +111,67 @@ function hexToBytes(hex: HexString): Uint8Array {
...
+    "0x21": "invalid enum value",
+    "0x22": "improperly encoded storage byte array",
+    "0x31": "pop on empty array",
+    "0x32": "array index out of bounds",
+    "0x41": "memory allocation overflow",
+    "0x51": "call to uninitialized internal function",
+};
+
+/** Decode a revert payload via the ABI, falling back to UTF-8 for raw `revert(bytes)` strings. */
+function decodeRevert(abi: AbiEntry[], data: Uint8Array): ContractRevertInfo {
```
pgherveou, line 142: this might be in viem.sh already

product-sdk/packages/contracts/src/errors.ts
```diff
@@ -55,6 +57,63 @@ export class ContractDryRunFailedError extends ContractError {
...
+    readonly decoded?: DecodedContractRevert;
+
+    constructor(
+        methodName: string,
+        data: HexString,
+        info?: { reason?: string; decoded?: DecodedContractRevert },
+    ) {
+        // `Error(string)` is the canonical "revert with a message" path -
+        // print the bare reason rather than `Error("...")` for readability.
+        const suffix =
```
Imod7, line 102: For Panic reverts, `info.reason` is `"Panic: arithmetic overflow"` but the message ends up `Panic(17)` because this branch only uses `info.reason` when `errorName === "Error"`. Nice to have: add one regression test.

product-sdk/packages/contracts/src/wrap.ts
```diff
@@ -186,6 +259,11 @@ async function buildReviveCall(
         if (!dryRun.result.success) {
             throw new ContractDryRunFailedError(methodName, dryRun.result.value);
         }
+        if ((dryRun.result.value.flags & REVERT_FLAG) !== 0) {
+            // Fail fast so callers don't pay gas on a call the chain already told us would revert.
```
Imod7, line 263: Worth maybe adding a `log.debug("Contract reverted", { methodName, reason: decoded?.reason })` here and same in wrap.ts:342?

Imod7 (commented):
Does the contract related [SKILL](https://github.com/paritytech/product-sdk/blob/main/product-sdk/skills/product-sdk-contracts/SKILL.md) need any update?

Imod7 (approved):

TarikGul:
Just needs a changeset in ./pending-changesets

TarikGul (approved):
Looks good, I'll update the pending changeset.

Issue #93, opened 2026-05-15
Migrate host-playground to @parity/product-sdk


decrypto21:
## Background

[paritytech/host-playground](https://github.com/paritytech/host-playground) (deployed at host-playground.dot.li) is the public reference playground for the Polkadot Host API — a tabbed test runner exercising each host-product method. It previously imported `host-api-wrapper@^0.7.9` throughout. This migrates it to **`@parity/product-sdk`** so the public playground demonstrates Parity's SDK.

## What shipped

The playground now runs **100% on `@parity/product-sdk`**. Three packages were added; no new wrapper packages were needed:

- `@parity/product-sdk-host` `^0.5.0`
- `@parity/product-sdk-signer` `^0.4.0`
- `@parity/product-sdk-chain-client` `^0.5.0`

A prerequisite cleanup PR (**#32**) landed on `main` first, deleting ~37 demo tests — the entire `chat` category, legacy sign / chain-head / extension probes, `payment-top-up`/`payment-request`, and legacy storage/statement/account variants. This migration was rebased on top of it, so it targets only the **63 surviving tests**.

## Key deviations from the original plan

| Original plan | What actually happened |
|---|---|
| Add `-host`, `-signer`, `-statement-store`, `-bulletin` | Only **`-host`, `-signer`, `-chain-client`**. Statement store, preimage, storage, theme, payments, entropy all come from **`-host`** wrappers — no separate `-statement-store`/`-bulletin` packages. |
| "Path 2": route 6 wrapper-less domains through raw `getTruApi().*` | Unnecessary — `-host` ships dedicated wrappers (`getThemeProvider`, `getPaymentManager`, `deriveEntropy`, `getStatementStore`, `getPreimageManager`, `getHostLocalStorage`). Raw `getTruApi()` (via a `truApi()` accessor) is used **only** for the TruAPI-surface calls that have no wrapper — `requestResourceAllocation`, `featureSupported`, `permission`, navigation, etc. |
| `createProductChatManager` → `getTruApi().chat.*` | **Chat removed entirely** by #32 — not migrated. |
| Swap hand-rolled Bulletin upload for `@parity/product-sdk-bulletin` | **Not done** — Bulletin upload stays hand-rolled (`multiformats` CID + `@noble/hashes` blake2b). #32 fixed its gateway URL + added a propagation retry. |
| Keep `@novasamatech/*` as deps | **Removed entirely** — nothing imports it. |
[...]

Issue #85, opened 2026-05-13
Merchant Payments Product SDK


replghost:
## Summary

Add Merchant Payments as a product-layer TypeScript SDK in `product-sdk`.

Merchant payments should be implemented as product-side composition over RFC17 CoinPayment primitives, not as a new primitive host/user-agent capability. A merchant product should be able to bring this SDK into its SPA, create checkout intents, create QR/deep-link invoices, subscribe to checkout status, read supplemental receipts, inspect merchant operating-balance summary state, and initiate refunds while the host only exposes the lower-level CoinPayment surface.

This is intentionally not a TrUAPI host capability. It is a product-layer merchant checkout SDK/profile over CoinPayment.

## Motivation

Merchant checkout needs more than a raw Coinage receive primitive:

- checkout intent lifecycle state;
- sale amount, payment asset, optional POS/display references;
- invoice/QR/deep-link creation;
- fiat quote locking for EUR-priced sales;
- ordered status events for terminal UI;
- supplemental receipt artifacts;
- refund linkage and refund policy;
- merchant operating-balance summary state;
- reconciliation references without payer identity or private Coinage data.

RFC17 provides the lower-level CoinPayment substrate: purses, receivables, cheques, invoices, deposits, clearing references, purse rebalancing, and refunds. The Merchant Payments Product SDK should not duplicate those host primitives. It should compose them into merchant acceptance semantics.

This lets products like T3rminal bundle the merchant SDK while mobile/PWA/desktop hosts only need to expose RFC17 `coinpayment` plus generic product-sdk storage/signing/statement-store capabilities as needed.

## Proposed Package Shape

Add a product-sdk leaf package matching existing package conventions:

```text
packages/merchant-payments
```

Suggested package name:

```text
@parity/product-sdk-merchant-payments
```

Suggested umbrella export:

```ts
import { createMerchantPaymentsSdk } from "@parity/product-sdk/merchant-payments";
```

The package should expose a product-layer SDK, not a host extension. The host/user-agent should not need to expose `window.ua.ext.merchantPayments` for the final architecture.

## Proposed Product API

Concrete TypeScript naming can follow product-sdk style, but the behavioral contract should cover merchant checkout semantics:

```ts
createIntent(request): Promise<{ intent }>
getIntent({ intentId }): Promise<MerchantPaymentIntent>
createInvoice(request): Promise<{ invoice }>
subscribeIntentStatus(request, callback, onError?): () => void
[...]

PR #118, opened 2026-05-20
feat: host playground migration support

### LINKED ISSUE #103: Host: widen typed surface of @novasamatech/product-sdk wrappers (Accounts, LocalStorage, StatementStore, Preimage)
## Summary

Four wrappers in `@parity/product-sdk-host` declare narrower TypeScript interfaces than the underlying `@novasamatech/product-sdk@0.7.8` objects they cast from. The runtime works, but call sites that need the omitted methods fail typecheck unless they fall back to a direct `@novasamatech/product-sdk` import or `as any`. This blocks `host-playground` (and downstream products) from dropping direct Novasama imports for these four domains.

### 1. `AccountsProvider` missing `requestLogin` / `getUserId`
- File: `product-sdk/packages/host/src/truapi.ts` — `AccountsProvider`
  interface (around line 420).
- Novasama exposes both, fully typed
  (`@novasamatech/product-sdk/dist/accounts.d.ts:11,16`):
  ```ts
  getUserId(): ResultAsync<{ primaryUsername: string }, GetUserIdErr>;
  requestLogin(reason?: string):
    ResultAsync<"success" | "alreadyConnected" | "rejected", LoginErr>; Background: PR #23 was the explicit "expand the typed surface" 
  ```
- Background: PR #23 was the explicit "expand the typed surface" pass but the body lists only 6 methods — requestLogin/getUserId were not surveyed. Oversight, not a deliberate carve-out.
- These methods correspond to RFC-0009 (Login / user identity).

### 2. HostLocalStorage missing readBytes / writeBytes
- File: product-sdk/packages/host/src/types.ts:7-17.
- Current shape: readString / writeString / readJSON / writeJSON / clear. No Uint8Array variants.
- Novasama exposes both at runtime (@novasamatech/product-sdk/dist/localStorage.js:9-26):

  ```ts
   readBytes(key: string): Promise<Uint8Array | null>;
   writeBytes(key: string, value: Uint8Array): Promise<void>;
   ```

- Apps storing binary data (signatures, preimages, SCALE-encoded blobs) have to drop to direct Novasama imports today.

### 3. HostStatementStore input types narrower than Novasama
- File: product-sdk/packages/host/src/types.ts:69-100.
- Parity:
   ```ts
    createProof(accountId: [string, number], statement: unknown):
      Promise<StatementProof>;
    subscribe(filter, callback: (page: StatementsPage) => void);
    // where StatementsPage.statements: unknown[]
    ```

- Novasama (@novasamatech/product-sdk/dist/statementStore.d.ts):
    ```ts
    createProof(accountId: ProductAccountId, statement: Statement):
      Promise<StatementProof>;        // same proof union
    // StatementsPage.statements: SignedStatement[]
    ```

- The return shape of createProof matches. What diverges is the input types ([string, number] vs ProductAccountId codec; statement: unknown vs Statement codec) plus statements: unknown[] on the subscription page.

### 4. PreimageManager.lookup subscription shape diverges
- File: product-sdk/packages/host/src/truapi.ts:197-215.
- Parity: 
    ```ts
    returns { unsubscribe: () => void; onInterrupt: (cb) => () => void }.
    ```
- Novasama (@novasamatech/product-sdk/dist/preimage.d.ts:3): returns Subscription<void> from @novasamatech/host-api.
- Apps consuming a Subscription<void> can't pass the Parity return value through without rewrapping.

## Acceptance criteria
host-playground/src/lib/tests.ts can replace the four direct Novasama imports (createAccountsProvider, hostLocalStorage, createStatementStore, preimageManager) with the corresponding Parity getters (getAccountsProvider, getHostLocalStorage, getStatementStore, getPreimageManager) without as any casts, and typecheck passes.

## Implementation
For each, two paths:
1. Hand-mirror the full Novasama shape in the Parity interface (current pattern). 
2. Re-export the Novasama types directly, e.g. export type { AccountsProvider } from "@novasamatech/product-sdk";.

## Related
#93 (Migrate host-playground to @parity/product-sdk)

### LINKED ISSUE #115: Host: add wrappers for upstream symbols missing from @parity/product-sdk-host
## Summary

`@parity/product-sdk-host` exposes wrappers for a subset of `@novasamatech/product-sdk@0.7.8`'s surface, but several upstream symbols have no Parity equivalent. Apps that need them (notably `host-playground`) must import them directly from `@novasamatech/product-sdk`, defeating the package boundary the host wrapper is meant to provide. Sibling issue to #103 - that one widens existing wrappers; this one adds missing ones.

### 1. Theme subscription
- Upstream: `createThemeProvider`, `ThemeMode` (`@novasamatech/product-sdk/dist/theme.d.ts`).
- No `@parity/product-sdk-host` getter. Apps wanting the host theme today must `await import("@novasamatech/product-sdk")` themselves.

### 2. Entropy derivation (RFC-0007)
- Upstream: `hostApi.deriveEntropy(...)` reachable via `getTruApi`, but consumers have to wrap in `enumValue("v1", ...)` and unwrap a neverthrow `ResultAsync` themselves.
- No throw-on-error wrapper matching the shape of `requestPermission` / `requestResourceAllocation`.

### 3. Device permission flow
- Upstream: `requestDevicePermission`, `DevicePermissionKind`, `RemotePermissionItem` (`@novasamatech/product-sdk/dist/permission.d.ts`).
- The `@parity/product-sdk-host` package has `requestPermission` (remote permission) but no device-permission equivalent.

### 4. Sandbox transport (escape hatch)
- Upstream: `sandboxTransport` singleton (`@novasamatech/product-sdk/dist/sandboxTransport.d.ts`).
- Required when calling factory forms that take an explicit `Transport` parameter; no Parity accessor today.

### 5. Wire-level connection status
- Upstream: `metaProvider` / `createMetaProvider` (`@novasamatech/product-sdk/dist/metaProvider.d.ts`).
- `SignerManager.state.status` covers the application-level status; products needing the raw transport heartbeat have to reach upstream directly.

### 6. Spektr / polkadot.js extension injection
- Upstream: `injectSpektrExtension`, `createLegacyExtensionEnableFactory` (`@novasamatech/product-sdk/dist/injectWeb3.d.ts`).
- Required by legacy polkadot.js-style consumers expecting `window.injectedWeb3.spektr`.

### 7. Chat surface (`host_chat_*`)
- Upstream: `createProductChatManager`, `matchChatCustomRenderers`, plus types `ChatBotRegistrationResult`, `ChatCustomMessageRenderer`, `ChatCustomMessageRendererParams`, `ChatMessageContent`, `ChatReceivedAction`, `ChatRoom`, `ChatRoomRegistrationResult` (`@novasamatech/product-sdk/dist/chat.d.ts`).
- No Parity wrapper.

### 8. Factory forms for storage / preimage
- Upstream exposes both singletons (`hostLocalStorage`, `preimageManager`) and constructors (`createLocalStorage(transport?)`, `createPreimageManager(transport?)`).
- `@parity/product-sdk-host` only wraps the singletons via `getHostLocalStorage` / `getPreimageManager`. Tests and advanced consumers needing a custom-transport instance have to drop to upstream.

### 9. `WellKnownChain` constants
- Upstream: `WellKnownChain` (`@novasamatech/product-sdk/dist/constants.d.ts`) - genesis hashes for Polkadot/Kusama/Westend/Rococo relays and asset hubs.
- No Parity equivalent in `@parity/product-sdk-chain-client` (the natural home, not host).

## Acceptance criteria

`host-playground/src/lib/tests.ts` can replace direct `@novasamatech/product-sdk` imports of `createThemeProvider`, `createMetaProvider` / `metaProvider`, `sandboxTransport`, `injectSpektrExtension`, `createLegacyExtensionEnableFactory`, `createProductChatManager`, `matchChatCustomRenderers`, `deriveEntropy`, `requestDevicePermission`, `createLocalStorage`, `createPreimageManager`, and `WellKnownChain` with their corresponding `@parity/product-sdk-host` (or `@parity/product-sdk-chain-client` for `WellKnownChain`) wrappers, without `as any` casts, and typecheck passes.

Only `createPaymentManager` should remain as a direct Novasama import after this issue lands (tracked separately under merchant payments / #85).

## Implementation

Each wrapper follows the existing `@parity/product-sdk-host` pattern: a dynamic `await import("@novasamatech/product-sdk")` inside `try`/`catch` returning `null` if the optional peer dep isn't installed. Types re-exported as `typeof upstreamSymbol` / `ReturnType<typeof upstreamFactory>` so they stay in lockstep.

`WellKnownChain` is the exception - placed in `@parity/product-sdk-chain-client` (its natural home) and hand-copied from upstream rather than re-exported, since genesis hashes are immutable and a runtime dep on Novasama is undesirable for chain-client.

## Related

- #103 (sibling; widens existing wrappers' typed surface)
- #93 (host-playground migration - this issue unblocks the remaining symbols)
- #85 (merchant payments; covers `createPaymentManager`)

decrypto21:
## Summary

Closes #103, closes #115. Unblocks #93 (host-playground migration) for every Novasama import except `createPaymentManager` (tracked separately in #85) and 5 low-level symbols (`metaProvider`, `createMetaProvider`, `sandboxTransport`, `injectSpektrExtension`, `createLegacyExtensionEnableFactory`) that are intentionally not wrapped — see *Out of scope* below.

### `feat(host, storage): widen typed surface of Novasama wrappers (#103)`

Four host wrappers declared narrower TypeScript interfaces than the upstream `@novasamatech/product-sdk` runtime objects they cast from. The runtime worked, but typecheck failed for any method the Parity interface omitted, forcing direct Novasama imports or `as any` at call sites.

- `HostLocalStorage = typeof hostLocalStorage` - adds `readBytes` / `writeBytes` and the typed `clear(key)` signature.
- `HostStatementStore = ReturnType<typeof createStatementStore>` - tightens `createProof` input types from `[string, number]` + `unknown` to `ProductAccountId` + `Statement` codecs, and `StatementsPage.statements` from `unknown[]` to `SignedStatement[]`.
- `PreimageManager = typeof preimageManager` - aligns `lookup` return type to `Subscription<void>`.
- `AccountsProvider = ReturnType<typeof createAccountsProvider>` - adds `requestLogin` and `getUserId` (RFC-0009), fixes the existing `getProductAccount` type lie that claimed `name?` when the runtime never returned one.
- New named re-exports added in `types.ts`: `Statement`, `SignedStatement`, `Topic`, `ProductAccountId`.
- `mockHostStorage` in `packages/storage/src/kv-store.ts` updated to satisfy the widened `HostLocalStorage`.

### `feat(host, chain-client): add missing wrappers for host-playground migration (#115)`

All wrappers ship **flat-in-host** (`getX()` getters returning `T | null`), matching the existing pattern (`getAccountsProvider`, `getStatementStore`, `getPreimageManager`). The optional-peer-dep contract is preserved via dynamic `await import("@novasamatech/product-sdk")` + try/catch null fallback.

**New domain wrappers**:

- `getThemeProvider` + `ThemeMode` / `ThemeProvider` types
- `deriveEntropy(key)` (RFC-0007) - throws-on-error Promise matching the `requestPermission` pattern
- `requestDevicePermission(kind)` + `DevicePermissionKind` / `RemotePermissionItem` aliases
- `WellKnownChain` constant (in `@parity/product-sdk-chain-client`, hand-copied genesis hashes - immutable values, avoids dragging Novasama into chain-client as a runtime dep)
- `getChatManager` + `matchChatCustomRenderers` + 7 chat type re-exports (see *Chat shape* below)

**Factory forms** (for callers needing custom transports):

- `createHostLocalStorage(transport?)`
- `createHostPreimageManager(transport?)`

**TruApi type narrowing**: `TruApi` changes from `any` to `typeof hostApi`, so `truApi.foo(...)` calls keep full inference instead of decaying to `any`. 

## Deliberate deviations from #93

**Chat lives flat-in-host as `getChatManager()`, not as `getTruApi().chat.*` per #93's migration table.** Two reasons:

1. The upstream JS `hostApi` is a flat object - there is no `.chat` accessor to mirror. Building `getTruApi().chat.*` would mean synthesizing the namespacing layer ourselves, which is the same amount of work as a flat wrapper and creates a JS structure not reflected upstream.
2. The other three wrappers in this PR (theme, entropy, device-permission) all ship flat-in-host. Chat follows the established pattern.

host-playground's own migration doc already flagged this gap: ["the Rust `TrUAPI::chat` module structure doesn't map directly to a JS `.chat.*` namespace yet on `hostApi`, which is a flat object."](https://github.com/paritytech/host-playground/blob/main/docs/migration/09-not-migrated.md). If a namespaced view is wanted later, it can be layered on top without breaking `getChatManager()`.

## Verification

Run from the repo root:

```bash
pnpm --filter "@parity/product-sdk-host" exec tsc --noEmit  # exit 0
pnpm --filter "@parity/product-sdk-host" test --run         # 28 tests pass
pnpm --filter "@parity/product-sdk-host" build              # dist/index.d.ts ~30 KB

for pkg in storage statement-store bulletin signer terminal chain-client; do
  pnpm --filter "@parity/product-sdk-$pkg" exec tsc --noEmit
  pnpm --filter "@parity/product-sdk-$pkg" test --run
done
```

Local results: 
- host 28 ✓, 
- storage 13 ✓, 
- statement-store 48 ✓, 
- bulletin 75 ✓, 
- signer 88 ✓, 
- terminal 131 ✓, 
- chain-client 28 ✓ 
- 411 tests total, typecheck clean across all packages.

## Out of scope

- Payments (createPaymentManager) - tracked in #85.
- Wrappers for metaProvider / createMetaProvider / sandboxTransport / injectSpektrExtension / createLegacyExtensionEnableFactory - intentionally not added. host-playground's production code supersedes these with higher-level Parity surfaces (SignerManager.state.status, isInsideContainer, SignerManager.connect()); only its tests.ts feature-matrix uses them directly, where direct Novasama imports are the right answer.
- host-playground side migration - this PR closes the product-sdk-side gaps; the actual find-and-replace in host-playground happens after this lands.
- getTruApi().chat.* namespacing - addressed by the flat-in-host choice above. Re-open if you want a different shape; the getChatManager() API doesn't preclude adding a namespaced view later.

## What host-playground gains
The Novasama import line in [host-playground/src/lib/tests.ts](https://github.com/paritytech/host-playground/blob/main/src/lib/tests.ts) can shrink from 19 symbols to 6 - createPaymentManager (tracked in #85) plus the 5 low-level symbols listed in Out of scope above. Everything else routes through @parity/product-sdk-host (or @parity/product-sdk-chain-client for WellKnownChain).

### DIFF
--- a/product-sdk/packages/host/src/chat.ts
+++ b/product-sdk/packages/host/src/chat.ts
@@ -12,6 +12,8 @@
  * @module
  */
 
+import { createLogger } from "@parity/product-sdk-logger";
+
 import type {
     ChatBotRegistrationResult as NovasamaChatBotRegistrationResult,
     ChatCustomMessageRenderer as NovasamaChatCustomMessageRenderer,
@@ -22,6 +24,8 @@ import type {
     ChatRoomRegistrationResult as NovasamaChatRoomRegistrationResult,
     createProductChatManager,
 } from "@novasamatech/product-sdk";
+
+const log = createLogger("host:chat");
 
 /** Chat message payload variants. Re-exported from `@novasamatech/product-sdk`. */
 export type ChatMessageContent = NovasamaChatMessageContent;
@@ -79,7 +83,8 @@ export async function getChatManager(): Promise<ChatManager | null> {
     try {
         const sdk = await import("@novasamatech/product-sdk");
         return sdk.createProductChatManager();
-    } catch {
+    } catch (err) {
+        log.debug("getChatManager unavailable", err);
         return null;
     }
 }
--- a/product-sdk/packages/host/src/chat.ts
+++ b/product-sdk/packages/host/src/chat.ts
@@ -23,29 +23,29 @@ import type {
     ChatRoom as NovasamaChatRoom,
     ChatRoomRegistrationResult as NovasamaChatRoomRegistrationResult,
     createProductChatManager,
-} from "@novasamatech/product-sdk";
+} from "@novasamatech/host-api-wrapper";
 
 const log = createLogger("host:chat");
 
-/** Chat message payload variants. Re-exported from `@novasamatech/product-sdk`. */
+/** Chat message payload variants. Re-exported from `@novasamatech/host-api-wrapper`. */
 export type ChatMessageContent = NovasamaChatMessageContent;
 
-/** Action received via {@link ChatManager.subscribeAction}. Re-exported from `@novasamatech/product-sdk`. */
+/** Action received via {@link ChatManager.subscribeAction}. Re-exported from `@novasamatech/host-api-wrapper`. */
 export type ChatReceivedAction = NovasamaChatReceivedAction;
 
-/** Room metadata delivered to {@link ChatManager.subscribeChatList}. Re-exported from `@novasamatech/product-sdk`. */
+/** Room metadata delivered to {@link ChatManager.subscribeChatList}. Re-exported from `@novasamatech/host-api-wrapper`. */
 export type ChatRoom = NovasamaChatRoom;
 
-/** Result of registering a chat room (`"New" | "Exists"`). Re-exported from `@novasamatech/product-sdk`. */
+/** Result of registering a chat room (`"New" | "Exists"`). Re-exported from `@novasamatech/host-api-wrapper`. */
 export type ChatRoomRegistrationResult = NovasamaChatRoomRegistrationResult;
 
-/** Result of registering a bot (`"New" | "Exists"`). Re-exported from `@novasamatech/product-sdk`. */
+/** Result of registering a bot (`"New" | "Exists"`). Re-exported from `@novasamatech/host-api-wrapper`. */
 export type ChatBotRegistrationResult = NovasamaChatBotRegistrationResult;
 
-/** Renderer callback for custom message types. Re-exported from `@novasamatech/product-sdk`. */
+/** Renderer callback for custom message types. Re-exported from `@novasamatech/host-api-wrapper`. */
 export type ChatCustomMessageRenderer = NovasamaChatCustomMessageRenderer;
 
-/** Parameters passed to a {@link ChatCustomMessageRenderer}. Re-exported from `@novasamatech/product-sdk`. */
+/** Parameters passed to a {@link ChatCustomMessageRenderer}. Re-exported from `@novasamatech/host-api-wrapper`. */
 export type ChatCustomMessageRendererParams<T = Uint8Array> =
     NovasamaChatCustomMessageRendererParams<T>;
 
@@ -55,14 +55,14 @@ export type ChatCustomMessageRendererParams<T = Uint8Array> =
  * registration.
  *
  * Type identical to `createProductChatManager()` from
- * `@novasamatech/product-sdk`.
+ * `@novasamatech/host-api-wrapper`.
  */
 export type ChatManager = ReturnType<typeof createProductChatManager>;
 
 /**
  * Get the host chat manager.
  *
- * Returns the chat manager from `@novasamatech/product-sdk`, or `null` if
+ * Returns the chat manager from `@novasamatech/host-api-wrapper`, or `null` if
  * the package is unavailable (running outside a host container or the
  * optional peer dep isn't installed).
  *
@@ -81,7 +81,7 @@ export type ChatManager = ReturnType<typeof createProductChatManager>;
  */
 export async function getChatManager(): Promise<ChatManager | null> {
     try {
-        const sdk = await import("@novasamatech/product-sdk");
+        const sdk = await import("@novasamatech/host-api-wrapper");
         return sdk.createProductChatManager();
     } catch (err) {
         log.debug("getChatManager unavailable", err);
@@ -93,7 +93,7 @@ export async function getChatManager(): Promise<ChatManager | null> {
  * Dispatch helper that composes multiple custom-message renderers into a
  * single {@link ChatCustomMessageRenderer} keyed by `messageType`.
  *
- * Mirrors `matchChatCustomRenderers` from `@novasamatech/product-sdk`
+ * Mirrors `matchChatCustomRenderers` from `@novasamatech/host-api-wrapper`
  * inline (the upstream implementation is pure dispatch logic with no
  * transport / runtime dependency on Novasama), so callers get the same
  * sync signature instead of an async-with-null wrapper.
--- a/product-sdk/packages/host/src/chat.ts
+++ b/product-sdk/packages/host/src/chat.ts
@@ -1,0 +1,119 @@
+/**
+ * Wrapper for the host's chat surface (`host_chat_*` family).
+ *
+ * Shipped flat-in-host rather than as `getTruApi().chat.*` (the shape
+ * sketched in issue #93) because the upstream JS `hostApi` is itself a
+ * flat object - there is no `.chat` accessor to mirror. A flat
+ * `getChatManager()` matches the pattern already used by
+ * {@link getThemeProvider}, {@link getAccountsProvider}, and
+ * {@link getStatementStore}; if a namespaced view is desirable later, it
+ * can be layered on top without breaking this surface.
+ *
+ * @module
+ */
+
+import type {
+    ChatBotRegistrationResult as NovasamaChatBotRegistrationResult,
+    ChatCustomMessageRenderer as NovasamaChatCustomMessageRenderer,
+    ChatCustomMessageRendererParams as NovasamaChatCustomMessageRendererParams,
+    ChatMessageContent as NovasamaChatMessageContent,
+    ChatReceivedAction as NovasamaChatReceivedAction,
+    ChatRoom as NovasamaChatRoom,
+    ChatRoomRegistrationResult as NovasamaChatRoomRegistrationResult,
+    createProductChatManager,
+} from "@novasamatech/product-sdk";
+
+/** Chat message payload variants. Re-exported from `@novasamatech/product-sdk`. */
+export type ChatMessageContent = NovasamaChatMessageContent;
+
+/** Action received via {@link ChatManager.subscribeAction}. Re-exported from `@novasamatech/product-sdk`. */
+export type ChatReceivedAction = NovasamaChatReceivedAction;
+
+/** Room metadata delivered to {@link ChatManager.subscribeChatList}. Re-exported from `@novasamatech/product-sdk`. */
+export type ChatRoom = NovasamaChatRoom;
+
+/** Result of registering a chat room (`"New" | "Exists"`). Re-exported from `@novasamatech/product-sdk`. */
+export type ChatRoomRegistrationResult = NovasamaChatRoomRegistrationResult;
+
+/** Result of registering a bot (`"New" | "Exists"`). Re-exported from `@novasamatech/product-sdk`. */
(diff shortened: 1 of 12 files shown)


product-sdk/packages/host/src/entropy.ts
```diff
@@ -0,0 +1,68 @@
...
+export async function deriveEntropy(key: Uint8Array): Promise<Uint8Array> {
+    const truApi = await getTruApi();
+    if (!truApi) {
+        throw new Error("deriveEntropy: TruAPI unavailable");
+    }
+    log.debug("deriveEntropy", { keyLen: key.length });
+
+    return await truApi.deriveEntropy(enumValue("v1", key)).match(
+        (envelope: { tag: "v1"; value: Uint8Array }) => envelope.value,
+        (err: unknown) => {
```
Imod7, line 46: The error callback receives the versioned envelope `{ tag: "v1", value: CodecError<...> }`, not the `CodecError` directly. So `err instanceof Error` is `false` and `String(err)` produces "[object Object]", losing the host's actual error. Same pattern in `requestDevicePermission` (permissions.ts:80-87) and `requestPermission`.

product-sdk/packages/host/src/container.ts
```diff
@@ -31,7 +32,29 @@ export async function getHostLocalStorage(): Promise<HostLocalStorage | null> {
...
+ * custom transport. Use this when you need a non-default transport (e.g.
+ * for tests); otherwise prefer {@link getHostLocalStorage}, which returns
+ * the shared singleton.
+ *
+ * Mirrors `createLocalStorage` from `@novasamatech/product-sdk`.
+ *
+ * @param transport - Optional transport; defaults to the sandbox transport.
+ * @returns A new `HostLocalStorage` instance, or `null` if unavailable.
+ */
+export async function createHostLocalStorage(
```
Imod7, line 56: `getHostLocalStorage` (L30) gates on `isInsideContainer()` but `createHostLocalStorage` doesn't. Outside a container the import still resolves, so this returns a non-null handle whose first call will hang or throw on the missing transport. Same situation in `createHostPreimageManager` (truapi.ts:217).

product-sdk/packages/host/src/chat.ts
```diff
@@ -0,0 +1,119 @@
...
+ *   await chat.registerBot({ botId: "echo", name: "Echo Bot", icon: "" });
+ *   chat.subscribeAction((action) => { ... });
+ * }
+ * ```
+ */
+export async function getChatManager(): Promise<ChatManager | null> {
+    try {
+        const sdk = await import("@novasamatech/product-sdk");
+        return sdk.createProductChatManager();
+    } catch {
```
Imod7: nit: all five new getters use `try { ... } catch { return null }`, so "peer dep not installed" and "factory threw" both surface as `null`. Suggestion: Add a `log.debug` in the catch so genuine failures leave a trace. Same pattern in theme.ts:65, container.ts:55, and truapi.ts:220.

product-sdk/packages/host/src/entropy.ts
decrypto21: Addressed in `668698d`

product-sdk/packages/host/src/container.ts
decrypto21: Addressed in `668698d`

product-sdk/packages/host/src/chat.ts
decrypto21: Addressed in `668698d`

Imod7 (approved):

TarikGul:
Let's resolve the merge conflicts then I think this is good to go!

PR #152, opened 2026-05-29
fix(contracts): use pallet-revive account as query fallback origin

pgherveou:
## What

Replace the Alice dev account fallback origin used for read-only contract query dry-runs with pallet-revive's own pallet account.

## Why

Using `//Alice` as the implicit query origin when no wallet is connected is misleading and ties query behavior to a dev seed. pallet-revive's pallet account is a stable, semantically meaningful origin that always exists on chain.

## How

This mirrors `Pallet::<T>::account_id()` in pallet-revive:

```rust
PalletId(*b"py/reviv").into_account_truncating()
```

`into_account_truncating` builds the 32-byte `AccountId32` as the PalletId `TYPE_ID` (`b"modl"`) followed by the id (`b"py/reviv"`), zero padded — i.e. the bytes `"modlpy/reviv"` + 20 trailing zero bytes. SS58-encoding that public key (default prefix 42) yields:

```
5EYCAe5ijiYfhaAUBd6H9WGRTsvwFFc7GnhQkiHvBYxdvpbV
```

The bytes are encoded in TS rather than hardcoding the address string, so it stays self-documenting and verifiably matches the Rust definition.

- Compute `QUERY_FALLBACK_ORIGIN` via `ss58Address(...)` instead of `seedToAccount(DEV_PHRASE, "//Alice")`
- Drop the now-unused `seedToAccount` / `DEV_PHRASE` imports
- Update the `log.warn` message and related comments

## Test

`pnpm test` in `packages/contracts` — all 103 tests pass.


kianenigma (approved):

UtkarshBhardwaj007 (approved):
LGTM

PR #159, opened 2026-06-01
feat: add @parity/product-sdk-auth — shared QR/mobile sign-in component

EnderOfWorlds007:
## What

New **`@parity/product-sdk-auth`** package: the QR/mobile sign-in + session-signing glue, lifted from playground-cli's `src/utils/{auth,signer,sessionSigner}.ts` + the RFC-0010 allocation helper, refactored so all env config (DAPP_ID, product id, metadata URL, People endpoints) is **injected via `createAuthClient(config)`** instead of importing playground's `config.ts`. Goal: one shared sign-in implementation for playground-cli and bulletin-deploy (and future product CLIs).

Public API: `createAuthClient` / `resolveSigner` / RFC-0010 `requestResourceAllocation` + a `./ui` subpath (QR render + login/logout status formatters). **18 unit tests pass** (session signer incl. the wallet-vs-product-account regression guard, allocations, status formatters, auth-client shape).

## ⚠️ Build issues we hit (flagging for you — mostly pre-existing, not from this package)

Building this in the monorepo surfaced a few problems:

1. **`json-rpc-provider` override is stranded (FIXED in this PR).** pnpm 11 no longer reads `pnpm.overrides` from `package.json` (`[WARN] … pnpm.overrides ignored`), so the existing `^0.2.0` override was a silent no-op. A stale transitive `@polkadot-api/substrate-client@0.1.4` (via old `host-api-wrapper@0.7.9`) then kept `json-rpc-provider@0.0.1` (no `isResponse`) alongside the `0.2.0` `host-papp`/`terminal` need → `SyntaxError: … does not provide an export named 'isResponse'` loading terminal. Fix here: moved the override to `pnpm-workspace.yaml` + clean reinstall → collapses to a single `0.2.0`.

2. **`product-sdk-terminal@0.2.1` ↔ `host-papp@0.7.9` (blocks `.d.ts`).** `packages/terminal/src/index.ts` imports `AttestationStatus` from `@novasamatech/host-papp`, but host-papp@0.7.9's type declarations don't export it → terminal's DTS build fails (`TS2305`) → cascades to block dependents' types. Needs a terminal/host-papp version reconcile (your call).

3. **`packages/tx` emits no build** (`dist/index.js` missing) → vite can't resolve `@parity/product-sdk-tx` for tests.

(1) is fixed in this PR; (2) and (3) are yours to reconcile.

## Note

bulletin-deploy currently **vendors** this code behind a one-file facade so it can ship without waiting on this package being published (paritytech/bulletin-deploy#791). Once you publish `@parity/product-sdk-auth`, bulletin-deploy swaps the vendored copy for the package in a single file.


ReinhardHatko:
Great to see this extracted — d3pot is building the same QR-login glue and will consume this. Two bits of feedback from porting it downstream (a code review surfaced both):

1. **QR-timeout timer leak** (`auth.ts`, the 60s `Promise.race` around L201–223): the `setTimeout` is never `clearTimeout`'d when the QR payload wins the race, so the timer stays armed and keeps the Node event loop alive — a CLI command can hang for up to 60s after login already completed. Capturing the handle and clearing it after the race (and in a `finally`) fixes it.

2. **Self-deriving the product key**: `sessionSigner.ts` hand-derives `deriveProductAccountPublicKey(rootAccountId, productId, index)` and passes it as `ref.publicKey` — as do playground-cli and bulletin-deploy. Would it make sense for `product-sdk-terminal`'s `createSessionSignerForAccount` to self-derive from `session.rootAccountId` when `publicKey` is omitted, instead of falling back to the selected account (which yields an invalid signature for a non-selected product)? That'd let this package and every consumer drop the derivation. Happy to open that small change against `product-sdk-terminal` if it'd help.

Thanks for landing this!

TarikGul:
Now that we have time, will give this priority soon! 🙏

TarikGul:
Going to take over this PR

product-sdk/packages/auth/package.json
```diff
@@ -0,0 +1,41 @@
+{
+    "name": "@parity/product-sdk-auth",
+    "version": "0.1.0",
+    "description": "Authentication and session signing for Polkadot CLI apps — QR/mobile sign-in via host-papp",
```
valentinfernandez1: host-papp is no longer used here so this mention, alongside the added dependency and a bunch of the code docs that mention it are no longer relevant

product-sdk/packages/auth/src/auth.ts
```diff
@@ -0,0 +1,454 @@
...
+                    onStatus({ step: "success", address, addresses });
+                } else {
+                    onStatus({
+                        step: "error",
+                        message: "Login succeeded but the local session was not available",
+                    });
+                }
+            }
+        } finally {
+            unsubPairing();
```
valentinfernandez1, line 321: The adapter isn't needed after login so we can destroy it to avoid it from leaking
```suggestion
            unsubPairing();
            adapter.destroy().catch(() => {});
```

product-sdk/pending-changesets/product-sdk-auth.md
```diff
@@ -0,0 +1,8 @@
+---
+"@parity/product-sdk-auth": minor
+"@parity/product-sdk": minor
```
valentinfernandez1: `product-sdk` doesn't import `product-sdk-auth` so I don't think it should be bumped
```suggestion
```

product-sdk/packages/auth/src/signer.ts
```diff
@@ -0,0 +1,115 @@
...
+        this.name = "SignerNotAvailableError";
+    }
+}
+
+const DEV_NAMES: readonly DevAccountName[] = ["Alice", "Bob", "Charlie", "Dave", "Eve", "Ferdie"];
+
+export function parseDevAccountName(suri: string): DevAccountName | null {
+    const name = suri.replace(/^\/\//, "");
+    return DEV_NAMES.find((n) => n.toLowerCase() === name.toLowerCase()) ?? null;
+}
```
valentinfernandez1, line 73: While `//Alice` is valid here, `Alice` is also valid. Is this intended?

valentinfernandez1 (changes requested):
I think the PR is in the right direction, one thing that I am not a fan of is that this package and `@parity/product-sdk-terminal` both implement a type of sign in flow, we should reconcile them instead of maintaining two similar solutions

product-sdk/packages/auth/package.json
TarikGul: Addressed thanks!

product-sdk/packages/auth/src/auth.ts
TarikGul: Addressed thanks

product-sdk/pending-changesets/product-sdk-auth.md
TarikGul: Good catch thanks

product-sdk/packages/auth/src/signer.ts
TarikGul: I made this more consistent with the rest of the sdk so `//<Name>` is required, no leniency.

TarikGul:
> I think the PR is in the right direction, one thing that I am not a fan of is that this package and @parity/product-sdk-terminal both implement a type of sign in flow, we should reconcile them instead of maintaining two similar solutions

@valentinfernandez1 totally agree with this, looks like there is some duplication with terminal. Especially with `createSessionSigner` etc. Going to look into some solutions.

TarikGul:
@valentinfernandez1 

Session signer - deleted auth's copy, now delegates to terminal's createSessionSignerForAccount. (The older impl was actually using a PJS-shim style path -- not good -- changed it to use terminals session.createTransaction style)

Resource allocation - extracted a shared sendResourceAllocation primitive in terminal and exported it


We should be good now :)

product-sdk/packages/auth/src/auth.ts
```diff
@@ -0,0 +1,457 @@
...
+                    unsub = adapter.sso.pairingStatus.subscribe(async (status: PairingStatus) => {
+                        if (status.step === "pairing" && !done) {
+                            done = true;
+                            unsub?.();
+                            resolve(await renderQrCode(status.payload));
+                        }
+                    });
+                }),
+                new Promise<never>((_, reject) =>
+                    setTimeout(
```
Imod7: Mentioned also in Reinhard's comment [here](https://github.com/paritytech/product-sdk/pull/159#issuecomment-4908592173)
> _This setTimeout is never clearTimeout'd when the QR promise wins the race, so the timer stays armed and keeps the Node event loop alive - the CLI can hang up to 60 s after login already completed. Capture the handle and clear it in a finally._

product-sdk/packages/auth/src/auth.ts
```diff
@@ -0,0 +1,457 @@
...
+     * kicks off authenticate(), waits for the QR payload, and returns the QR
+     * code + a handle to await the auth result.
+     */
+    async function connect(): Promise<ConnectResult> {
+        const adapter = createAdapter();
+
+        const sessions = await waitForSessions(adapter);
+        if (sessions.length > 0) {
+            const addresses = deriveSessionAddresses(sessions[0]);
+            return { kind: "existing", address: addresses.productAddress, addresses };
```
Imod7: The `existing` branch returns without destroying the adapter opened at `L196`, and `ConnectResult.existing` gives the caller no handle to destroy it - so this WebSocket leaks on the normal already-logged-in path (and again if `deriveSessionAddresses` throws on a stale session). Destroy the adapter before returning, e.g. in a try/finally. Same teardown-on-throw gap exists in `getSessionSigner` (L329-330).

product-sdk/packages/auth/src/auth.test.ts
```diff
@@ -0,0 +1,49 @@
+// Copyright (C) Parity Technologies (UK) Ltd.
```
Imod7, line 1: nice to have: `connect/waitForLogin/getSessionSigner/waitForLogout` are untested - which is where the two adapter/timer leaks live. A fake `TerminalAdapter` would let you regression-guard "timer cleared on QR win" and "adapter destroyed on existing-session + on derivation throw".

Imod7 (approved):

product-sdk/packages/auth/src/auth.ts
TarikGul: Addressed :)

product-sdk/packages/auth/src/auth.ts
TarikGul: Addressed :)

product-sdk/packages/auth/src/auth.test.ts
TarikGul: addressed :)

TarikGul:
NOTE the package still needs to be added to npm publish automation.

product-sdk/packages/auth/src/sessionSigner.ts
```diff
@@ -0,0 +1,94 @@
...
+import { createSessionSignerForAccount } from "@parity/product-sdk-terminal";
+import type { PolkadotSigner } from "polkadot-api";
+import { deriveProductAccountPublicKey } from "@parity/product-sdk-keys";
+
+export interface ProductAccountRef {
+    productId: string;
+    derivationIndex: number;
+}
+
+export const INCOMPLETE_SESSION_MESSAGE =
```
valentinfernandez1: dup on terminal

product-sdk/packages/auth/src/sessionSigner.ts
```diff
@@ -0,0 +1,94 @@
...
+ * `ProductAccountRef.publicKey` — so `createSessionSigner` here derives that key
+ * from the session root and delegates the rest.
+ */
+
+import type { UserSession } from "@parity/product-sdk-terminal";
+import { createSessionSignerForAccount } from "@parity/product-sdk-terminal";
+import type { PolkadotSigner } from "polkadot-api";
+import { deriveProductAccountPublicKey } from "@parity/product-sdk-keys";
+
+export interface ProductAccountRef {
```
valentinfernandez1: dup on terminal

product-sdk/packages/auth/src/sessionSigner.ts
```diff
@@ -0,0 +1,94 @@
...
+    productId: string;
+    derivationIndex: number;
+}
+
+export const INCOMPLETE_SESSION_MESSAGE =
+    'Stored login session is missing the root account public key. Run "logout" and then "login" to pair again.';
+
+export function sessionRootPublicKey(session: UserSession): Uint8Array {
```
valentinfernandez1: Dup on terminal

valentinfernandez1 (commented):
There is still a good overlap between `terminal` and `auth` packages. 

I am fine if we want to re export functions from `terminal` from `auth` but let's not implement them twice.

product-sdk/packages/auth/src/sessionSigner.ts
TarikGul: Apologies for missing, these - fixing them as well.

product-sdk/packages/auth/src/sessionSigner.ts
TarikGul: fixed

product-sdk/packages/auth/src/sessionSigner.ts
TarikGul: fixed

TarikGul:
So the duplication is fixed @valentinfernandez1 but the only one left is `DEFAULT_STORAGE_DIR`, but that should probably be done in a follow up PR since it touched multiple places.

valentinfernandez1 (approved):
Great job @TarikGul @EnderOfWorlds007 🚀

PR #164, opened 2026-06-02
fix(host): surface ChainNotSupportedError instead of hanging on unsupported chains

### LINKED ISSUE #160: Paseo Asset Hub queries hang indefinitely: host silently drops provider instead of surfacing an error
## Environment

* `@parity/product-sdk` `0.7.0`
* `@parity/product-sdk-chain-client` `0.5.1`
* `@parity/product-sdk-descriptors` `0.5.1`

## Steps to Reproduce

1. Run a web app inside the Polkadot Desktop host container.
2. Connect using `createChainClient` with `paseo_asset_hub` and `paseo_bulletin`.
3. Call any query on `client.assetHub`, for example:

```ts
client.assetHub.query.System.Number.getValue();
```

## Expected

Either:

* The query resolves with chain data, or
* The SDK surfaces a clear error (thrown or via a `Result` type) indicating that the chain is not supported by the current host.

## Actual

The query never resolves or rejects. The browser console logs:

```text
Provider for chain 0xbf0488dbe9daa1de1c08c5f743e26fdc2a4ecd74cf87dd1b4b1eeb99ae4ef19f was not started because Host doesn't support it
```

All `assetHub` calls hang indefinitely with no way to detect failure without implementing an external timeout.

The Paseo Bulletin chain is unaffected. For example:

```ts
client.bulletin.query.TransactionStorage.ByteFee.getValue();
```

returns data normally.

## Analysis

`createPapiProvider` calls `checkIfReady()`, which sends a `host_feature_supported` request for the chain's genesis hash.

When the host returns `false`—either because Paseo Asset Hub is not enabled in the current Desktop build, or because the genesis hash in the descriptor (`0xbf0488...`) no longer matches what the Desktop application recognizes after a network reset—`checkIfReady()` resolves to `false`.

In that case, `getSyncProvider()` invokes `onResult()` with a no-op provider whose `send()` method only logs the error and does nothing.

PAPI receives this no-op provider and waits indefinitely for JSON-RPC responses that will never arrive, causing every caller to remain suspended with no rejection, error propagation, or built-in timeout mechanism.

valentinfernandez1:
## Summary

When the host doesn't recognize a chain (not enabled in the current Desktop/Browser build, or a descriptor genesis hash that drifted after a network reset), the wrapper's createPapiProvider hands PAPI a no-op provider whose send() only logs. PAPI then waits forever for JSON-RPC responses that never arrive, so every query against that chain hangs with no rejection, error, or timeout.

getHostProvider now verifies host support (via the same host_feature_supported check the wrapper performs internally) before handing a provider to PAPI. Unsupported chains throw the new ChainNotSupportedError

closes #160


TarikGul (approved):
🚀

PR #165, opened 2026-06-02
refactor(chain-client): drop unused rpcs from ChainClientConfig

valentinfernandez1:
## Summary

Removes the `rpcs` field from `ChainClientConfig`. It was a **required but
unused** property: chain-client routes every connection through the host
provider (`getHostProvider` + genesis-hash matching), so the supplied RPC
endpoints were never read. The field only forced callers to construct and
pass a no-op argument.


product-sdk/examples/chain-client-demo/src/main.ts
```diff
@@ -9,7 +9,7 @@
  * Flow inside the host-api-test-sdk test host:
  *   1. SignerManager.connect() auto-detects -> HostProvider (product-sdk)
  *   2. Host responds with Bob's non-product account
- *   3. BYOD: createChainClient({ chains: { assetHub }, rpcs: {...} }) connects via host
+ *   3. BYOD: createChainClient({ chains: { assetHub } }) connects via host
```
pgherveou, line 12: Dq whats the D in BYOD

pgherveou (approved):

product-sdk/examples/chain-client-demo/src/main.ts
TarikGul: Build your own descriptors

TarikGul:
I think the Bundle CI failing here is totally fine since its related to the terminal package. We might actually just want to remove it from the bundle check all together.

TarikGul (approved):
Thanks 🚀

PR #163, opened 2026-06-02
chore(deps): bump shared catalog deps to latest 0.8.x / patch line

UtkarshBhardwaj007:
## What

Workspace-catalog dependency bumps to the latest within-range versions. **Dependency-range updates only — no source changes.**

| Catalog entry | From → To | Affected published packages |
|---|---|---|
| `@novasamatech/host-api`, `host-api-wrapper` | `^0.8.0` → `^0.8.3` | signer (optional deps); host/statement-store (dev/peer only) |
| `polkadot-api` | `^2.1.2` → `^2.1.5` | chain-client, cloud-storage, contracts, descriptors, host, keys, signer, statement-store, tx, umbrella |
| `@polkadot-labs/hdkd-helpers` | `^0.0.27` → `^0.0.30` | contracts, keys, tx |
| `viem` | `^2.46.2` → `^2.52.0` | contracts |
| `@parity/host-api-test-sdk` | `^0.8.6` → `^0.9.0` | examples (dev) — clears the last transitive `@novasamatech/host-api@0.7.9` |
| `tsup` / `@playwright/test` | `8.5.1` / `1.60.0` | dev tooling |

Also switches `keys`/`tx` `hdkd-helpers` and `utils`/root `tsup` to `catalog:` references.

## Scope / what's deferred

This is the **low-risk "rest" slice**. The `terminal` package's `@novasamatech/host-papp` **0.7.x → 0.8 migration carries breaking changes** (removed `./register` export + WASM patch, removed `AttestationStatus`, deprecated `metadataUrl`, session-codec mirror update) and is **intentionally deferred to a separate PR**. `terminal` is unchanged here.

As a result, `@novasamatech/*@0.7.x` still appears in the lockfile **only transitively via `terminal`** — full "zero 0.7.x" lands once the terminal PR follows.

## Changeset

One `patch` changeset covering every published package whose shipped dependency range changed (chain-client, cloud-storage, contracts, descriptors, host, keys, signer, statement-store, tx). The umbrella `@parity/product-sdk` and internal dependents auto-cascade at patch. No `minor` bumps this wave, so the umbrella-minor policy doesn't apply.

## Test plan

- `pnpm -r build` → exit 0
- `pnpm check` (biome) → exit 0
- All affected package test suites pass (keys, tx, contracts, signer, host, statement-store, cloud-storage, terminal)
[...]

PR #167, opened 2026-06-02
feat(terminal): route signing through createTransaction + product-account key

### LINKED ISSUE #80: product-sdk-terminal cannot sign Paseo v2 `Revive.map_account` because `AsPgas` is not representable through the PJS signing bridge

decrypto21:
## What

Migrates `@parity/product-sdk-terminal` to `@novasamatech/host-papp` 0.8 and reworks the signer:

- **Transaction signing routes through `session.createTransaction`** (the host-papp `CreateTransactionRequest`/`Response` SSO pair) instead of `signRaw({ tag: "Payload" })`. The paired wallet builds **and** signs the structured tx, so it can decode/display it and every signed extension (incl. `AsPgas`) survives — and the mobile app's `<Bytes>…</Bytes>` anti-phishing envelope no longer corrupts transaction payloads. Fixes #80.
- **The signer uses the product-account key**: `createSessionSigner(session, adapter, publicKey?)` / `ProductAccountRef.publicKey` accept the host-derived product-account sr25519 key, which PAPI stamps into the extrinsic and verifies against — instead of the wallet's selected/root account.
- Removes the obsolete `verifiablejs` WASM loader: drops the `/register` entrypoint + `postinstall` patch, and the `AttestationStatus` re-export (gone upstream in 0.8). `metadataUrl` is now a deprecated no-op.

## Why

Unblocks consumers routing through `createTransaction` to drop their hand-rolled PJS signer: the two requirements are (1) sign via `createTransaction`, and (2) sign with the **product** account, not the wallet's selected account. See #80 for the root `AsPgas`/PJS-bridge limitation.

## Verification

`build` + `pnpm check` (biome) + terminal tests **132/132** green. Rebased onto `main` (catalog 0.8.x landed via #163).

## Follow-up

A later PR will switch the product-account key to self-derivation via `@novasamatech/statement-store`'s `deriveProductAccountPublicKey` (triangle-js-sdks #197) and remove the selected-account fallback entirely — once that publishes.

## Tracking

Part of #168.


product-sdk/packages/terminal/src/adapter.ts
```diff
@@ -30,8 +30,6 @@ const log = createLogger("terminal");
 export interface TerminalAdapterOptions {
     /** Unique app identifier. Used as the storage namespace. */
     appId: string;
-    /** URL to the app's metadata JSON (name + icon), shown during pairing. */
```
TarikGul, line 33: In the changeset we mention we deprecate this, but it looks like we remove this completely. Best to update the wording in the changeset.

TarikGul (approved):
Looks solid - just that one nit.

product-sdk/packages/terminal/src/testing.ts
```diff
@@ -38,15 +38,21 @@ import {
...
+// extras); the order and the fixed `Bytes(65)` inner for the chat key must match
+// the upstream codec exactly or the real SsoSessionManager fails to decode.
 const storedUserSessionCodec = Struct({
     id: str,
     localAccount: LocalSessionAccountCodec,
     remoteAccount: RemoteSessionAccountCodec,
+    rootAccountId: AccountIdCodec,
+    identityAccountId: Option(AccountIdCodec),
+    identityChatPublicKey: Option(Bytes(65)),
 });
```
pgherveou: shouldn't this be pulled as a dependency rather than mirrored?

decrypto21: Host-papp only exports types + createPappAdapter (read/pair only), not the codec or a write helper. So the mirror is required - I've pinned it to the exported StoredUserSession type via satisfies (build-time) + the interop round-trip, so drift is caught. An upstream fix (export the codec or a write helper) is what would let us drop the mirror

UtkarshBhardwaj007 (approved):
LGTM

TarikGul (approved):
LGTM - we can ignore the bundle CI fail

PR #170, opened 2026-06-03
feat(signer): productAccount option on HostProvider for product-only flows

### LINKED ISSUE #169: Add a productAccount option to HostProvider so product apps don't all reimplement the same connect-and-swap subclass
**Related:**

- PR #96 (`fix(signer): route product-account signing through host_create_transaction`) — merged; `getProductAccountSigner` is now public and pinned to `createTransaction`. ✓
- PR #155 (`chore(deps): migrate to @novasamatech/host-api(-wrapper) v0.8`) — merged; tightened the legacy-empty check that turned this from "ergonomic gap" into a hard blocker for product-only apps.

## Summary

Every product app embedded in a Polkadot host (Desktop / Mobile) needs to sign with a **product account** (derived per-dapp via `dotNsIdentifier`), not the legacy/wallet accounts that `HostProvider.connect()` returns by default. Signing with a legacy account fails on the host side with:

```
SigningErr::Unknown {
  reason: "Account can't be derived from product account id"
}
```

The only way to get a working product-account signer today is a ~25-line subclass: override `connect()`, call `super.connect()` for the transport handshake, swallow `NoAccountsError` (see below), then call `this.getProductAccount(dotNsId)` and return a single-element array. Every product-account-only app on host-api ≥ 0.7.x writes the same one.

**Since `signer@0.5.0` + `host-api(-wrapper)@0.8` the subclass also has to swallow a hard error.** `HostProvider.connect()` now rejects with `NoAccountsError("host")` when the host returns an empty legacy list:

```ts
// packages/signer/src/providers/host.ts ~ line 461
if (rawAccounts.length === 0) {
    log.warn("host returned no accounts");
    return err(new NoAccountsError("host"));
}
```

For product-only apps an empty legacy list is **the expected state**, not an error. Without the swallow, `SignerManager` stays in `status: "disconnected"` with `selectedAccount: null` and every "are you signed in" UI gate fails silently. The swallow only works because `this.accountsProvider = provider` is set *before* the empty-check throws — an **undocumented internal ordering** that, if upstream ever moves, silently re-breaks every product-only app.

Two secondary issues compound it:

- Legacy accounts have a `name` field (the user's wallet name); product accounts don't get one back from the host. The naive subclass renders the user as a truncated address until each app also fetches the username separately and patches it in.
- The `super.connect()` path **always fetches `host_get_legacy_accounts`** even when the consumer only cares about product accounts. That call is rate-limited host-side and is the source of opaque `RequestCredentialsErr::Unknown` errors on slow startups — blocking connect even though the legacy result is immediately discarded.

## Current workaround in `paritytech/playground-app`

[src/utils/contracts.ts](https://github.com/paritytech/playground-app/blob/main/src/utils/contracts.ts) — current shape after the v0.5.0 regression fix and the PR #96 cleanup:

```ts
class ProductAccountHostProvider extends HostProvider {
  constructor(private readonly dotNsId: string, options?: HostProviderOptions) {
    super(options);
  }

  async connect(signal?: AbortSignal): Promise<Result<SignerAccount[], SignerError>> {
    // signer 0.5 / host-api 0.8 makes super.connect fail with NoAccountsError
    // when the host has no LEGACY accounts. Our flow only uses the per-app
    // product account, so swallow that specific error and proceed —
    // accountsProvider is set inside super.connect *before* the legacy-empty
    // check, so getProductAccount works regardless.
    const base = await super.connect(signal);
    let walletName: string | null = null;
    if (base.ok) {
      walletName = base.value.find(a => a.name)?.name ?? null;
    } else if (!(base.error instanceof NoAccountsError)) {
      return base;
    }
    const product = await this.getProductAccount(this.dotNsId);
    if (!product.ok) return product;
    const acc = product.value;
    const productAccount: ProductAccount = {
      dotNsIdentifier: this.dotNsId,
      derivationIndex: 0,
      publicKey: acc.publicKey,
    };
    return ok([{
      ...acc,
      name: acc.name ?? walletName,
      getSigner: () => this.getProductAccountSigner(productAccount),
    }]);
  }
}
```

Then plumbed through `SignerManager`'s `createProvider` hook:

```ts
new SignerManager({
  dappName: "playground-dot",
  createProvider: (type) =>
    type === "host"
      ? new ProductAccountHostProvider(PLAYGROUND_DOTNS_ID)
      : new HostProvider(),
});
```

## Proposed API — non-breaking

Add an optional `productAccount` config to `HostProviderOptions`:

```ts
interface HostProviderOptions {
  // ... existing options
  /**
   * If set, `connect()` returns a single product account derived for the
   * given `dotNsIdentifier` instead of the host's legacy account list.
   * The legacy fetch (`host_get_legacy_accounts`) is skipped entirely.
   *
   * `SignerAccount.name` is populated on a best-effort basis from
   * `accounts.getUserId().primaryUsername`. Failures (NotConnected,
   * PermissionDenied, codec drift) leave the name null and connect
   * still succeeds — callers fall back to whatever display rule they
   * use (truncated address, app-claimed username, etc.).
   *
   * Signing is routed through `getProductAccountSigner` pinned to
   * `createTransaction` (see PR #96) so unknown signed extensions
   * (e.g. Paseo Next's `AsPgas`, `AuthorizeValueTransfer`) survive
   * end-to-end.
   */
  productAccount?: {
    /** App identifier (e.g., `"playground.dot"`). */
    dotNsIdentifier: string;
    /** Derivation index within the app scope. Default: 0. */
    derivationIndex?: number;
  };
}
```

When `productAccount` is set, `HostProvider.connect()`:

1. Does its existing transport handshake (`createAccountsProvider`).
2. **Skips `host_get_legacy_accounts`** — that call is removed from this code path entirely.
3. Calls `host_account_get` for the product account.
4. Best-effort calls `accounts.getUserId()` for the user's primary username (failures logged, do not abort).
5. Requests `ChainSubmit` permission (unchanged).
6. Returns a single-element `SignerAccount[]` with the product account; `getSigner` is wired through `getProductAccountSigner(account, "createTransaction")`.

When `productAccount` is unset, behavior is exactly as today — current consumers see no change.

### Caller usage

```ts
new HostProvider({
  productAccount: { dotNsIdentifier: "myapp.dot" },
});
```

No subclassing, no `createProvider` swap, no explicit `getProductAccount` plumbing, no `NoAccountsError` workaround in user code.

### Out of scope for this PR (potential follow-ups)

- **`dotNsIdentifier` derivation from `window.location`.** Originally proposed (`localhost` → host, `<name>.dot.li` → strip `.li`, `<name>.dot` → as-is). Excluded from this PR to keep the surface minimal — apps already hand-roll this in `config.ts`. Lift into the SDK in a follow-up if there's appetite.
- **`name: false` opt-out for the `getUserId` call.** Useful for tests or for apps that explicitly want a nameless account. Trivially additive once needed.
- **Public `HostProvider.getUserId()` method.** Auto-fetch in `connect()` is best-effort and intentionally silent on failure — non-regressing for today's display chain (registry username → host name → truncated address). Apps that want to react to `PermissionDenied` explicitly (e.g., prompt the user to grant the identity permission) need a public method that returns a structured `Result<{ primaryUsername }, SignerError>`. Mirrors the existing public `getProductAccount` / `getProductAccountAlias` pattern. Strictly additive — small follow-up.
- **Sibling `ProductHostProvider` class.** Considered and rejected: leaves the legacy `HostProvider` as a class with no practical product-app consumers, and the option-on-existing-class shape is strictly additive / non-breaking. Future deprecation path stays clean — if "wallet-like products that enumerate legacy accounts" turns out to have no real consumers, `productAccount` can become the implicit default in a later major.

## Scope notes — what's out

- **No host-API changes.** Everything proposed uses methods already shipping in `@novasamatech/host-api` 0.7.x (`host_account_get`, `host_get_user_id`, `host_create_transaction`). Host-side schema stays the same.
- **No changes to `SignerManager`.** The `createProvider` hook stays as-is; the option flows through `HostProvider`'s constructor.
- **Multi-account product flows** stay routed through the existing `getProductAccount(dotNsIdentifier, derivationIndex)` method. The new option is for the common single-account case.

ReinhardHatko:
Apps that sign exclusively with a per-dapp derived product account can now pass `productAccount: { dotNsIdentifier }` to HostProvider, skipping the legacy-account fetch entirely. SignerAccount.name is populated best-effort from accounts.getUserId().primaryUsername (fetched in parallel with the account itself) so apps render "Hello, alice" instead of a truncated address; failures leave name null without aborting connect.

Also fixes the v0.5.0 regression where product-only apps were broken whenever the host returned no legacy accounts (super.connect rejected with NoAccountsError before any product-account fetch could happen).

Obsoletes the ~25-line `class extends HostProvider` workaround every product app was carrying.

No behaviour change for consumers that don't set the option.

Closes #169

### DIFF
--- a/product-sdk/packages/signer/src/providers/host.ts
+++ b/product-sdk/packages/signer/src/providers/host.ts
@@ -50,6 +50,21 @@ export interface HostProviderOptions {
     requestChainSubmitPermission?: boolean;
     /** @deprecated Renamed to `requestChainSubmitPermission`. */
     requestTransactionSubmitPermission?: boolean;
+    /**
+     * If set, `connect()` returns a single product account for the given
+     * `dotNsIdentifier`, skipping the legacy fetch entirely. For apps
+     * that sign exclusively with a per-dapp derived account.
+     *
+     * `SignerAccount.name` is populated best-effort from
+     * `accounts.getUserId().primaryUsername`; failures leave it null.
+     * Signing is pinned to `createTransaction` (see PR #96).
+     */
+    productAccount?: {
+        /** App identifier (e.g., `"playground.dot"`). */
+        dotNsIdentifier: string;
+        /** Derivation index within the app scope. Default: 0. */
+        derivationIndex?: number;
+    };
 }
 
 /**
@@ -134,6 +149,7 @@ export interface AccountsProvider {
         dotNsIdentifier: string,
         derivationIndex?: number,
     ) => NeverthrowResultAsync<ContextualAlias, unknown>;
+    getUserId: () => NeverthrowResultAsync<{ primaryUsername: string }, unknown>;
     createRingVRFProof: (
         dotNsIdentifier: string,
         derivationIndex: number,
@@ -195,6 +211,7 @@ export class HostProvider implements SignerProvider {
     private readonly loadSdk: () => Promise<ProductSdkModule>;
     private readonly loadHostApiEnum: () => Promise<HostApiEnumHelper>;
     private readonly requestChainSubmitPermission: boolean;
+    private readonly productAccount: HostProviderOptions["productAccount"];
 
     private accountsProvider: AccountsProvider | null = null;
     private statusCleanup: (() => void) | null = null;
@@ -212,6 +229,7 @@ export class HostProvider implements SignerProvider {
             options?.requestChainSubmitPermission ??
             options?.requestTransactionSubmitPermission ??
             true;
+        this.productAccount = options?.productAccount;
     }
 
     async connect(signal?: AbortSignal): Promise<Result<SignerAccount[], SignerError>> {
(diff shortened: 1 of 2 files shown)


TarikGul (approved):
Thank you!

PR #175, opened 2026-06-04
fix(terminal): derive slot-account pubkey from canonical sr25519 to_bytes form

### LINKED ISSUE #176: createSlotAccountSigner derives the wrong public key from 64-byte slot account keys
## Summary
`createSlotAccountSigner` in `@parity/product-sdk-terminal` derives an incorrect public key from a 64-byte `slotAccountKey`. The resulting address and signatures don't match the slot account the host allocated, so submissions signed by the slot signer are rejected.

## Cause
Hosts return `slotAccountKey` as schnorrkel `SecretKey::to_bytes()` material: the canonical scalar (32 bytes) concatenated with the nonce (32 bytes).

`buildKeypair`'s 64-byte branch passes those bytes straight into `@scure/sr25519`'s `getPublicKey`. But `@scure` expects the `to_ed25519_bytes()` form, where the scalar is pre-multiplied by the curve cofactor (×8) so that its internal `decodeScalar` (a `>> 3` shift) recovers the real scalar. The canonical `to_bytes()` form is not cofactor-multiplied, so the scalar `@scure` reconstructs is off by a factor of 8 - producing a public key (and therefore address and signatures) that don't correspond to the actual slot account.

The 32-byte mini-secret path is unaffected; only the 64-byte branch is wrong.

## Impact
Any terminal/CLI product that receives a 64-byte slot key (Bulletin allowance, statement-store slot) and signs locally via the slot signer will derive the wrong account and have its submissions rejected.

## Fix
Convert the canonical scalar to the cofactor-multiplied (×8) form before deriving, so the derived key matches the host's slot account.

decrypto21:
## What

`createSlotAccountSigner` derived the wrong public key (and therefore wrong address + signatures) from a 64-byte slot account key, so submissions signed by the slot signer were rejected by the host.

Closes https://github.com/paritytech/product-sdk/issues/176

## Why

Mobile hosts return `slotAccountKey` as schnorrkel `SecretKey::to_bytes()` — canonical scalar (32 bytes) + nonce (32 bytes). See Android's `SlotAccountKey.kt`: *"32-byte sr25519 private key concatenated with 32-byte nonce"*.

`buildKeypair`'s 64-byte branch fed those bytes straight into `@scure/sr25519`'s `getPublicKey`, which expects the `to_ed25519_bytes()` form — where the scalar is pre-multiplied by the cofactor (×8) so its `decodeScalar` (`>> 3`) recovers the real scalar. Canonical bytes skip that ×8, so the derived key was off by the cofactor.

## Fix

Convert the canonical scalar to the cofactor-multiplied form before deriving. The 32-byte mini-secret path is unchanged.

```diff
 if (secret.length === 64) {
-    return {
-        publicKey: sr25519.getPublicKey(secret),
-        sign: (data) => sr25519.sign(data, secret),
-    };
+    const ed25519Secret = canonicalSr25519SecretToEd25519Bytes(secret);
+    return {
+        publicKey: sr25519.getPublicKey(ed25519Secret),
+        sign: (data) => sr25519.sign(data, ed25519Secret),
+    };
 }

### DIFF
--- a/product-sdk/packages/terminal/src/host-signer.ts
+++ b/product-sdk/packages/terminal/src/host-signer.ts
@@ -16,6 +16,26 @@ import type { TerminalAdapter } from "./adapter.js";
 import { type CachedAllocation, loadCache, readCacheEntry } from "./host-cache.js";
 import type { AllocatableResource } from "./host.js";
 
+function bytesToNumberLE(bytes: Uint8Array): bigint {
+    let n = 0n;
+    for (let i = bytes.length - 1; i >= 0; i--) n = (n << 8n) | BigInt(bytes[i]);
+    return n;
+}
+function numberToBytesLE(value: bigint, length: number): Uint8Array {
+    const out = new Uint8Array(length);
+    for (let i = 0; i < length; i++) out[i] = Number((value >> BigInt(8 * i)) & 0xffn);
+    return out;
+}
+
+// schnorrkel to_bytes() (canonical scalar) -> to_ed25519_bytes() form @scure expects (×8 cofactor).
+function canonicalSr25519SecretToEd25519Bytes(secret: Uint8Array): Uint8Array {
+    const ed25519Scalar = numberToBytesLE(bytesToNumberLE(secret.subarray(0, 32)) << 3n, 32);
+    const out = new Uint8Array(64);
+    out.set(ed25519Scalar, 0);
+    out.set(secret.subarray(32, 64), 32);
+    return out;
+}
+
 // Wire encoding for slotAccountKey isn't pinned: mobile may send a
 // 32-byte mini-secret or a 64-byte expanded secret. Detect length and
 // route; any other length throws.
@@ -29,9 +49,10 @@ function buildKeypair(secret: Uint8Array): {
         return { publicKey: keypair.publicKey, sign: (data) => keypair.sign(data) };
     }
     if (secret.length === 64) {
+        const ed25519Secret = canonicalSr25519SecretToEd25519Bytes(secret);
         return {
-            publicKey: sr25519.getPublicKey(secret),
-            sign: (data) => sr25519.sign(data, secret),
+            publicKey: sr25519.getPublicKey(ed25519Secret),
+            sign: (data) => sr25519.sign(data, ed25519Secret),
         };
     }
     throw new Error(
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: product-sdk/packages/terminal/src/host-signer.ts [hunk 1]
```diff
@@ -16,6 +16,26 @@ import type { TerminalAdapter } from "./adapter.js";
...
+}
+
+// schnorrkel to_bytes() (canonical scalar) -> to_ed25519_bytes() form @scure expects (×8 cofactor).
+function canonicalSr25519SecretToEd25519Bytes(secret: Uint8Array): Uint8Array {
+    const ed25519Scalar = numberToBytesLE(bytesToNumberLE(secret.subarray(0, 32)) << 3n, 32);
+    const out = new Uint8Array(64);
+    out.set(ed25519Scalar, 0);
+    out.set(secret.subarray(32, 64), 32);
+    return out;
+}
```
pgherveou: can we check that we are not re-implementing something already defined else where (Claude loves to do that)

also can't we use `DataView` for this, or is there a reason not to use it (just asked AI for standard API to do these stuff)

```js
function bytesToNumberLE(bytes) {
  const view = new DataView(bytes.buffer);
  return view.getBigUint64(0, true); // true means little-endian
}

function numberToBytesLE(value, length) {
  const buffer = new ArrayBuffer(length);
  const view = new DataView(buffer);
  view.setBigUint64(0, value, true); // true means little-endian
  return new Uint8Array(buffer);
}
```

pgherveou (approved):

UtkarshBhardwaj007 (approved):
LGTM

product-sdk/packages/terminal/src/host-signer.ts
TarikGul: Yea this is reimplementing `numberToBytesLE / bytesToNumberLE` from @noble/curves.|

```
  v2.x (currently @noble/curves@2.0.1):

  import { numberToBytesLE, bytesToNumberLE } from "@noble/curves/utils.js";
  
  v1.9.x (legacy):

  import { numberToBytesLE, bytesToNumberLE } from "@noble/curves/abstract/utils.js";
```

The main thing is how much does this increase the bundle by, and if this code never changes then we can just reference where we got the code from. That being said we absolutely need to reference where the code comes from and give it credit.

TarikGul (approved):
See my comment above.

PR #186, opened 2026-06-08
refactor: replace @novasamatech with @parity/truapi

### LINKED ISSUE #211: product-sdk-host drags in deprecated @polkadot/api@16 stack (@substrate/connect@0.8.11 + node-domexception) → install warnings for consumers
## Summary

`@parity/product-sdk-host` transitively drags in the **legacy `@polkadot/api@16` stack**, which carries two **deprecated** packages. Consumers of any product-sdk-host-dependent package (e.g. `@parity/polkadot-app-deploy`, `bulletin-deploy`) see deprecation warnings on a plain install:

```
$ npm install -g @parity/polkadot-app-deploy@rc
npm warn deprecated node-domexception@1.0.0: Use your platform's native DOMException instead
npm warn deprecated @substrate/connect@0.8.11: versions below 1.x are no longer maintained
```

## Dependency chain (verified)

```
@parity/product-sdk-host
  → @novasamatech/host-api-wrapper@0.8.7
    → @polkadot/extension-inject@^0.63.1
      → @polkadot/api@^16          ← the heavy legacy stack
        → @polkadot/rpc-provider → @substrate/connect@0.8.11        (deprecated)
        → @polkadot/x-fetch → node-fetch@3 → fetch-blob → node-domexception   (deprecated)
```

`@polkadot/extension-inject` is a hard dependency of `@novasamatech/host-api-wrapper` (both 0.8.6 and 0.8.7), and it pulls the full `@polkadot/api@16` legacy stack — even though the host wrapper only needs a small slice of functionality.

## Impact

- **Cosmetic but consumer-facing**: every consumer doing `npm install` / `npm install -g` sees the warnings. No functional breakage.
- **Not fixable downstream.** We verified:
  - `node-domexception` has no successor (deprecation = "use native DOMException"); even `fetch-blob@4` hard-requires `node-domexception@^1`, so it can't be overridden away.
  - npm `overrides` do **not** reach `npm install -g` consumers — empirically confirmed (packed a package with `overrides: {"@substrate/connect":"2.1.8"}`, `npm install -g` from the tarball → `@substrate/connect` still resolved to `0.8.11`). Overrides only apply when the package is the root project, not when it's the install target of a global install.
- It regressed for our consumers when product-sdk was bumped to the 0.13.0 line (which brought `host-api-wrapper@0.8.7`).

## Ask

`@parity/product-sdk-host` (and/or `@novasamatech/host-api-wrapper`) should avoid pulling the full legacy `@polkadot/api@16` stack — e.g. depend only on the lighter `@polkadot/util` / `@polkadot/keyring` pieces actually used, or on `polkadot-api` (the modern stack) rather than `@polkadot/extension-inject`. That removes both deprecated transitives for all downstream consumers.

Filed from bulletin-deploy / polkadot-app-deploy maintenance (their issues #979 / #49).

valentinfernandez1:
## Summary

Replaces the third-party `@novasamatech/host-api-wrapper` (plus `host-api` and `sdk-statement`) with our in-house `@parity/truapi` client. The whole host-API surface now goes through truapi; neither `@parity/product-sdk-host` nor `@parity/product-sdk-signer` imports the wrapper at runtime anymore.

## What moved to `@parity/truapi`

- **Host API:** `getTruApi`, `requestPermission` / `requestDevicePermission`, `deriveEntropy`, `requestResourceAllocation`, host `localStorage`, `isInsideContainer`.
- **Statement store:** `getStatementStore`, `createProofAuthorized` (the RFC-10 sponsored path), `subscribe` / `submit`. `@parity/product-sdk-statement-store` no longer depends on `@novasamatech/sdk-statement`, and its statement types are derived from the truapi wire types now, so protocol changes flow through automatically.
- **Preimage:** `getPreimageManager` / `createHostPreimageManager`.
- **Theme / chat / payments:** `getThemeProvider`, `getChatManager`, `getPaymentManager`.
- **Notifications:** `getNotificationManager`, on `notifications.*` (`sendPushNotification` / `cancelPushNotification`).
- **Navigation / chain-spec / features / tx-broadcast:** `navigateTo` (`system.navigateTo`), `featureSupported` / `isChainSupported` (`system.featureSupported`), `getChainSpec` (`chain.getSpec*`), and `broadcastTransaction` / `stopTransaction` (`chain.*`). These thin wrappers keep their Promise-shaped signatures; only the underlying call moved off the flat `hostApi`.
- **Chain provider:** `getHostProvider`, the PAPI `JsonRpcProvider`. This one is a backport of the wrapper's `createPapiProvider` into a new `papi-provider` module, rebuilt on `chain.*` + `system.featureSupported`.
- **Accounts + signer:** `getAccountsProvider`, a new `accounts` module (backport of `createAccountsProvider`) on `account.*` + `signing.*`, including the PAPI `PolkadotSigner` factories. The signer's `HostProvider` now just consumes `host`'s `getAccountsProvider` and asks for `ChainSubmit` through `host`'s `requestPermission`, instead of importing the wrapper itself.

## Dependencies

Removed `@novasamatech/host-api`, `host-api-wrapper`, and `sdk-statement` everywhere they were declared: `host`, `signer`, the example demos, the catalog, and the `tsup` external lists. `@parity/truapi` is now a hard dependency of `host`, along with `@polkadot-api/json-rpc-provider` (for the PAPI provider) and `@polkadot-api/substrate-bindings` (the signer needs it to decode metadata).

`@parity/product-sdk-terminal` is out of scope here. It depends on a different `@novasamatech` family (`host-papp` / `statement-store` / `storage-adapter`) that has nothing to do with the host-api wrapper.

> **Before merge:** `@parity/truapi@0.3.1` is published and wired in. The local `.pnpmfile.cjs` link has been removed and the workspace resolves it from the registry via the catalog (`^0.3.1`); the lockfile points at the published package. The branch is caught up with `main` (merge, not rebase). The only remaining step is promoting the parked changeset (`pending-changesets/migrate-host-statement-store-to-truapi.md` → `.changeset/`), done in the dedicated release PR.

## Breaking change: the `getTruApi()` client shape

If you only use the high-level helpers (`getStatementStore`, `getPaymentManager`, `getThemeProvider`, `getChatManager`, `getNotificationManager`, `getHostProvider`, `getAccountsProvider`, `requestPermission`, …), nothing changes for you.

The raw client from `getTruApi()` is a different story. It went from the flat, versioned wrapper shape to truapi's namespaced client. Four things change at basically every call site:

1. Methods sit under a domain now. `truApi.devicePermission(…)` becomes `truApi.permissions.requestDevicePermission(…)`.
2. The `{ tag: "v1", value }` envelope is gone. Pass the request straight in; truapi handles versioning internally.
3. Requests and responses use named fields instead of tuples or a bare `res.value`. An account id is `{ dotNsIdentifier, derivationIndex }` (not `[dotNs, idx]`), a chain is `{ genesisHash }`, raw bytes are `{ bytes: "0x…" }`, and responses read `res.granted` / `res.supported` / `res.signature` / `res.outcomes`.
4. Errors follow truapi's shapes. `GenericError` is `{ reason }` and tagged errors are `{ tag, value? }` (previously `{ value: { name, … } }`).

```ts
// before
const r = await truApi.devicePermission({ tag: "v1", value: "Camera" });
r.match((res) => res.value, (err) => err.value.name);

// after
const r = await truApi.permissions.requestDevicePermission("Camera");
r.match((res) => res.granted, (err) => err.reason);
```

A few smaller ones worth calling out: `ChatMessageContent`'s `Text` variant wraps its string as `{ text }` now, chat custom-message rendering is gone (truapi has no equivalent yet), host-mode statement submission uses the RFC-10 sponsored path, and `PushNotificationError` is now the truapi `{ tag }` tagged union (`"ScheduleLimitReached"` / `"Unknown"`) — `push` / `cancel` reject with an `Error` carrying the host error as its `cause`, so branch on `(err as Error).cause?.tag` instead of `instanceof`.

The full cheat-sheet and worked examples are in the guide: [Migrating to `@parity/truapi`](https://github.com/paritytech/product-sdk/blob/vf/truapi-migration/docs/content/guides/migrating-to-truapi.mdx).

## Validation

I tested this against [host-playground](https://github.com/paritytech/host-playground), which has a harness in `src/lib/tests.ts` with one entry per host-API method. Each one calls the real method against the SDK inside a live host container and checks the result.

What I did:
1. Linked the working-tree `host`, `signer`, and `chain-client` packages into host-playground.
2. Updated every `getTruApi()` call in `tests.ts` to the new shapes (the same changes the guide describes: domain methods, no `{ tag: "v1" }` envelopes, object account ids, `{ bytes }` payloads, the new response and error fields).
3. Ran the whole harness.

Unit + type coverage in this repo: full `pnpm -r build`, `pnpm check`, and the package test suites are green (host, signer, statement-store).

closes #211


TarikGul:
This is ready for review?

valentinfernandez1:
> This is ready for review?

yes, just need to fix the e2e tests

product-sdk/packages/host/src/accounts.ts
```diff
@@ -0,0 +1,448 @@
...
+    ): HostSubscription;
+}
+
+/** Ensure a `0x` prefix on a hex string (PJS payload fields arrive with or without it). */
+function asHex(value: string): HexString {
+    return (value.startsWith("0x") ? value : `0x${value}`) as HexString;
+}
+
+/** Map a PJS `SignerPayloadJSON` onto the host's {@link HostSignPayloadData}. */
+function buildSigningPayloadFields(payload: SignerPayloadJSON): HostSignPayloadData {
```
TarikGul, line 150: We should take this opportunity to completely remove any connection to PJS types - and or api's

valentinfernandez1: As is the PR is fairly big, we can if you want but I would do it as a follow up PR

product-sdk/packages/host/src/accounts.ts
```diff
@@ -0,0 +1,448 @@
...
+        version: payload.version,
+    };
+}
+
+/** Build an {@link AccountsProvider} over a TruAPI client's `account` / `signing` domains. */
+function adaptAccountsProvider(client: TrUApiClient): AccountsProvider {
+    const account = client.account;
+    const signing = client.signing;
+
+    /** Build the product-account signer's PJS-bridge signing callbacks (deprecated `signPayload` mode). */
```
TarikGul, line 179: > (deprecated `signPayload` mode)

We can avoid anything like this when we remove any tie to PJS.

product-sdk/packages/host/src/accounts.ts
```diff
@@ -0,0 +1,448 @@
...
+ * so apps get their own account that the user controls but is scoped to the app.
+ */
+export interface ProductAccount {
+    /** App identifier (e.g., "mark3t.dot"). */
+    dotNsIdentifier: string;
+    /** Derivation index within the app scope. Default: 0 */
+    derivationIndex: number;
+    /** Raw public key (32 bytes). */
+    publicKey: Uint8Array;
+}
```
pgherveou: these types exists in the Rust specs already, we should not need to hand roll them here

product-sdk/packages/host/src/payments.ts
```diff
@@ -1,61 +1,96 @@
...
 
-/** Status of an in-flight payment request. Re-exported from `@novasamatech/host-api-wrapper`. */
-export type PaymentStatus = NovasamaPaymentStatus;
+/** Status of an in-flight payment request (`{ tag: "Processing" | "Completed" | "Failed" }`). Re-exported from `@parity/truapi`. */
+export type PaymentStatus = HostPaymentStatusSubscribeItem;
 
-/** Source for {@link PaymentManager.topUp}. Re-exported from `@novasamatech/host-api-wrapper`. */
-export type TopUpSource = NovasamaTopUpSource;
+/** Source for {@link PaymentManager.topUp} (`ProductAccount` / `PrivateKey` / `Coins`, hex keys). Re-exported from `@parity/truapi`. */
+export type TopUpSource = PaymentTopUpSource;
```
pgherveou: these re-export under a different export name add an unnecessary level of indirection imo

product-sdk/packages/host/src/transport.ts
```diff
@@ -0,0 +1,288 @@
...
+ *
+ * The third-party host-api-wrapper auto-detected the host sandbox and handed
+ * apps a ready-to-use `hostApi` singleton. `@parity/truapi` is lower level:
+ * the product must detect its environment (iframe vs. webview), build a
+ * transport provider, create the client, and run the `system.handshake`. This
+ * module centralises that bootstrap and exposes the resulting client through a
+ * lazy, cached singleton so the rest of `@parity/product-sdk-host` can call
+ * {@link getClient} without re-implementing the wiring.
+ *
+ * Ported from the truapi playground's `lib/transport.ts` reference.
```
pgherveou: maybe we should properly export that from truapi instead of porting?

product-sdk/packages/host/src/truapi.ts
```diff
@@ -3,273 +3,248 @@
...
- * (from `String(err)`) or a JSON-stringified envelope.
+ * Extract a human-readable message from a host-side error.
+ *
+ * TruAPI errors arrive already unwrapped from their versioned wire envelope, in
+ * one of a few shapes: the catch-all `GenericError` (`{ reason }`), a tagged
+ * variant carrying a reason (`{ tag, value: { reason } }`), or a unit tagged
+ * variant (`{ tag }`). For resilience it also still unwraps the legacy novasama
+ * envelope (`{ tag: "v1", value: { name, message } }`), since this is a public
+ * helper and the surfaces still on the novasama wrapper (PAPI provider, accounts)
+ * may surface that shape.
```
pgherveou: not sure about that,
this function should probably have better types

product-sdk/packages/host/src/truapi.ts
```diff
@@ -3,273 +3,248 @@
...
+ * the repeated `.match(ok, err => throw)` dance the host wrappers would otherwise
+ * each spell out.
+ */
+export function unwrapHostResult<T, E>(result: ResultAsync<T, E>, label: string): Promise<T> {
+    return result.match(
+        (value) => value,
+        (error: E) => {
+            throw new Error(`${label}: ${formatHostError(error)}`, { cause: error });
+        },
     );
```
pgherveou, line 111: this seems to be a counter productive to 
unwrap result to throw errors, aren't we better off with Result type rather than throwing opaque errors ?

product-sdk/packages/host/src/truapi.ts
```diff
@@ -3,273 +3,248 @@
...
 }
 
 // ─────────────────────────────────────────────────────────────────────────────
-// Helpers from @novasamatech/host-api (re-exported from @novasamatech/scale)
+// Enum / Result / hex helpers
+//
+// `@parity/truapi`'s generated client wraps the versioned wire envelope
+// internally, so most callers no longer build these by hand. They are kept as
+// part of the public surface (and used by the surfaces still on novasama) as
```
pgherveou: this smell like something that should be refactored ?

product-sdk/packages/host/src/truapi.ts
```diff
@@ -278,32 +253,28 @@ export async function getAccountsProvider(): Promise<AccountsProvider | null> {
 
 /**
  * Resource types requestable via {@link requestResourceAllocation}.
- * Derived from the upstream codec so variant renames surface as compile
+ * Re-exported from `@parity/truapi` so variant renames surface as compile
  * errors, not runtime failures.
```
pgherveou: same comment as before regarding re-export adding unnecessary level of indirection

product-sdk/packages/host/src/accounts.ts
```diff
@@ -0,0 +1,448 @@
...
+/**
+ * A contextual alias obtained from Ring VRF.
+ *
+ * Proves account membership in a ring without revealing which account.
+ */
+export interface ContextualAlias {
+    /** Ring context (32 bytes). */
+    context: Uint8Array;
+    /** The Ring VRF alias bytes. */
+    alias: Uint8Array;
```
filvecchiato: Mirrors truapi's `HostAccountGetAliasResponse` with `Uint8Array` instead of `HexString`. Worth a doc comment linking to the wire type.

product-sdk/packages/host/src/accounts.ts
```diff
@@ -0,0 +1,448 @@
...
+/**
+ * One of the user's existing wallet accounts, surfaced through the host and
+ * identified by its public key and an optional name. Contrast with
+ * {@link ProductAccount}, which is also user-controlled but derived by the
+ * host for a specific app rather than picked from the user's existing keys.
+ */
+export interface HostAccount {
+    publicKey: Uint8Array;
+    name?: string;
+}
```
filvecchiato: Same shape as truapi's `LegacyAccount` but with `Uint8Array`. If the rename is intentional, a note linking to the wire type would help.

product-sdk/packages/host/src/features.ts
```diff
@@ -3,36 +3,34 @@
...
  *
- * As of `host-api` v0.8 the only variant is `Chain`, carrying the chain's
- * `0x`-prefixed genesis hash. Modeled locally (rather than derived from an
- * upstream codec) because the protocol exposes the feature only inline; new
- * variants surface here as a widening of the union.
+ * The only variant today is `Chain`, carrying the chain's `0x`-prefixed genesis
+ * hash. Modeled with the hash inline (the truapi wire type nests it as
+ * `{ genesisHash }`); new variants surface here as a widening of the union.
  */
 export type Feature = { tag: "Chain"; value: HexString };
```
filvecchiato, line 29: Flattened version of truapi's `HostFeatureSupportedRequest`. Could re-export directly or note the reshaping.

filvecchiato (commented):
Type duplication review: flagging host-defined types that structurally duplicate truapi wire types.

Also: `ResultAsync<T, E>` in `truapi.ts:351` is a subset of neverthrow's `ResultAsync` which truapi already re-exports. Could import instead of redeclaring.

product-sdk/packages/host/src/types.ts
```diff
@@ -27,67 +23,92 @@ import type { Subscription } from "@novasamatech/host-api";
...
+/** Statement bundled with its required {@link StatementProof}. Re-exported from `@parity/truapi`. */
+export type SignedStatement = TruSignedStatement;
 
-/** Statement bundled with its {@link StatementProof}. Re-exported from `@novasamatech/host-api-wrapper`. */
-export type SignedStatement = NovasamaSignedStatement;
+/**
+ * Topic-based subscription filter. The host delivers statements that match
+ * either *all* of the listed topics (`matchAll`) or *any* of them (`matchAny`).
+ */
+export type StatementTopicFilter = { matchAll: Topic[] } | { matchAny: Topic[] };
```
filvecchiato, line 64: Reshapes truapi's `RemoteStatementStoreSubscribeRequest` from tagged union to field-discriminated. Worth a doc comment noting the wire counterpart.

product-sdk/packages/host/src/types.ts
```diff
@@ -27,67 +23,92 @@ import type { Subscription } from "@novasamatech/host-api";
...
- * Pages arrive sequentially. `isComplete` is `true` on the final page of a
- * subscription's initial backfill; subsequent pages contain new statements
- * as they appear on chain. `statements` is `SignedStatement[]` (typed,
- * not `unknown[]`).
+ * Pages arrive sequentially. `isComplete` is `false` while the host streams the
+ * historical backfill and `true` once it's done (and on every subsequent
+ * live-update page). `statements` is `SignedStatement[]`.
  */
-export type StatementsPage = NovasamaStatementsPage;
+export interface StatementsPage {
```
filvecchiato: Identical to truapi's `RemoteStatementStoreSubscribeItem`. Could be a direct re-export.

product-sdk/packages/host/src/accounts.ts
TarikGul: That is fine for me, as long as we merge this into a staging branch, and not main.

product-sdk/packages/statement-store/src/transport.ts
```diff
@@ -45,14 +43,7 @@ class HostTransport implements StatementTransport {
 
         try {
             const sub = this.store.subscribe(hostFilter, (page) => {
```
decrypto21, line 45: This never registers sub.onInterrupt(...), so host-side teardown (close/interrupt) never reaches onError - the stream silently dies and the consumer thinks it's still subscribed. The try/catch only covers sync setup

product-sdk/packages/host/src/papi-provider.ts
```diff
@@ -0,0 +1,653 @@
...
+            switch (method) {
+                case "chainHead_v1_follow": {
+                    const [withRuntime] = params as [boolean];
+                    const syntheticSubId = getNextSubId();
+                    // Hold the handle in a ref so the Stop branch can release the
+                    // host subscription BEFORE forwarding the event: the consumer's
+                    // substrate-client refollows synchronously inside the handler,
+                    // and the transport dedupes by (method, payload-hash) — unless
+                    // this dead subscription is gone first, the refollow shares it
+                    // and never reaches the host, so events stop flowing.
```
decrypto21: truapi's subscribeRaw mints a fresh requestId per call and doesn't dedupe by (method, payload-hash) - this comment describes the old transport. Worth updating so it doesn't mislead.

product-sdk/packages/host/src/papi-provider.ts
```diff
@@ -0,0 +1,653 @@
...
+                        );
+                    break;
+                }
+                case "chainHead_v1_storage": {
+                    const [followSubscriptionId, hash, items, childTrie] = params as [
+                        string,
+                        HexString,
+                        Array<{ key: HexString; type: string }>,
+                        HexString | undefined,
+                    ];
```
decrypto21, line 297: If items (or other positional params) is missing, items.map throws synchronously before the hostError(id) boundary and the request never settles → caller hangs. Could we wrap the switch in try/catch → sendJsonRpcError(id, ...)?

product-sdk/packages/host/src/accounts.ts
```diff
@@ -0,0 +1,448 @@
...
+        nonce: asHex(payload.nonce),
+        specVersion: asHex(payload.specVersion),
+        transactionVersion: asHex(payload.transactionVersion),
+        tip: asHex(payload.tip),
+        metadataHash: payload.metadataHash ? asHex(payload.metadataHash) : undefined,
+        // PJS types assetId as `number | object`; the host expects the encoded
+        // hex form, which is what PAPI produces at runtime. Passed through as
+        // the wrapper did.
+        assetId:
+            payload.assetId !== undefined ? (payload.assetId as unknown as HexString) : undefined,
```
decrypto21, line 166: Nit - Unchecked as unknown as HexString on the fee-asset field - worth a runtime guard or TODO(issue)

TarikGul:
As a note - going to help get this PR over the finish line.

product-sdk/packages/host/src/accounts.ts
TarikGul: Yea actually I worked through the process and will keep this to a follow up PR - there is a lot on the testing and e2e testing side that will bloat this PR.

product-sdk/packages/host/src/accounts.ts
TarikGul: Okay updated, no longer hand rolling these specific types (moving on to the next ones below), each type now derives from the `@parity/truapi` spec types.

The only divergence is the decoded fields: truapi carries them as `HexString` (the wire format), whereas the sdk surfaces decoded as UintArrays. So we keep the spec type and `Omit` only the byte field we decode, then add it back as a `Uint8Array`.

```typescript
  HostAccount     = Omit<LegacyAccount, "publicKey"> & { publicKey: Uint8Array }
  ProductAccount  = ProductAccountId & Omit<ProductAccount, "publicKey"> & { publicKey: Uint8Array }
  ContextualAlias = { [K in keyof HostAccountGetAliasResponse]: Uint8Array }
```

product-sdk/packages/host/src/payments.ts
TarikGul: updated.

product-sdk/packages/host/src/transport.ts
TarikGul: Some things are not exported from the truapi, and we will need them otherwise porting is the only option for now: 


  1. Environment detection
  - `isCorrectEnvironment(): boolean` — iframe / cross-origin-throw / `__HOST_WEBVIEW_MARK__` / `__HOST_API_PORT__` heuristics. (Synchronous, so it can gate hot paths.)
  - `isIframe(): boolean` — internal helper; may stay private if isCorrectEnvironment is enough.

  2. Provider construction / environment plumbing
  - `resolveHostOrigin(): string | null` — pins targetOrigin from `document.referrer` / `ancestorOrigins`; fails closed (returns null, never "*") because frames carry signed payloads + account ids. Security-sensitive — important this lives upstream
  rather than being re-implemented per consumer.
  - `waitForWebviewPort(timeoutMs?): Promise<MessagePort>` — polls `window.__HOST_API_PORT__` (default 20s timeout). Plus the `WEBVIEW_PORT_TIMEOUT_MS` default.
  - `createSandboxProvider(): Provider` — the iframe-vs-webview decision that selects + builds the right provider (the glue between detection and `createIframeProvider`/`createMessagePortProvider)`.

  3. Cached client lifecycle (the actual singleton)
  - `getClientSync(): TrUApiClient | null` — build-or-return-cached client, no handshake. null outside a container.
  - `getClient(): Promise<TrUApiClient | null>` — same, but runs system.handshake once (memoized on the promise) with a best-effort 5s timeout. Plus the `HANDSHAKE_TIMEOUT_MS` default.
  - `isReady(): Promise<boolean>` — handshake succeeded?
  - `disposeClient(): void `— tear down cached provider/transport/client (mainly for tests).
  
  4. Window-globals contract (shared type)
  - The HostWindow shape: `{ __HOST_WEBVIEW_MARK__?: boolean; __HOST_API_PORT__?: MessagePort }`. The injected-globals names are a host↔SDK contract; owning them in truapi prevents drift.


PR to address this in the truapi: https://github.com/paritytech/truapi/pull/234

product-sdk/packages/host/src/truapi.ts
TarikGul: Added types here, and a note on making them more foolproof upstream.

TarikGul:
```typescript
  export type HostError =
    | GenericError                                  // imported from @parity/truapi
    | { tag: string; value?: undefined }            // local
    | { tag: string; value: { reason: string } };   // local
```

product-sdk/packages/host/src/truapi.ts
TarikGul: Yea I think this along with the PJS signPayload api can be deferred to a follow up PR. I agree this is some technical debt brought on from the previous workflow, and is just really a DRY helper, but if it were to be correctly fixed in this PR it would touch A LOT of packages.

product-sdk/packages/host/src/transport.ts
TarikGul: Merged, we just need to release the new version of the truapi

product-sdk/packages/host/src/truapi.ts
TarikGul: I removed the dead novasama-era shims. This should resolve this issue.

product-sdk/packages/host/src/truapi.ts
TarikGul: Fixed.

product-sdk/packages/host/src/accounts.ts
TarikGul: handled above :)

product-sdk/packages/host/src/accounts.ts
TarikGul: Solved above :)

product-sdk/packages/host/src/features.ts
TarikGul: In this case noted the reshaping.

product-sdk/packages/host/src/types.ts
TarikGul: Fixed more Tru* aliases like earlier. Also added the doc note.

product-sdk/packages/host/src/types.ts
TarikGul: Fixed.

product-sdk/packages/host/src/truapi.ts
pgherveou: agreed let's create some legacy debt ticket and add that in there

product-sdk/packages/host/src/papi-provider.ts
pgherveou: to comment on that, I think it's fine to keep the truapi simple here, de-duping should be up to the app.
in rxjs for example you would the shareReplay operator for that for example and specify how you want to deal with past event.

decrypto21: Totally aligned. Just pointing out the same - this comment 
`// and the transport dedupes by (method, payload-hash)`
should drop the dedup rationale just for documentation. Doesn't affect functionality.

product-sdk/packages/host/src/accounts.ts
TarikGul: I think this will go away when we remove PJS signPayload

product-sdk/packages/host/src/transport.ts
TarikGul: This is now resolved.

product-sdk/packages/host/src/truapi.ts
TarikGul: Issues:

https://github.com/paritytech/product-sdk/issues/228
https://github.com/paritytech/product-sdk/issues/229

TarikGul (approved):
🚀

PR #196, opened 2026-06-09
feat(host): add getNotificationManager() wrapper

Imod7:
### Description

Adds `getNotificationManager()` to `@parity/product-sdk-host`: a flat wrapper over the host's `notificationManager` singleton (`push` / `cancel`), matching the `getPaymentManager` / `getPreimageManager` pattern. Also re-exports `PushNotificationError` and the derived `NotificationId` /
`PushNotificationInput` types.

### Why

It's the one host surface with no `@parity/product-sdk-*` facade: because no product has used host push-notifications until now (audited t3ams, linktr33, ja3x, sh33ts, sourc3s, mark3t: none use it). The w3s conference app is the first consumer; this lets it reach notifications without importing `@novasamatech/host-api(-wrapper)` directly.

### Changes

- New `src/notifications.ts` (wrapper + re-exports), wired into `src/index.ts`.
- Mirrors `src/payments.ts`: dynamic-import the peer, return the singleton, `catch → null`, inline vitest.
- `PushNotificationError` re-exported as a value so consumers can `instanceof`-branch on the host's pending-notification cap.
- Umbrella `@parity/product-sdk` picks it up automatically via `export *`.

### Test
- Verified locally: build, test (37, +2 notifications), `pnpm check`

### DIFF
--- a/product-sdk/packages/host/src/index.ts
+++ b/product-sdk/packages/host/src/index.ts
@@ -95,3 +95,11 @@ export type {
 // Payments (RFC-0006)
 export { getPaymentManager } from "./payments.js";
 export type { PaymentManager, PaymentBalance, PaymentStatus, TopUpSource } from "./payments.js";
+
+// Notifications
+export { getNotificationManager, PushNotificationError } from "./notifications.js";
+export type {
+    NotificationManager,
+    NotificationId,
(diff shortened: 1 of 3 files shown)


TarikGul (approved):
Nice and clean thanks

PR #199, opened 2026-06-10
docs(cloud-storage): update cloud-storage skill to the store()-builder API

Imod7:
### Description

The `product-sdk-cloud-storage` skill (`SKILL.md` + `references/cloud-storage-api.md`) documented a **pre-builder API that no released version exposes** (verified against both installed `0.5.3` and source `0.6.0`). Updated the stale examples, and the read-side described a standalone IPFS-gateway fallback that doesn't exist. This rewrites both files against the current source.

### Changes

- Switched every example to the current `store()` upload builder.
- Removed APIs that no longer exist: `upload`, `batchUpload`, `computeCid`, `cidExists`, `gatewayUrl`, `getGateway`.
- Made clear that reads only work inside a host container — no IPFS-gateway fallback.
- Fixed the environments list to `"paseo" | "summit"`.
- Brought the API-reference types and signatures in line with the source.

### Verification

- Docs/comments only - **no behavior change**. 
- Every documented symbol, signature, and enum value was checked against `packages/cloud-storage/src/{client,index,networks,types,verify,cid}.ts` and the `@parity/bulletin-sdk@0.3.0` d.ts.

### DIFF
--- a/product-sdk/packages/sdk/src/core/createApp.ts
+++ b/product-sdk/packages/sdk/src/core/createApp.ts
@@ -94,7 +94,9 @@ export async function createApp(config: AppConfig): Promise<App> {
     // The signer is wrapped lazily so the cloud storage client can be built before
     // an account is selected. Uploads will throw a clear error if no signer
     // is available at submission time. Reads (fetch / fetchJson) don't need
-    // a signer and work regardless.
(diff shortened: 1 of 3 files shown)


TarikGul (approved):
🚀

PR #201, opened 2026-06-11
fix(host): make @novasamatech/* runtime deps

Imod7:
### Summary

`@parity/product-sdk-host` declared both `@novasamatech/host-api` and `@novasamatech/host-api-wrapper` as **optional `peerDependencies`**. This PR moves them to regular **`dependencies`** (via the existing `catalog:` range) and removes the `peerDependencies` / `peerDependenciesMeta` blocks and the redundant `devDependencies` entries. Also adds `tsup` to `devDependencies` (`catalog:`) so the package's own `build` script resolves it without relying on root hoisting — matching the other tsup-built packages.

### Why

- **`@novasamatech/host-api` was always required, not optional.** The shipped `dist/index.js` statically imports `enumValue` from it (line 2; also in `container.ts` and `truapi.ts`). Marking it an *optional* peer is incorrect - a consumer that omitted it would crash on `import '@parity/product-sdk-host'`.
- **`@novasamatech/host-api-wrapper`** is the host bridge, loaded lazily via `await import(...)`. As a regular dep it's still code-split (no main-bundle cost), and consumers no longer need to declare it.
- **Net effect:** host-embedded apps reach the host APIs purely through `@parity/product-sdk-host` with **no direct `@novasamatech/*` dependency** of their own.

### Singleton safety

The host-bridge transport must be a single instance. All product-sdk packages range `@novasamatech/* >=0.8.0` (catalog `^0.8.7-2`), so npm/pnpm dedupes to one hoisted copy - same single instance as the previous peer-hoisted layout.

### Heads-up for reviewers (version-pin tradeoff)

**Before:** `product-sdk-host` said *"I might need novasama, but I won't install it - the
app should bring its own version."* Each app picked its own.

**After:** `product-sdk-host` says *"I always need novasama, and I install this exact
version (`^0.8.7-2`) myself."* Every app using `product-sdk-host` now gets that version
handed to it automatically.

**The catch:** an app that already asks for a *different* novasama version would now have two requests in play - its own (`0.6.1`) and the one `product-sdk-host` brings (`0.8.7-2`). The package manager can't merge two versions, so it installs **both copies side by side**. That wastes space and can occasionally cause subtle bugs (two versions of the same library at once - e.g. the host "bridge" state split across them).

**Why it's fine:** those apps should be on the standard catalog version anyway; the fix doesn't break them, it just makes a pre-existing version mismatch visible. Please check whether any consumer ends up with a duplicate novasama after this and nudge it onto `^0.8.7-2`. This is a "glance and confirm," not a blocker.

### Testing

- `pnpm --filter @parity/product-sdk-host build`: dist still externalizes novasama.
- Throwaway consumer depending only on `@parity/product-sdk-host` (no novasama declared): novasama resolves transitively; `getTruApi` / `enumValue` work.
- `pnpm -r build` / `pnpm -r test` green.

### DIFF
--- a/product-sdk/packages/host/package.json
+++ b/product-sdk/packages/host/package.json
@@ -20,24 +20,13 @@
         "clean": "rm -rf dist"
     },
     "dependencies": {
+        "@novasamatech/host-api": "catalog:",
+        "@novasamatech/host-api-wrapper": "catalog:",
         "@parity/product-sdk-logger": "workspace:*",
         "polkadot-api": "catalog:"
     },
-    "peerDependencies": {
-        "@novasamatech/host-api-wrapper": ">=0.8.0",
-        "@novasamatech/host-api": ">=0.8.0"
-    },
-    "peerDependenciesMeta": {
-        "@novasamatech/host-api-wrapper": {
-            "optional": true
-        },
-        "@novasamatech/host-api": {
-            "optional": true
-        }
-    },
     "devDependencies": {
-        "@novasamatech/host-api-wrapper": "catalog:",
-        "@novasamatech/host-api": "catalog:",
+        "tsup": "catalog:",
         "typescript": "^5.7.0",
         "vitest": "^3.0.0"
     },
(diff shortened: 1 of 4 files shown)


TarikGul (approved):
🚀

PR #212, opened 2026-06-15
feat: sign messages with DotNS identity

pgherveou:
## Summary                                                                                                                                                                                                                                        
                                         
  Adds `wallet.signMessageWithDotNsIdentity` to `@parity/product-sdk` for signing with the account that owns a People / People Lite username. Resolves `Resources.UsernameOwnerOf` on the caller-supplied People / Individuality chain descriptor, matches the resolved owner against connected wallet accounts, and signs via `getSigner().signBytes(...)`. Falls back to the host's primary username when none is supplied.
                                                                                                                                                                                                                                                    
  Also adds:                                                
  - `useWallet().signMessageWithDotNsIdentity` for the React surface.
  - `SignerManager.getUserId()` wrapping the existing `HostProvider.getUserId()` for non-product-account identity fetches.                                                                                                                          
                                                                                                                                                                                                                                                    
  Chain-connection lifecycle is automatic: reuses an already-connected client when `app.chain.connect({ ..., <name>: peopleChain })` was called upfront, otherwise opens a transient one and caches it.                                             
                                                                                                                                                                                                                                                    
  ## Usage                                                                                                                                                                                                                                          
                                                            
  ```ts
  const app = await createApp({ name: "my-app" });
  await app.wallet.connect();                                                                                                                                                                                                                       
  
  const { username, accountId, signature } = await app.wallet.signMessageWithDotNsIdentity({                                                                                                                                                        
      peopleChain: paseo_individuality,                                                                                                                                                                                                             
      message: "verifying ownership",
  });                                                                                                                                                                                                                                               
  ```                                                       

  Optional: pre-warm the connection for predictable first-call latency.                                                                                                                                                                             
  
  ```ts                                                                                                                                                                                                                                             
  await app.chain.connect({ people: paseo_individuality }); 
  // ...signMessageWithDotNsIdentity calls now skip the cold path
  ``` 
  
  Note: Edited by @TarikGul

PR #214, opened 2026-06-15
feat(signer,sdk): sign DotNS identity via explicit legacy signer

ReinhardHatko:
Stacks on #212 (`feat/dotns-identity-signing`) — merge that first; this targets its branch.

## Problem

`signMessageWithDotNsIdentity` resolves the username → owner `AccountId`, then signs by finding that account in the **connected-accounts list**. On Proof-of-Personhood / product-account hosts (e.g. Polkadot Desktop) the host exposes only a per-dapp **product account** and never *enumerates* the user's identity account — so the owner is never found and signing fails with:

> DotNS username "…" resolves to …, but no connected wallet account matches that AccountId.

That makes DotNS-identity signing unusable on exactly the hosts the product targets.

## Fix

Probing Polkadot Desktop confirmed these hosts will still sign with an account when it is **named explicitly** via `host_sign_raw_with_legacy_account`, even though they withhold it from `getLegacyAccounts()` enumeration. So:

- **`HostProvider.signRawWithLegacyAccount(publicKey, data)`** — signs with an account named by public key, via a shared `legacySignerFor(publicKey)` helper now also used by `mapAccounts` (dedupes the legacy-account construction).
- **`SignerManager.signRawWithLegacyAccount(publicKey, data)`** — thin delegation, mirrors `getUserId` / `createRingVRFProof`.
- **`signMessageWithDotNsIdentity`** now prefers an already-connected account (no extra prompt) and otherwise signs via the explicit legacy signer, instead of throwing when the owner isn't enumerable.

## Validation

- Probed inside Polkadot Desktop: the identity account that owns the People username is withheld from `getLegacyAccounts()` enumeration but signs when named explicitly — the case this PR unblocks.
- `tsc --noEmit` clean for `signer` + `sdk`; existing signer suite passes (102/102).
- Dedicated unit tests for the new path are a follow-up.

### DIFF
--- a/product-sdk/packages/signer/src/providers/host.ts
+++ b/product-sdk/packages/signer/src/providers/host.ts
@@ -498,6 +498,43 @@ export class HostProvider implements SignerProvider {
         }
     }
 
+    /**
+     * Sign a raw message with a specific account identified by its public key,
+     * via the host's legacy-account signing path.
+     *
+     * Unlike `connect()`, this does NOT require the account to be returned by
+     * `getLegacyAccounts()` — it signs through `host_sign_raw_with_legacy_account`
+     * using the SS58 address derived from `publicKey`, which the host resolves
+     * itself. Hosts that withhold the user's identity account from *enumeration*
+     * (e.g. Proof-of-Personhood / product-account hosts that only expose a
+     * per-dapp derived account) may still honour an explicitly-named signer,
+     * typically behind a user-approval prompt.
+     *
+     * This is what makes DotNS-identity signing work on such hosts: the owning
+     * account resolved from the name is never enumerable, but it can still be
+     * named explicitly here. See {@link SignerManager.signRawWithLegacyAccount}.
+     *
+     * Requires a prior successful `connect()` call.
+     */
+    async signRawWithLegacyAccount(
+        publicKey: Uint8Array,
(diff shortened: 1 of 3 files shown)


product-sdk/packages/signer/src/providers/host.ts
```diff
@@ -709,6 +746,25 @@ export class HostProvider implements SignerProvider {
...
+     * legacy accounts (the host-api shape just reuses the product-account
+     * struct). Shared by `mapAccounts` (enumerated accounts) and
+     * `signRawWithLegacyAccount` (an explicitly-named, non-enumerated account).
+     */
+    private legacySignerFor(publicKey: Uint8Array): import("polkadot-api").PolkadotSigner {
+        if (!this.accountsProvider) {
+            throw new Error("Host provider is disconnected");
+        }
+        return this.accountsProvider.getLegacyAccountSigner({
+            dotNsIdentifier: "",
```
mordamax, line 762: should it be empty?

ReinhardHatko: AFAIK yes, to get the legacy account signer you just leave the dotns identifier empty

TarikGul:
Closing since its been completed in: https://github.com/paritytech/product-sdk/pull/212

PR #227, opened 2026-06-26
fix: review comments

decrypto21:
## Summary
Small, contained fixes surfaced by a review pass of the migration (3 files, +37/−7). Stacked on vf/truapi-migration; validated (host + statement-store build, 120 unit tests, biome clean) and adversarially reviewed (verdict: ship).

## Changes
- statement-store: surface host subscription interrupts as `onError`. The host subscription's `onInterrupt` was never registered, so a host-side teardown (transport close / interrupt) never reached the consumer - a silent dead stream.
- host (papi-provider): settle the JSON-RPC request on a synchronous handler throw. A malformed positional param threw out of `send()` before the error boundary, leaving the request unsettled (caller hangs). Now returns a JSON-RPC error + logs.
- host (accounts): reject empty `extrinsic.version` instead of silently signing as Extrinsic V4.
- host (papi-provider): correct a stale code comment - @parity/truapi mints a fresh requestId per subscription and does not dedupe by (method, payload-hash).

## Notes
- These are improvements on the migrated code, not migration regressions (the same behaviors exist in the novasama baseline). Flag for reviewers: the statement-store change is observable behavior product-sdk didn't have before - fine to land here, or split into resilience work if we want #186 kept strictly drop-in.
- Covered by the migration's existing pending changeset (host + statement-store minor); no new changeset.

Refs #186

### DIFF
--- a/product-sdk/packages/host/src/accounts.ts
+++ b/product-sdk/packages/host/src/accounts.ts
@@ -270,6 +270,9 @@ function adaptAccountsProvider(client: TrUApiClient): AccountsProvider {
                     // otherwise the latest supported version. unifyMetadata
                     // normalizes v14/v15 so `.extrinsic.version` is an array.
                     const versions = unifyMetadata(decAnyMetadata(metadata)).extrinsic.version;
+                    if (versions.length === 0) {
+                        throw new Error("No extrinsic version found in metadata");
+                    }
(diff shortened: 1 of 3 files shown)


TarikGul (approved):
🚀

PR #230, opened 2026-06-26
refactor(host): remove PJS bridge from account signer factories

### LINKED ISSUE #228: Remove PJS types/APIs from `@parity/product-sdk-host`'s account signer factories
## Summary

`packages/host/src/accounts.ts` still routes two signing paths through
`polkadot-api/pjs-signer` (`getPolkadotSignerFromPjs` + `SignerPayloadJSON`). truapi
exposes a first-class `createTransactionWithLegacyAccount`, so the legacy-account signer
can be rebuilt as a plain `PolkadotSigner` with no PJS bridge — matching the product-account
path, which is already PJS-free. This removes the SDK's last dependency on
`polkadot-api/pjs-signer`.

Deferred out of the truapi migration PR to keep that diff focused (it changes the public
`AccountsProvider` interface). Originating review comment:
> "We should take this opportunity to completely remove any connection to PJS types - and/or APIs."

## Background

`getPolkadotSignerFromPjs` is an *adapter* that wraps PJS-style
`signPayload(SignerPayloadJSON)` / `signRaw` callbacks into PAPI's minimal `PolkadotSigner`
(`{ publicKey, signTx, signBytes }`). It's used in exactly two places:

1. `productPjsSigner` — the **deprecated `"signPayload"`** mode of `getProductAccountSigner`.
2. `getLegacyAccountSigner` — currently the **only** legacy-account signing path.

The default product path (`signerType: "createTransaction"`) is already PJS-free — a plain
`PolkadotSigner` literal whose `signTx` calls `signing.createTransaction(...)` directly.

## Key enabler

`@parity/truapi`'s `SigningClient` exposes `createTransactionWithLegacyAccount(LegacyAccountTxPayload)`.
`LegacyAccountTxPayload` is identical to `ProductAccountTxPayload` except `signer` is a raw
account id (`AccountId = HexString`) instead of `{ dotNsIdentifier, derivationIndex }`. So the
legacy signer can mirror the product `createTransaction` path with no PJS.

## Decision

**Drop the `"signPayload"` mode entirely** (rather than reimplement it PJS-free). The signer
package already hard-pins `"createTransaction"` (`PRODUCT_SIGNER_TYPE` in
`signer/src/providers/host.ts`), so `"signPayload"` is unused — and `createTransaction` is the
only path that forwards opaque signed extensions (e.g. Paseo Next's `AsPgas`) that PJS strips.

### `packages/host/src/accounts.ts`
- [ ] Remove `import { getPolkadotSignerFromPjs, type SignerPayloadJSON } from "polkadot-api/pjs-signer"`. Keep `import { AccountId, type PolkadotSigner } from "polkadot-api"` (PAPI, not PJS).
- [ ] Delete the dead helpers: `asHex`, `buildSigningPayloadFields`, `productPjsSigner`.
- [ ] `getProductAccountSigner(account)` — drop the `signerType` param and the `"signPayload"` branch; always return the `createTransaction` literal.
- [ ] `getLegacyAccountSigner(account)` — rebuild as a plain `PolkadotSigner` literal:
  - `signTx` → `signing.createTransactionWithLegacyAccount({ signer: toHex(publicKey), genesisHash: toHex(CheckGenesis.additionalSigned), callData: toHex(callData), extensions, txExtVersion })`
  - `signBytes` → `signing.signRawWithLegacyAccount({ signer: AccountId().dec(publicKey), payload: { tag: "Bytes", value: { bytes: toHex(data) } } })`
- [ ] Update the `AccountsProvider` interface signature + JSDoc (drop `signerType`).
- [ ] Drop the now-unused `HexString` / `HostSignPayloadData` type imports.

### Field gotchas
- `AccountId = HexString` in truapi → `createTransactionWithLegacyAccount.signer` takes `toHex(publicKey)` (hex), not raw bytes.
- `signRawWithLegacyAccount.signer` is `string` ("SS58 or hex") → pass the **SS58 address** (`AccountId().dec(publicKey)`); capture it in the closure since `signBytes(data)` has no address arg.
- Keep the `signedExtensions.CheckGenesis` guard and the `txExtVersion` derivation (`unifyMetadata(decAnyMetadata(metadata)).extrinsic.version` → reduce-max → `v4 ? 0 : v`).

### `packages/signer/src/providers/host.ts`
- [ ] Remove `PRODUCT_SIGNER_TYPE` + its doc comment.
- [ ] Drop `signerType?` from the locally re-declared `AccountsProvider` interface.
- [ ] Update the two call sites to stop passing the 2nd arg (`getSigner` closure + `getProductAccountSigner`).
- [ ] Refresh the now-stale "pinned to `createTransaction`" comments.

### Folds in: decrypto21's `assetId` comment
- [ ] The unchecked `as unknown as HexString` on the fee-asset field lives in `buildSigningPayloadFields`, which is **deleted** by this work — so this open review comment resolves here (no separate guard needed; the code is gone).

### Tests
- [ ] `host/src/accounts.ts` in-source vitest: extend the fake `signing` mock with `createTransactionWithLegacyAccount` + `signRawWithLegacyAccount`; add a legacy `signBytes` test (asserts SS58 `signer`) and a legacy `signTx` CheckGenesis-missing throw test.
- [ ] `signer/src/providers/host.ts`: the regression test asserting `getProductAccountSigner` is called with `"createTransaction"` (3 assertions) → reframe to "delegates to the host accounts provider; no signerType arg".
- [ ] Demos `contracts-demo` / `tx-demo` header comments claim the legacy path "always routes through PJS" — update (both paths now go through the host create-transaction path).

TarikGul:
## Summary

Removes the last PolkadotJS (`polkadot-api/pjs-signer`) dependency from `@parity/product-sdk-host`'s account signer factories. truapi exposes `createTransactionWithLegacyAccount`, so the legacy-account signer can build a PAPI `PolkadotSigner` directly — the same way the product-account path already did.

Follow-up to the truapi migration (#186).

## Changes

- **`host/src/accounts.ts`**
  - `getLegacyAccountSigner` is rebuilt over `signing.createTransactionWithLegacyAccount` (`signTx`) and `signRawWithLegacyAccount` (`signBytes`) — no PJS bridge. So opaque signed extensions (e.g. Paseo Next's `AsPgas`) now survive end-to-end for legacy accounts too, not just product accounts.
  - `getProductAccountSigner` drops its `signerType` parameter; the deprecated `"signPayload"` (PJS) mode is gone.
  - Removed the dead helpers (`getPolkadotSignerFromPjs` import, `buildSigningPayloadFields`, `asHex`, `productPjsSigner`).
- **`signer/src/providers/host.ts`** — removed `PRODUCT_SIGNER_TYPE` and the `signerType` argument; the `HostProvider` just calls `getProductAccountSigner(account)`.
- **`tx-demo` / `contracts-demo`** — refreshed comments that claimed the legacy path "routes through PJS".

## Breaking change

`getProductAccountSigner` no longer takes a `signerType` argument. Product-account signing always uses the host's `createTransaction` path (this was already the only path the signer used).

closes: https://github.com/paritytech/product-sdk/issues/228

### DIFF
--- a/product-sdk/packages/host/src/accounts.ts
+++ b/product-sdk/packages/host/src/accounts.ts
@@ -138,6 +138,10 @@ export interface AccountsProvider {
  * Derive the host's extrinsic-extension version from SCALE-encoded metadata:
  * v4 → 0, otherwise the latest supported version. `unifyMetadata` normalizes
  * v14/v15 so `.extrinsic.version` is an array.
+ *
+ * Indirected through {@link deps} so the SCALE decode (which needs a real
+ * metadata blob) can be stubbed in unit tests while the rest of the `signTx`
+ * flow — genesis extraction, extension mapping, the host call — is exercised.
  */
 function deriveTxExtVersion(metadata: Uint8Array): number {
     const versions = unifyMetadata(decAnyMetadata(metadata)).extrinsic.version;
@@ -147,6 +151,9 @@ function deriveTxExtVersion(metadata: Uint8Array): number {
     const latestVersion = versions.reduce((acc, v) => Math.max(acc, v), 0);
     return latestVersion === 4 ? 0 : latestVersion;
 }
+
+/** Internal seam so `import.meta.vitest` can stub the metadata decode. @internal */
+const deps = { deriveTxExtVersion };
 
 /**
  * Map a PAPI `signTx` call's signed extensions onto the host's
@@ -233,7 +240,7 @@ function adaptAccountsProvider(client: TrUApiClient): AccountsProvider {
                             genesisHash: toHex(checkGenesis.additionalSigned),
                             callData: toHex(callData),
                             extensions: toHostExtensions(signedExtensions),
-                            txExtVersion: deriveTxExtVersion(metadata),
+                            txExtVersion: deps.deriveTxExtVersion(metadata),
                         }),
                         "createTransaction failed",
                     );
@@ -272,7 +279,7 @@ function adaptAccountsProvider(client: TrUApiClient): AccountsProvider {
                             genesisHash: toHex(checkGenesis.additionalSigned),
                             callData: toHex(callData),
                             extensions: toHostExtensions(signedExtensions),
-                            txExtVersion: deriveTxExtVersion(metadata),
+                            txExtVersion: deps.deriveTxExtVersion(metadata),
                         }),
                         "createTransactionWithLegacyAccount failed",
                     );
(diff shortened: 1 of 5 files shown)


product-sdk/packages/host/src/accounts.ts
```diff
@@ -458,4 +413,32 @@ if (import.meta.vitest) {
...
+            "signRawWithLegacyAccount",
+            {
+                signer: AccountId().dec(publicKey),
+                payload: { tag: "Bytes", value: { bytes: toHex(new Uint8Array([7, 7])) } },
+            },
+        ]);
+        expect(signature).toEqual(fromHex("0xcafe"));
+    });
+
+    test("the legacy signer's signTx throws without a CheckGenesis extension", async () => {
```
valentinfernandez1, line 442: Missing happy path test for `signTx()`

valentinfernandez1 (commented):
All looks good just have one comment that needs addressing

product-sdk/packages/host/src/accounts.ts
TarikGul: fixed :)

valentinfernandez1 (approved):

PR #231, opened 2026-06-29
feat(host): return Result instead of throwing from public API

### LINKED ISSUE #229: Make `@parity/product-sdk-host`'s public API return `Result` instead of throwing
## Summary

The entire `@parity/product-sdk-host` public API is throw-style: every exported function
returns a `Promise<T>` (or `Promise<T | null>`) and throws on error, with `unwrapHostResult`
as the shared DRY helper that collapses the `.match(ok, err => throw)` dance. Move the
throw→`Result` boundary down into `host` so consumers get typed `Result<T, E>` errors instead
of opaque thrown `Error`s.

Deferred out of the truapi migration PR. Originating review comment (on `unwrapHostResult`):
> "this seems counter productive to unwrap result to throw errors, aren't we better off with
> Result type rather than throwing opaque errors?"

And the decision to defer:
> "Yea I think this along with the PJS signPayload api can be deferred to a follow up PR. I
> agree this is some technical debt brought on from the previous workflow, and is just really
> a DRY helper, but if it were to be correctly fixed in this PR it would touch A LOT of packages."

## Background / context

- The throw-style is **pre-existing**, not introduced by the truapi migration — `requestPermission`,
  `deriveEntropy`, etc. already threw on `main`. `unwrapHostResult` is a migration-era DRY helper
  that *implements* the existing convention; it didn't create it.
- A `Result` boundary **already exists one layer up**: `@parity/product-sdk-signer`
  (`SignerManager` / `HostProvider`) consumes host's throwing functions inside `try/catch` and
  surfaces typed `Result<T, SignerError>`. So this issue is really about **where the boundary
  sits** — push it down into `host` so every consumer benefits, not just the signer.
- `host` errors already have structure to lean on: `formatHostError` + the exported `HostError`
  type (`GenericError | { tag } | { tag, value: { reason } }`), and the original error is
  preserved as `cause` on the thrown `Error`.

## Scope (this is large — measured)

- **~23 throwing public host functions** to convert to `Result` / `ResultAsync`.
- **~154 call sites** across **8 consumer packages** (`chain-client`, `cloud-storage`,
  `local-storage`, `sdk`, `signer`, `statement-store`, + host-internal) **and 3 example demos**
  (`host-demo`, `keys-demo`, `storage-demo`).
- A host error-type design (a `HostError`-class hierarchy, or reuse the existing `HostError`
  union + a thin error wrapper).
- Breaking ripple through the umbrella `@parity/product-sdk`.

This is why it's its own PR — folding it into the migration would have made that diff
un-reviewable and conflated "did the port preserve behavior?" with "did we redesign error
handling correctly?".

## Open design questions (decide before implementing)

- [ ] **Return shape:** neverthrow `ResultAsync<T, HostError>` (host already depends on neverthrow
      transitively via truapi), or a plain tagged `Promise<Result<T, E>>` like the signer's
      `{ ok, value } | { ok, error }`? Prefer matching whatever the signer already exposes so the
      two layers compose cleanly.
- [ ] **Error type:** a typed `HostError` class hierarchy (à la `signer`'s `HostUnavailableError`
      / `HostRejectedError`), or carry truapi's `HostError` union through directly? The signer's
      error classes are a good template.
- [ ] **`null`-returning getters:** functions like `getThemeProvider(): Promise<T | null>` already
      encode "not in a host" as `null`. Decide whether those stay `null`-returning or fold into
      the `Result` (e.g. `err(HostUnavailable)`).
- [ ] **`getChainSpec` etc.** that return `null` on "outside container" but throw on real failure —
      reconcile into one error model.

## Tasks (once shape is decided)

- [ ] Convert the ~23 host public functions to the chosen `Result` shape; `unwrapHostResult`
      becomes an internal `mapHostResult` (or is removed) — it stops being the throw point.
- [ ] Define/extend the host error type(s).
- [ ] Update the ~154 consumer call sites across the 8 packages + 3 demos.
- [ ] Simplify `signer`'s `HostProvider`: it currently wraps host throws in `try/catch → err(...)`;
      with host returning `Result`, that try/catch collapses into a `.map`/`.mapErr`.
- [ ] Update skills / docs that show `await host.foo()` throwing patterns.

valentinfernandez1:
Closes #229.

## Summary

Moves the throw→`Result` boundary down into `@parity/product-sdk-host`. The flat
public host operations now return a tagged `Result<T, HostError>` instead of
throwing opaque `Error`s, so **every** consumer gets typed errors — not just the
signer, which previously wrapped host's throws in its own `try/catch` and
re-exposed a `Result` one layer up.

The underlying `@parity/truapi` client already returns typed `Result`s; host was
deliberately throwing them away via the `unwrapHostResult` helper. This change
stops doing that for the public operations, direct host consumers now get typed
errors on a `Result` channel instead of catching opaque `Error`s. (Previously only
the signer recovered any structure, by wrapping host's throws in `try/catch` and
re-exposing a `Result` one layer up.)

## What changed

**New exports (`@parity/product-sdk-host`)**

- `Result<T, E>` (`{ ok: true; value } | { ok: false; error }`) + `ok()` / `err()`
  constructors. The shape is intentionally identical to `@parity/product-sdk-signer`'s
  `Result`, so the two layers compose with no adapter.
- A `HostError` class hierarchy mirroring the signer's error classes:
  - `HostError` — base, extends `Error`
  - `HostUnavailableError` — running outside a host container
  - `HostCallFailedError` — a call reached the host but failed; carries the
    structured truapi error as `.payload` (and as `cause`)
  - `isHostError(e)` type guard

**Breaking**

- These now return `Promise<Result<T, HostError>>` instead of throwing:
  `requestPermission`, `requestDevicePermission`, `requestResourceAllocation`,
  `createProofAuthorized`, `deriveEntropy`, `navigateTo`, `broadcastTransaction`,
  `stopTransaction`, `featureSupported`, `isChainSupported`.
- `getChainSpec` now returns `Promise<Result<ChainSpec | null, HostError>>`.
  `ok(null)` still means "outside a host container" (an expected state, not a
  failure); real call failures move to the `err` channel.
- The exported error-payload type **`HostError` is renamed to `HostErrorPayload`**
  (the structured truapi `Err`-channel shape), freeing the name `HostError` for the
  new base class. The payload now rides inside `HostCallFailedError.payload`.
  
  ## Migration

```ts
// Before: threw on failure, required try catch logic.
const granted = await requestPermission({ tag: "ChainSubmit", value: undefined });

// After
const r = await requestPermission({ tag: "ChainSubmit", value: undefined });
if (!r.ok) handle(r.error);      // r.error is a typed HostError
else use(r.value);
```
Consumers that imported the HostError type should rename it to HostErrorPayload


product-sdk/packages/host/src/truapi.ts
```diff
@@ -23,15 +25,17 @@ import type {
 } from "@parity/truapi";
 import { createLogger } from "@parity/product-sdk-logger";
 
+import { type HostError, HostCallFailedError, HostUnavailableError } from "./errors.js";
```
TarikGul: Just a nit, but we have circular imports here between errors.ts and truapi.ts, harmless now but could cause issues later.

product-sdk/packages/signer/src/providers/host.ts
```diff
@@ -180,6 +181,18 @@ async function defaultLoadAccountsProvider(): Promise<AccountsProvider | null> {
     return (await getAccountsProvider()) as unknown as AccountsProvider | null;
 }
 
+/**
+ * Default `requestChainSubmitPermissionFn`: bridge host's `Result`-returning
+ * {@link requestPermission} back to the option's `Promise<boolean>` contract by
+ * throwing the typed {@link HostError} on the `err` channel. The throw is caught
```
TarikGul: (claude suggestion) - but saying this link for HostError could be broken.

TarikGul (approved):
Overall, looks great!

PR #238, opened 2026-07-08
feat!: SDK-wide Result error model

TarikGul:
## SDK-wide Result error model

Implements the [Result error model spec](https://github.com/paritytech/sdk-team/pull/40): fallible operations
across `@parity/product-sdk-*` now return a typed `Result<T, E>` instead of
throwing. Consumers branch on `r.ok` and read `r.value` / `r.error`.

### What's here

- **New package `@parity/result`** — a zero-dependency leaf with
  `Result<T, E>`, `ok()` / `err()`, and a cross-package `SdkError` marker +
  `isSdkError(e)`. Re-exported from `@parity/product-sdk`.
- **`SdkError` marker** — each package keeps its own error hierarchy; every base
  error implements the marker (with a `source` string) so `isSdkError(e)`
  recognizes any SDK-origin error without importing per-package classes.
- **Converted packages** — `tx`, `contracts`, `cloud-storage`, `statement-store`
  return `Result` from their fallible operations. `host` and `signer` (already
  Result-returning) migrate onto the shared package.
- **Umbrella** — `createApp().cloudStorage.upload` / `fetch` return `Result`.
- **Migration guide** — `guides/migrating-to-result`.

### Deliberate carve-outs

- `contract.<method>.query` keeps returning `QueryResult<T>` — a dry-run revert
  is an expected outcome, not an error.
- Methods implementing an external contract stay unchanged — the PAPI
  `PolkadotSigner` interface, and `CloudStorageClient` methods forwarding to the
  upstream client.
- Pure/sync helpers, factories, codegen, and lifecycle/subscription methods keep
  throwing.

### Breaking changes

The converted operations return `Result` instead of throwing (or, for
`statement-store`'s `publish` / `ChannelStore.write`, instead of returning
`boolean`). See the changeset and migration guide for the full per-package list.

⚠️ Boolean → Result is **not** caught by the compiler — a bare `if (result)` is
always truthy for a `Result` object. Audit `statement-store` call sites for `.ok`.


valentinfernandez1:
Some Claude skills like [`product-sdk-transactions/SKILL.md`](https://github.com/paritytech/product-sdk/blob/main/product-sdk/skills/product-sdk-transactions/SKILL.md#quick-start-submit-a-transaction-in-10-lines) have to be updated with this new Result model.

docs/content/guides/_meta.ts
```diff
@@ -2,4 +2,5 @@
 // SPDX-License-Identifier: Apache-2.0
 export default {
```
valentinfernandez1, line 3: I didn't though about this back on the truapi migration but maybe this guides related to migrations should be under their own "Migration" section on the docs and sub sectioned by release version, so that users looking at the docs knows whether a migration applies to them.

Will not block the PR from merging because of this though we can refactor that on a different PR

product-sdk/packages/result/package.json
```diff
@@ -0,0 +1,27 @@
+{
+    "name": "@parity/result",
```
valentinfernandez1, line 2: This package name is too generic, if published on npm it would be really hard for someone else to know that this is for product-sdk specifically.

We should probably keep the naming convention we already have, so this package would be "@parity/product-sdk-result".

product-sdk/packages/contracts/src/manager.ts
```diff
@@ -141,45 +147,56 @@ async function queryLiveAddress(
...
-        patchContractAddress(resolved, library, address);
+        const resolved = cloneCdmJson(cdmJson);
+        for (const [library, address] of liveAddresses) {
+            patchContractAddress(resolved, library, address);
+        }
+        return ok(resolved);
+    } catch (cause) {
+        // Internal helpers throw ContractError subclasses; surface them on the err channel.
+        if (cause instanceof ContractError) return err(cause);
+        throw cause;
```
valentinfernandez1: This can still throw, if the registry .query() fails with a raw RPC/network error it's not a `ContractError`, so it throw even though the signature and doc above promise err on a failed lookup.

product-sdk/packages/cloud-storage/src/query.ts
```diff
@@ -66,44 +91,55 @@ export async function executeQuery(
...
     }
-    return out;
 }
 
 if (import.meta.vitest) {
     const { beforeAll, describe, test, expect, vi } = import.meta.vitest;
     const { calculateCid } = await import("@parity/bulletin-sdk");
 
+    /** Assert a Result is `ok` and return its value (test-only). */
+    function unwrap<T>(r: Result<T, unknown>): T {
```
valentinfernandez1: I have seen this function in tests re implemented several times across tests. Should be exported from the @parity/result package to avoid duplication

product-sdk/packages/cloud-storage/src/verify.ts
```diff
@@ -215,6 +225,13 @@ function matchesEntry(entry: ChainEntry, target: ParsedCid): boolean {
 if (import.meta.vitest) {
     const { describe, test, expect, vi } = import.meta.vitest;
 
+    /** Assert a Result is `ok` and return its value (test-only). */
+    function unwrap<T>(r: Result<T, unknown>): T {
```
valentinfernandez1: duplicated

product-sdk/packages/cloud-storage/src/authorization.ts
```diff
@@ -225,18 +231,26 @@ export async function authorizeAccount(
...
 
-    return { blockHash: result.block.hash };
+    return ok({ blockHash: result.value.block.hash });
 }
 
 if (import.meta.vitest) {
     const { describe, test, expect, vi } = import.meta.vitest;
 
+    /** Assert a Result is `ok` and return its value (test-only). */
+    function unwrap<T>(r: Result<T, unknown>): T {
```
valentinfernandez1: duplicated

product-sdk/packages/statement-store/src/client.ts
```diff
@@ -154,16 +159,27 @@ export class StatementStoreClient {
...
-            throw new StatementConnectionError("Not connected. Call connect() first.");
+            return err(new StatementConnectionError("Not connected. Call connect() first."));
+        }
+
+        let dataBytes: Uint8Array;
+        try {
+            // encodeData throws StatementDataTooLargeError / StatementEncodingError.
+            dataBytes = encodeData(data);
+        } catch (error) {
+            return err(error as StatementStoreError);
```
valentinfernandez1: There are methods like `toTxError()` and `toCloudStorageError()` but not for `toStatementStoreError()`.

I see 2 things here, one we could implement the function instead of just casting, but also this surfaces a small issue and it's that there is no common or defined way to cast errors across the code. Maybe better to define that strategy and stick to it on every instance

product-sdk/packages/result/package.json
TarikGul: So this actually comes from: https://github.com/paritytech/sdk-team/pull/40#discussion_r3545610371

The idea being if we use this in other packages then it would need to be generic.

valentinfernandez1: Thanks, will continue the discussion there https://github.com/paritytech/sdk-team/pull/40#discussion_r3559741777

product-sdk/packages/contracts/src/manager.ts
TarikGul: fixed

product-sdk/packages/statement-store/src/client.ts
TarikGul: Fixed.

valentinfernandez1: I don't see the fix, have you pushed the commits?

TarikGul: Just needed to push it up

valentinfernandez1 (approved):

filvecchiato (approved):

PR #241, opened 2026-07-09
feat: ship test fakes under /testing across SDK packages

valentinfernandez1:
## What

Adds a dev-only `/testing` subpath to each package, exporting a working in-memory
**fake** of that package's interface so consumers can unit-test SDK-dependent code
with no host container, chain, or wallet:

| Package | Fake |
|---|---|
| `local-storage` | `createFakeHostLocalStorage` |
| `signer` | `createFakeSignerProvider`, `fakeSignerAccount` |
| `statement-store` | `createFakeStatementTransport` |
| `contracts` | `createFakeContractRuntime`, `fakeDryRunResult` |
| `host` | `createFakeTruApiClient`, `createFakeHost` |
| `@parity/product-sdk` (umbrella) | `createFakeApp` + re-exports of all of the above |

## How it's packaged

Each fake is its own tsup entry with a matching `exports` map key, is
framework-agnostic (no vitest/jest in shipped code), and is **absent from every
`dist/index.js`** — so it never lands in a consumer's production bundle. Each
fake is driven through the *real* consumer in its tests (round-trip guards), so a
fake that drifts from the interface it stands in for fails CI.

## Docs
It also includes a testing guide inside the product-sdk docs (will be available at https://paritytech.github.io/product-sdk/guides/testing) that details how to use the fakes in unit tests.


TarikGul:
Probably needs a rebase with the new Result types

product-sdk/packages/sdk/src/testing.ts
```diff
@@ -0,0 +1,282 @@
...
+    WalletApi,
+} from "./core/types.js";
+
+const textEncoder = new TextEncoder();
+
+// Re-export the per-package fakes so `@parity/product-sdk/testing` is a one-stop import.
+export * from "@parity/product-sdk-local-storage/testing";
+export * from "@parity/product-sdk-signer/testing";
+export * from "@parity/product-sdk-contracts/testing";
+export * from "@parity/product-sdk-host/testing";
```
TarikGul, line 36: Are we missing the statement-store here?

product-sdk/packages/host/src/testing.ts
```diff
@@ -0,0 +1,313 @@
...
+                preimages.set(key, value);
+                return okAsync(key);
+            },
+        },
+        // chain.* is read lazily by the PAPI provider; the full JSON-RPC surface
+        // is out of scope (see module header), so this stays a stub.
+        chain: {},
+    };
+
+    return client as unknown as TrUApiClient;
```
TarikGul: Is it possible for us to make this typing stronger - so we dont have to do the unknown?

TarikGul:
Really awesome - Just a few nits!

product-sdk/packages/sdk/src/testing.ts
valentinfernandez1: It's not there on purpose, the umbrella package doesn't wrap the statement-store right now so to avoid adding an additional dependency I decided to also skip it here.

product-sdk/packages/host/src/testing.ts
valentinfernandez1: Fixed

product-sdk/packages/sdk/src/testing.ts
TarikGul: Ahh makes sense thanks for clarifying

TarikGul (approved):
🚀

PR #249, opened 2026-07-16
feat(chains): drop the Summit Network environment

valentinfernandez1:
## What

The Web3 Summit is over and its chains have been decommissioned, so the `summit` environment is no longer needed. `paseo` (Paseo Next v2) and `devnet` (public Paseo testnet) remain the two live presets.


TarikGul (approved):
🚀

PR #259, opened 2026-07-20
fix(signer): degrade gracefully when resolving a product account signed out

### LINKED ISSUE #253: product-sdk-signer: HostProvider error-logs an opaque, retried failure when resolving a productAccount while signed out
# product-sdk-signer: `HostProvider` error-logs an opaque, retried failure when resolving a `productAccount` while signed out

## Summary

When a `HostProvider` configured with `productAccount` (or `dappName`) runs `connect()` while the host has **no signed-in session**, resolving the product account fails with `NotConnected` and the signer emits a top-level `console.error`:

```
[signer:host] failed to get product account { cause: <Error> }
```

This is the **expected signed-out state** — the app should simply fall back to read-only — but the SDK surfaces it as a hard error, opaquely, and repeated several times. Three separate DX problems, none of which are the app's to fix.

## Environment

- `@parity/product-sdk-signer` **0.9.0**
- `@parity/product-sdk-logger` **0.1.1**
- `@parity/product-sdk-host` **0.12.0**
- Host: Polkadot Desktop; app is a static export loaded in the host container.

## Repro

1. Configure a `HostProvider` with a `productAccount` (e.g. `new HostProvider({ productAccount: { dotNsIdentifier: 'my-app.dot' } })`) and drive it via `SignerManager`.
2. Open the app inside a Polkadot host while **signed out** (or sign out, then reload).
3. `connect()` runs on mount → watch the console.

## Actual

Full error (after unwrapping the `Error` — see problem #2):

```
[signer:host] failed to get product account
  cause: Error: Host rejected product account request: NotConnected
    at HostProvider.getProductAccount (product-sdk-signer/dist/index.js:295)
    at HostProvider.fetchProductSignerAccount (…/index.js:536)
    at HostProvider.tryConnect (…/index.js:462)
    at withRetry (…/index.js:193)
    at SignerManager.connectToProvider (…/index.js:867)
```

### 1. `error` level for an expected state
`NotConnected` just means "no signed-in user". It is not a fault — the consumer falls back to read-only. Logging it at `error` makes every signed-out session look broken.

Note the asymmetry already in the codebase: the **`dappName`** branch of `connect()` degrades gracefully on the same failure (`log.warn(...)` + resolve with empty accounts), while the **`productAccount`** branch hard-returns the `err`. The `productAccount` path should degrade the same way.

### 2. Opaque payload — the reason is unreachable
The log call passes the raw `Error` object:

```js
// dist/index.js ~296
log2.error("failed to get product account", { cause });
```

Any structured sink (`configure({ handler })`) or `JSON.stringify` renders this as `{"cause":{}}` / `[object Object]`, because an `Error`'s `message`/`stack` are non-enumerable. Consumers literally cannot see *why* it failed without monkey-patching. Please log `cause.message` (or `formatError(cause)`), not the raw object — the logger already has `formatError`.

### 3. Pointless retries
`connect()` runs the resolution under `withRetry` (default `maxRetries` 3). `NotConnected` is not transient — the user is signed out — so it retries and re-logs the same error up to 3× per load. Non-transient rejections (`NotConnected`) should short-circuit the retry loop.

## Expected

Signed-out `connect()` with a configured `productAccount` should:
- resolve with empty accounts (read-only), matching the `dappName` branch, and log at `debug`/`warn`, not `error`;
- if it does log, include the human-readable reason, not an opaque object;
- not retry a non-transient `NotConnected`.

## Consumer workaround (not a fix)

We temporarily installed a `configure({ handler })` bridge that `Error`-unwraps log data so the message is legible, then removed it in favor of fixing this upstream.

## Suggested fix

In `HostProvider` (`dist/index.js` ~437–465): make the `productAccount` branch mirror the `dappName` branch — on `NotConnected` (or any resolution failure), `log.debug/warn` with the message and resolve with empty accounts instead of erroring; and either exclude `NotConnected` from `withRetry` or don't retry account resolution during connect. Independently, change the `catch` at ~296 to log `formatError(cause)` rather than `{ cause }`.

TarikGul:
## Fix opaque, retried error when resolving a product account while signed out

Closes #253

### Problem
When `HostProvider` is configured with `productAccount` and `connect()` runs without an active user session, the SDK treated the signed-out state as a fault. Three symptoms:

1. **Error-level logging for a normal state** — `NotConnected` (signed out) logged at `error`, even though it's expected. The `dappName` branch already degrades gracefully; the `productAccount` branch didn't.
2. **Unreadable messages** — `log.error("failed to get product account", { cause })` serialized to `{}` because `Error` properties are non-enumerable.
3. **Wasteful retries** — `connect()` retried the non-transient `NotConnected` failure 3×, repeating the same message per load.

### Fix
- **Classify the failure.** A new `isNonTransientHostError` helper walks the tagged-enum / versioned-envelope chain (the same structure `formatError` handles) looking for `NotConnected`. Signed-out failures log at `debug`; the `productAccount` branch soft-degrades to `warn` + empty accounts (read-only), matching `dappName`. Genuine faults still log at `error`.
- **Readable messages.** Logs now carry the `formatError`-derived string instead of a raw `Error`. The original tagged error is preserved via `Error.cause` so classification still works at the catch site.
- **No wasted retries.** Non-transient failures resolve to `ok([])` inside `tryConnect`, so they never reach the retry loop as an error — no `withRetry` change needed. Transient failures still error and retry.

### DIFF
--- a/product-sdk/packages/signer/src/providers/host.ts
+++ b/product-sdk/packages/signer/src/providers/host.ts
@@ -346,7 +346,9 @@ export class HostProvider implements SignerProvider {
             const raw = cause instanceof Error ? cause.cause : cause;
             const nonTransient = isNonTransientHostError(raw);
             if (nonTransient) {
-                log.debug("product account unavailable (signed out)", { error: message });
+                log.debug("product account unavailable (expected, non-transient)", {
+                    error: message,
+                });
             } else {
                 log.error("failed to get product account", { error: message });
             }
@@ -563,7 +565,7 @@ export class HostProvider implements SignerProvider {
                 const error = accountResult.error;
                 if (error instanceof HostRejectedError && error.nonTransient) {
                     log.warn(
-                        "product account unavailable (signed out); resolving with empty accounts",
+                        "product account unavailable (signed out or unregistered); resolving with empty accounts",
                         { dotNsIdentifier: this.productAccount.dotNsIdentifier },
                     );
                     signerAccounts = [];
@@ -703,7 +705,14 @@ export class HostProvider implements SignerProvider {
  * degrades to read-only (empty accounts) instead of erroring and retrying —
  * it's the user not being signed in, which no amount of retrying will fix.
  */
-const NON_TRANSIENT_HOST_TAGS: ReadonlySet<string> = new Set(["NotConnected"]);
+const NON_TRANSIENT_HOST_TAGS: ReadonlySet<string> = new Set([
+    // User is signed out.
+    "NotConnected",
+    // The product's dotNS identifier isn't registered/valid for this user —
+    // the same condition the `dappName` branch already soft-degrades for.
+    // Retrying can't fix either; both resolve to read-only.
+    "DomainNotValid",
+]);
 
 /**
  * Does a host error represent a non-transient, expected condition (e.g. the
@@ -1450,6 +1459,16 @@ if (import.meta.vitest) {
             expect(isNonTransientHostError({ tag: "NotConnected" })).toBe(true);
         });
 
+        test("matches DomainNotValid (unregistered dotNS identifier)", () => {
+            expect(isNonTransientHostError({ tag: "DomainNotValid" })).toBe(true);
+            expect(
+                isNonTransientHostError({
+                    tag: "Domain",
+                    value: { tag: "V1", value: { tag: "DomainNotValid" } },
+                }),
+            ).toBe(true);
+        });
+
         test("matches NotConnected nested inside a versioned envelope", () => {
             expect(isNonTransientHostError({ tag: "v1", value: { tag: "NotConnected" } })).toBe(
                 true,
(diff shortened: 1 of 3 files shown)


product-sdk/packages/signer/src/providers/host.ts
Imod7: **[nice to have]** Here it still uses `log.error("…", { cause })` - the same opaque-`{}` serialization (via a structured `configure({ handler })` sink) that #253 is about. Suggest logging `{ error: cause instanceof Error ? cause.message : String(cause) }` here too so the readable-log fix is consistent across all host RPC wrappers.

Same in `:441` (`getUserId`) and `:482` (`createRingVRFProof`).

Imod7 (commented):
LGTM! Just one nice-to-have and consider adding a changeset in `pending-changesets`.

product-sdk/packages/signer/src/providers/host.ts
TarikGul: added :)

Imod7 (approved):

PR #260, opened 2026-07-20
feat(terminal): add deriveEntropy(session, productId, key) (RFC-0007)

### LINKED ISSUE #254: product-sdk-terminal: expose entropy / encryption-secret derivation (RFC-0007) from a UserSession
## Summary

`@parity/product-sdk-host` exposes a first-class `deriveEntropy(context)` helper (RFC-0007), so in-container (web) apps can derive deterministic, wallet-bound entropy — and from it an encryption identity — without touching `truApi` directly.

`@parity/product-sdk-terminal` has **no equivalent**, even though the QR/SSO `UserSession` already carries the needed material and `@novasamatech/host-papp` already implements the crypto. Terminal (CLI/headless) consumers therefore have to reach *around* the SDK into the low-level transport package and re-derive RFC-0007 themselves.

Please surface a supported entropy / encryption-secret derivation API on `product-sdk-terminal`, at parity with `product-sdk-host.deriveEntropy`.

## Environment

- `@parity/product-sdk-terminal` **0.5.4**
- `@parity/product-sdk-host` **0.12.0** (has `deriveEntropy`)
- `@novasamatech/host-papp` **0.8.10** (has the crypto + the session field)

## What already exists (so this is a re-export/wrapper, not new crypto)

**Web side — clean wrapper (the parity target):**
```ts
// @parity/product-sdk-host
declare function deriveEntropy(key: Uint8Array): Promise<Result<Uint8Array, HostError>>;
// "Derive deterministic entropy from a context key (RFC-0007). The host derives
//  entropy from the user's wallet + the provided context key. Same key + same
//  wallet → same bytes; different keys/wallets → uncorrelated entropy."
```

**Terminal side — material is present but unexposed:**
- The `UserSession` (re-exported by `product-sdk-terminal`) already carries `rootEntropySource: Uint8Array` — per host-papp's `handshakeV2`, `blake2b256_keyed(rootAccountSecret, "product-entropy-derivation")`, i.e. RFC-0007 layer-1.
- `@novasamatech/host-papp`'s `crypto` module already implements the encryption primitives:
  ```ts
  createEncrSecret(entropy: Uint8Array): EncrSecret
  getEncrPub(secret: EncrSecret): EncrPublicKey
  createSharedSecret(secret: EncrSecret, publicKey: Uint8Array): SharedSecret   // ECDH
  ```

So everything needed is present — it's just not reachable through the terminal package's public surface (`.`, `./host`, `./testing` expose adapter/allowance/signing only, no entropy).

## The problem

To derive an app-scoped encryption identity from a terminal session today, a consumer must:
1. import `@novasamatech/host-papp` directly (an internal/transport dep, not the public SDK), and
2. read `session.rootEntropySource` and re-implement the RFC-0007 context derivation (`host_derive_entropy`) that `product-sdk-host` performs for web,
3. then call `createEncrSecret` / `getEncrPub` / `createSharedSecret`.

That couples consumers to internal package layout and to a security-sensitive derivation detail that should live in one place.

## Proposed API

Mirror `product-sdk-host.deriveEntropy`, taking the session explicitly (terminal is out-of-container, so there's no ambient host):

```ts
// @parity/product-sdk-terminal (or a ./crypto subpath)
export function deriveEntropy(session: UserSession, context: Uint8Array): Result<Uint8Array, TerminalError>;

// and/or a higher-level encryption-identity helper:
export function getEncrSecret(session: UserSession, context: Uint8Array): EncrSecret;
export function getEncrPublicKey(session: UserSession, context: Uint8Array): EncrPublicKey;
export function createSharedSecret(session: UserSession, context: Uint8Array, peerPublicKey: Uint8Array): SharedSecret;
```

Same determinism guarantee as the host version (same wallet + same context ⇒ same key), so keys survive device loss as long as the wallet is recoverable.

## Why

- **Feature parity** between in-container (host) and QR/SSO (terminal) apps for entropy-derived encryption.
- **E2E encryption under host-only auth.** Our use case (d3pot — decentralized git host on Polkadot) does per-user, wallet-recoverable repo encryption: derive an encryption identity from RFC-0007 entropy, register its public key on-chain, wrap a per-repo symmetric key to collaborators via ECDH. This works on web via `deriveEntropy` today; the CLI is blocked purely on the missing terminal surface. (We previously derived x25519 from a local keystore seed; migrating to host/QR auth removed seed access, which is what this would restore — deterministically and recoverably.)
- **Avoids consumers depending on `@novasamatech/host-papp` internals** or re-implementing a security-sensitive derivation.

## Notes / open questions

- Should the derivation be done client-side from `rootEntropySource` (deterministic, no round-trip), or routed through the host like the web `truApi.entropy.derive`? Client-side matches the fact that `rootEntropySource` is already local to the session.
- A `./crypto` subpath re-exporting host-papp's branded `EncrSecret`/`EncrPublicKey`/`SharedSecret` types + `createSharedSecret` would let consumers do per-user key wrapping without importing the internal package.

ReinhardHatko:
Refs #254.

## What

Adds `deriveEntropy(session, productId, key): Uint8Array` to `@parity/product-sdk-terminal` — client-side RFC-0007 product-entropy derivation for QR/SSO (out-of-container) sessions, at parity with `@parity/product-sdk-host`'s `deriveEntropy`.

Scoped to the entropy half of #254 deliberately — the encryption helpers it also asks for (`getEncrSecret` / `getEncrPublicKey` / `createSharedSecret`) are **not** in this PR, so the issue stays open. Entropy is the drift-prone part and is worth landing on its own; `@parity/product-sdk-crypto` already covers the rest with `boxEncrypt` / `sealedBoxEncrypt`.

## Why

The host package derives entropy in-container via `truApi.entropy.derive`. Terminal apps run out-of-container and had no equivalent, even though the paired `UserSession` already carries `rootEntropySource` (RFC-0007 layer 1). Consumers were forced to import `@novasamatech/host-papp` internals and re-implement the derivation.

## How

The scheme is specified normatively by [host-spec §C.8](https://github.com/paritytech/host-spec/blob/adb3989208ae1c2107dbf0159611353e6989422c/spec/C-account-derivation.md?plain=1#L129-L147):

```
layer1 = blake2b256_keyed(key = utf8("product-entropy-derivation"), msg = rootAccountSecret)
layer2 = blake2b256_keyed(key = blake2b256(utf8(productId)),        msg = layer1)
result = blake2b256_keyed(key = callerKey,                          msg = layer2)
```

The session already carries layer 1 as `rootEntropySource`, so layers 2 and 3 are computed locally with no host round-trip. Output matches what an in-container app gets from `@parity/product-sdk-host`'s `deriveEntropy` for the same wallet + product + key, so entropy-derived keys interoperate across web (in-container) and terminal (QR/SSO) clients. Deterministic from the wallet (not the device), so derived keys survive device loss as long as the wallet is recoverable.

- `productId` is the calling product's identifier — `<label>.dot` for a dotNS product, `localhost:<port>` for local dev. Used verbatim and case-sensitively; the host scopes it **per deployment**, so production, PR previews and local dev all derive different entropy. Required and positional by design: a wrong default would silently yield undecryptable data rather than a loud error.
- `key` is the caller key, 1..32 bytes (the layer-3 BLAKE2b key). Both bounds are normative.

## Notes

- `rootEntropySource` is read as the real typed `UserSession` field (an upstream rename is a compile error here), with a runtime guard for older V1 persisted sessions that predate RFC-0007.
- In-source tests (`import.meta.vitest`) pin the implementation against the **canonical cross-language conformance suite** ([`useragent-kit@92641ed conformance/rfc0007-product-entropy-v1.json`](https://github.com/paritytech/useragent-kit/blob/92641ed2cba46e39e9682230b91252cd21dfa396/conformance/rfc0007-product-entropy-v1.json)) — all 8 vectors, shared byte-for-byte with the Rust (canonical), Kotlin and Swift implementations. Plus determinism, product/key scoping, verbatim-`productId`, both key-length bounds accepted and both rejected, and the missing/short `rootEntropySource` guards.
- `treeshake: true` added to the package's tsup config: without it the in-source test blocks (including the vector hexes) were being shipped in `dist/index.js`. `terminal` was the only package in `packages/*` missing the flag.
- Documented that the derivation stays here rather than moving to a shared package: the scheme has five implementations across three languages, so the shared artifact is the spec, not a TS package. A follow-up lifts `blake2b256Keyed` into `@parity/product-sdk-crypto` and the pure derivation into `@parity/product-sdk-keys`, with no behaviour change.
- Motivating consumer: d3pot (decentralized git host) needs per-repo, wallet-recoverable encryption identities under host-only auth.

### DIFF
--- a/product-sdk/packages/terminal/src/entropy.ts
+++ b/product-sdk/packages/terminal/src/entropy.ts
@@ -1,0 +1,123 @@
+// Copyright 2026 Parity Technologies (UK) Ltd.
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Client-side RFC-0007 product-entropy derivation for terminal (QR/SSO) sessions.
+ *
+ * Byte-for-byte identical to the host's `host_derive_entropy` handler — i.e.
+ * `@novasamatech/host-container`'s `deriveProductEntropyFromSource` — so entropy
+ * derived here matches what an in-container app gets from
+ * `@parity/product-sdk-host`'s `deriveEntropy` for the same wallet + product +
+ * key. Derived keys therefore interoperate across web (in-container) and
+ * terminal (QR/SSO) clients.
+ *
+ * The paired {@link UserSession} carries `rootEntropySource` (RFC-0007 layer 1,
+ * `blake2b256_keyed(rootAccountSecret, "product-entropy-derivation")`), so
+ * layers 2 and 3 are computed locally with no host round-trip:
+ *
+ *   perProduct = blake2b256(rootEntropySource, key = blake2b256(utf8(productId)))
+ *   entropy    = blake2b256(perProduct,        key = key)
+ *
+ * NOTE: this re-derives the RFC-0007 scheme locally and MUST stay byte-identical
+ * to `host-container`'s `deriveProductEntropyFromSource`. The golden-vector test
+ * below guards against drift, but the derivation ideally belongs in a shared
+ * crypto package that both host-container and terminal import (see PR #260).
+ *
+ * @module
+ */
+
+import { blake2b } from "@noble/hashes/blake2.js";
+import type { UserSession } from "@novasamatech/host-papp";
+
+const ROOT_ENTROPY_LEN = 32;
+const textEncoder = new TextEncoder();
+
+/** BLAKE2b-256, optionally keyed — the RFC-0007 primitive. */
+const b2 = (message: Uint8Array, key?: Uint8Array): Uint8Array =>
+    key ? blake2b(message, { dkLen: 32, key }) : blake2b(message, { dkLen: 32 });
+
+/**
+ * Derive 32 bytes of deterministic entropy for a terminal session, scoped to the
+ * calling product and a caller key (RFC-0007 layers 2 + 3).
+ *
+ * Same wallet + product + key ⇒ same bytes; any difference ⇒ uncorrelated
+ * entropy. Because it derives from the wallet (not the device), the entropy is
+ * recreatable after device loss as long as the wallet is recoverable.
+ *
+ * @param session   - A QR-paired {@link UserSession}. Must carry
+ *   `rootEntropySource` (present since host-papp 0.8.6 / RFC-0007).
+ * @param productId - The calling product's dotNS identifier, e.g. `"my-app.dot"`.
+ *   The host scopes entropy per product; pass the same identifier the host would.
+ * @param key       - Caller key, 1..32 bytes (the layer-3 BLAKE2b key).
+ * @returns 32 bytes of derived entropy.
+ * @throws if the session lacks `rootEntropySource`, or `key` is not 1..32 bytes.
+ */
+export function deriveEntropy(
+    session: UserSession,
+    productId: string,
+    key: Uint8Array,
+): Uint8Array {
+    const rootEntropySource: Uint8Array | undefined = session.rootEntropySource;
+    if (!rootEntropySource || rootEntropySource.length !== ROOT_ENTROPY_LEN) {
+        throw new Error(
+            "deriveEntropy: session is missing rootEntropySource; re-pair with an RFC-0007 host",
+        );
+    }
+    if (key.length === 0 || key.length > 32) {
+        throw new Error(`deriveEntropy: key must be 1..32 bytes, got ${key.length}`);
+    }
+    const perProduct = b2(rootEntropySource, b2(textEncoder.encode(productId)));
+    return b2(perProduct, key);
+}
+
+if (import.meta.vitest) {
+    const { test, expect, describe } = import.meta.vitest;
+
+    describe("deriveEntropy", () => {
+        const rootEntropySource = new Uint8Array(32).fill(1);
+        const session = { rootEntropySource } as unknown as UserSession;
+        const toHex = (b: Uint8Array) =>
+            Array.from(b, (x) => x.toString(16).padStart(2, "0")).join("");
+
+        test("matches host-container deriveProductEntropyFromSource (golden vector)", () => {
+            // Computed with @novasamatech/host-container@0.8.9's
+            // deriveProductEntropyFromSource(fill(1), "my-app.dot", [1,2,3,4]).
+            expect(toHex(deriveEntropy(session, "my-app.dot", new Uint8Array([1, 2, 3, 4])))).toBe(
+                "993750d5f3f4b941cef5a8084fdd0bcd6a6946fdc0e1fe87c0c575fe65e7dc03",
+            );
+        });
+
+        test("is 32 bytes, deterministic, and product- and key-scoped", () => {
+            const key = new Uint8Array([1, 2, 3, 4]);
+            const a = deriveEntropy(session, "my-app.dot", key);
+            expect(a).toHaveLength(32);
+            expect(deriveEntropy(session, "my-app.dot", key)).toStrictEqual(a);
+            expect(deriveEntropy(session, "other.dot", key)).not.toStrictEqual(a);
+            expect(deriveEntropy(session, "my-app.dot", new Uint8Array([9]))).not.toStrictEqual(a);
+        });
+
+        test("different root entropy yields uncorrelated entropy", () => {
+            const other = {
+                rootEntropySource: new Uint8Array(32).fill(2),
+            } as unknown as UserSession;
+            const key = new Uint8Array([1, 2, 3, 4]);
+            expect(deriveEntropy(other, "my-app.dot", key)).not.toStrictEqual(
+                deriveEntropy(session, "my-app.dot", key),
+            );
+        });
+
+        test("rejects a key outside 1..32 bytes", () => {
+            expect(() => deriveEntropy(session, "my-app.dot", new Uint8Array(0))).toThrow(
+                /1\.\.32/,
+            );
+            expect(() => deriveEntropy(session, "my-app.dot", new Uint8Array(33))).toThrow(
+                /1\.\.32/,
+            );
+        });
+
+        test("throws when the session has no rootEntropySource", () => {
+            expect(() =>
+                deriveEntropy({} as UserSession, "my-app.dot", new Uint8Array([1])),
+            ).toThrow(/rootEntropySource/);
+        });
+    });
+}
--- a/product-sdk/packages/terminal/src/index.ts
+++ b/product-sdk/packages/terminal/src/index.ts
@@ -51,6 +51,12 @@ export type {
 // Session helpers
 export { waitForSessions } from "./sessions.js";
 
+// Entropy derivation (RFC-0007). Client-side product-entropy derivation from a
+// paired session, at parity with `@parity/product-sdk-host`'s `deriveEntropy`
+// (same bytes for the same wallet + context) but computed locally — no host
+// round-trip — since the session already carries `rootEntropySource`.
+export { deriveEntropy } from "./entropy.js";
+
 // QR Encoding
 export { renderQrCode } from "./qr-encode.js";
 export type { QrRenderOptions } from "./qr-encode.js";
--- a/product-sdk/pending-changesets/terminal-derive-entropy.md
+++ b/product-sdk/pending-changesets/terminal-derive-entropy.md
@@ -1,0 +1,14 @@
+---
+"@parity/product-sdk-terminal": minor
+---
+
+Add `deriveEntropy(session, productId, key)` — client-side RFC-0007 product-entropy derivation for terminal (QR/SSO) sessions, byte-for-byte identical to the host's `host_derive_entropy` (i.e. `@novasamatech/host-container`'s `deriveProductEntropyFromSource`).
+
+The paired `UserSession` carries `rootEntropySource` (RFC-0007 layer 1), so layers 2 and 3 are computed locally with no host round-trip:
+
+```
+perProduct = blake2b256(rootEntropySource, key = blake2b256(utf8(productId)))
+entropy    = blake2b256(perProduct,        key = key)
+```
+
+Entropy derived here matches what an in-container app gets from `@parity/product-sdk-host`'s `deriveEntropy` for the same wallet + product + key, so entropy-derived keys interoperate across web and terminal clients. A golden-vector test pins the construction against `host-container`. (issue #254)


ReinhardHatko:
Closing — the implementation here is **incorrect** and should not merge.

It derives entropy as `khash(rootEntropySource, context)` (blake2b keyed by `rootEntropySource`), on the assumption that this reproduces the host's `host_derive_entropy`. Empirical testing (d3pot wallet-derived repo encryption, CLI ↔ web) shows it does **not**: for the same wallet + context, the terminal-derived x25519 pubkey differs from the one derived in-container via `@parity/product-sdk-host`'s `deriveEntropy`.

Constructions ruled out (both mismatch the host):
- `blake2b(context, key=rootEntropySource)` (this PR)
- `blake2b(rootEntropySource, key=context)` (the swap — the host validates `context` as a ≤32-byte key, which suggested this)

Root problem: `host_derive_entropy` is a **host-side RPC** (`@novasamatech/host-api` → `transport.request('host_derive_entropy', …)`); its algorithm isn't in any published SDK package or the specs, so a client-side reimplementation can't be verified to match and is a silent-mismatch footgun.

The underlying need still stands — see #254, which I'm reframing accordingly.

ReinhardHatko:
Reopened as a **draft** — parking, not abandoning.

The public surface (`deriveEntropy(session, context)` on `product-sdk-terminal`) is still what we want; only the derivation *internals* are wrong (the local `khash` doesn't match the host — details above). Once we have the exact `host_derive_entropy` construction, or decide to invoke it as a session RPC (see #254 option a), this branch is where the fix lands. Keeping it in draft so the branch and discussion stay put.

ReinhardHatko:
Updated with the **correct** derivation (still draft).

The original `khash(rootEntropySource, context)` was a single keyed-BLAKE2b layer and does **not** match the host — verified empirically in d3pot (CLI-derived vs web/host-derived x25519 pubkeys differed). The real `host_derive_entropy` (per `@novasamatech/host-container`'s `deriveProductEntropyFromSource`, and the host handler in `dotli-community` `packages/ui/src/container.ts`) is **two layers, scoped by the calling product**:

```
perProduct = blake2b256(rootEntropySource, key = blake2b256(utf8(productId)))
entropy    = blake2b256(perProduct,        key = key)
```

Changes:
- New signature `deriveEntropy(session, productId, key)` (the out-of-container analogue of `deriveProductEntropyFromSource`; `key` must be 1..32 bytes — it's the layer-3 key).
- Verified **byte-identical** to `host-container` and **end-to-end against a real host** (web now decodes a CLI-created private repo with no pasted key).
- Golden-vector test pins the construction so it can't silently drift.

Open question for reviewers: this re-derives the RFC-0007 scheme client-side. To avoid two implementations drifting, `deriveProductEntropyFromSource` ideally moves to a shared crypto package that both `host-container` and `terminal` import — happy to go that route instead. Leaving as draft pending that call.

product-sdk/packages/terminal/src/entropy.ts
```diff
@@ -0,0 +1,123 @@
...
+ * calling product and a caller key (RFC-0007 layers 2 + 3).
+ *
+ * Same wallet + product + key ⇒ same bytes; any difference ⇒ uncorrelated
+ * entropy. Because it derives from the wallet (not the device), the entropy is
+ * recreatable after device loss as long as the wallet is recoverable.
+ *
+ * @param session   - A QR-paired {@link UserSession}. Must carry
+ *   `rootEntropySource` (present since host-papp 0.8.6 / RFC-0007).
+ * @param productId - The calling product's dotNS identifier, e.g. `"my-app.dot"`.
+ *   The host scopes entropy per product; pass the same identifier the host would.
```
Imod7, line 49: Please document the productId contract - it's now the only thing between a caller and a silent mismatch. The host doesn't take it from the caller - it computes ${label}.dot from the iframe's deployment label (dotli's packages/ui/src/container.ts passes exactly that into deriveProductEntropyFromSource). So the correct value is that deployment's label, and it differs per deployment: production, each PR preview, and local dev (localhost:5173.dot) all yield a different id. A CLI hardcoding the production id derives valid-but-different bytes against a preview, with no error anywhere.

Please spell out that per-deployment rule, add "must match the productId you pass to requestResourceAllocation (#262)", and say why it's required rather than defaulting to adapter.appId.

product-sdk/packages/terminal/src/entropy.ts
```diff
@@ -0,0 +1,123 @@
...
+        const toHex = (b: Uint8Array) =>
+            Array.from(b, (x) => x.toString(16).padStart(2, "0")).join("");
+
+        test("matches host-container deriveProductEntropyFromSource (golden vector)", () => {
+            // Computed with @novasamatech/host-container@0.8.9's
+            // deriveProductEntropyFromSource(fill(1), "my-app.dot", [1,2,3,4]).
+            expect(toHex(deriveEntropy(session, "my-app.dot", new Uint8Array([1, 2, 3, 4])))).toBe(
+                "993750d5f3f4b941cef5a8084fdd0bcd6a6946fdc0e1fe87c0c575fe65e7dc03",
+            );
+        });
```
Imod7, line 87: The golden vector pins this file against itself. Nothing here can reproduce the hex, and it can't catch host-container drift - which is what the NOTE above says we're guarding against.

Cheapest fix, and stronger than minting a new hex: reuse the vectors the other implementations already share with each other. truapi's Rust tests carry vectors copied byte-for-byte from polkadot-app-ios-v2's ProductRootEntropyDeriverTests. I ran your math against both sets and got 6/6 matches - e.g. secret = bytes 0..31, productId "myapp.dot", key [1] gives 4bafd6a34182959bad8914dcff88c6b6842d551d6f0067afbd407e9584223404. They start from the raw secret rather than rootEntropySource, so the test needs the layer-1 step first (blake2b256 keyed with "product-entropy-derivation") - one line. No new dependency, and it pins us against Rust and Swift rather than against a single TS package.

@novasamatech/host-container is also public on npm, so the devDependency plus an entropy.interop.test.ts asserting equality against the real deriveProductEntropyFromSource is worth having too - same pattern as testing.interop.test.ts. The shared vectors prove we match today; the interop test catches future drift on its own. Keep the golden vector as well; it documents the expected bytes for a human reader.

product-sdk/packages/terminal/src/entropy.ts
```diff
@@ -0,0 +1,123 @@
...
+ *
+ * @param session   - A QR-paired {@link UserSession}. Must carry
+ *   `rootEntropySource` (present since host-papp 0.8.6 / RFC-0007).
+ * @param productId - The calling product's dotNS identifier, e.g. `"my-app.dot"`.
+ *   The host scopes entropy per product; pass the same identifier the host would.
+ * @param key       - Caller key, 1..32 bytes (the layer-3 BLAKE2b key).
+ * @returns 32 bytes of derived entropy.
+ * @throws if the session lacks `rootEntropySource`, or `key` is not 1..32 bytes.
+ */
+export function deriveEntropy(
```
Imod7, line 54: `deriveEntropy` reads nothing account-identifying, and terminal supports multiple paired sessions (`waitForSessions`). Pass the wrong one and you get valid entropy for a different wallet: the intended user can't decrypt, the other wallet's holder can. Explicit `session` is the right design - could you add a `@remarks` telling callers to pin identity via the exported `sessionRootPublicKey(session)` before deriving long-lived keys?  Same block could note that the return value is raw key material - don't log it or persist it unwrapped.

product-sdk/packages/terminal/src/entropy.ts
```diff
@@ -0,0 +1,123 @@
...
+            const other = {
+                rootEntropySource: new Uint8Array(32).fill(2),
+            } as unknown as UserSession;
+            const key = new Uint8Array([1, 2, 3, 4]);
+            expect(deriveEntropy(other, "my-app.dot", key)).not.toStrictEqual(
+                deriveEntropy(session, "my-app.dot", key),
+            );
+        });
+
+        test("rejects a key outside 1..32 bytes", () => {
```
Imod7, line 108: [nice-to-have]: The 1..32 tests only probe outside the boundary (0 and 33). Could you also assert that 1 and 32 are *accepted*? An off-by-one to `>= 32` would pass the suite while rejecting the max-length key the host allows. Same for `rootEntropySource.length !== 32` on line 60: only the missing-field case is covered. Note `@noble` accepts keys up to 64 bytes, so line 65 is the only thing enforcing the host's bound — a compatibility contract, not defensive boilerplate.

Worth knowing that the implementations already disagree at exactly this boundary: iOS and Android only check `<= 32` and accept an empty key, where you, truapi's Rust and host-container all reject it. So these two bounds are the one place a cross-platform mismatch is already live, which is more reason to assert both ends rather than just outside them.

product-sdk/packages/terminal/src/entropy.ts
Imod7, line 54: #262 made `productId` an optional option defaulting to `adapter.appId`; here it's a required positional. Required is *correct* for entropy — a wrong default produces undecryptable data rather than a loud error, and #262 notes d3pot's `appId` differs. Could you add one JSDoc line saying it's deliberate, so nobody "fixes" it into a default?

product-sdk/packages/terminal/src/entropy.ts
```diff
@@ -0,0 +1,123 @@
...
+ * `blake2b256_keyed(rootAccountSecret, "product-entropy-derivation")`), so
+ * layers 2 and 3 are computed locally with no host round-trip:
+ *
+ *   perProduct = blake2b256(rootEntropySource, key = blake2b256(utf8(productId)))
+ *   entropy    = blake2b256(perProduct,        key = key)
+ *
+ * NOTE: this re-derives the RFC-0007 scheme locally and MUST stay byte-identical
+ * to `host-container`'s `deriveProductEntropyFromSource`. The golden-vector test
+ * below guards against drift, but the derivation ideally belongs in a shared
+ * crypto package that both host-container and terminal import (see PR #260).
```
Imod7, line 23: On your open question: keep the derivation here, but split it in three. Doesn't block the PR.

The shared package the NOTE wants can't exist — this algorithm has five implementations in three languages (host-container, truapi's Rust `host_logic/entropy.rs`, Android's `RealDeriveEntropyUseCase.kt`, iOS's `ProductRootEntropyDeriver.swift`, and this one). Swift and Kotlin can't import a TS package, so the shared artifact is the spec (host-spec §C.8) and the safety net is shared test vectors — which Rust and iOS already swap byte-for-byte. Importing host-container here isn't a shortcut either — nothing `terminal` already depends on exports the derivation (host-papp carries `rootEntropySource`, host-api only the error type), so it trades the novasama import we're shedding for a new one.

`keys/src/product-account.ts` already answers the in-repo version of the question. Nothing here needs doing in this PR — an issue is enough for now, so the NOTE has something to link to, and the move lands as its own PR later (~40 lines, no behaviour change):

- `blake2b256Keyed` → `crypto/src/hashing.ts`: the missing sibling of the `blake2b256` already there, and we're the only one of the five platforms without a keyed BLAKE2b in its crypto layer. (`utils/src/hashing.ts` has a byte-identical copy of those three, so worth settling which is canonical - I'd say `crypto`, where `keys` already imports `blake2b256` from.)
- the pure derivation → `keys/src/product-entropy.ts`, beside `product-account.ts` (same species: client-side mirror of a wallet-side derivation, header listing its mirrors, frozen-vector test). `terminal` already depends on `keys`, so this adds no dependency.
- `deriveEntropy(session, productId, key)` stays exactly as you wrote it — the session wrapper. Public API and changeset unchanged.

And please turn the NOTE into the decision plus an issue link - an open question in merged code stays open. While rewriting it, cite host-spec C.8 rather than host-container's implementation (mirrors drift; these five already disagree on whether an empty key is legal), and steal iOS's clarification that layer 1's input is raw BIP-39 entropy, not the 64-byte PBKDF2 seed. Happy to open the issue and write up the full reasoning there.

Imod7 (commented):
Three more notes on top of the inline comments — none of them blocking.

**1. `dist` ships the test suites.** Pre-existing, but this PR adds to it: `terminal/dist/index.js` ships all 14 in-source `describe` blocks, including `entropy.ts`'s golden-vector hex. The `define: {"import.meta.vitest": "undefined"}` rewrites the condition, but esbuild only drops the dead block with `treeshake: true` - `packages/signer` has the flag and ships 0 of them. `fix/tsup-treeshake-toplevelawait` adds the flag to 14 packages but **not** `terminal`.

→ **Action:** add `treeshake: true` to `packages/terminal/tsup.config.ts`, or fold `terminal` into that branch.

**2. "Closes #254" overstates what ships.** This adds `deriveEntropy` but not #254's `getEncrSecret` / `getEncrPublicKey` / `createSharedSecret`. Fine as a cut - entropy is the drift-prone part, and `@parity/product-sdk-crypto` already has `boxEncrypt`/`sealedBoxEncrypt` for the rest. But as written, the issue closes with its encryption helpers undelivered.

→ **Action:** either narrow the description to "Refs #254" and open a follow-up for the three helpers, or add them here.

**3. The RFC-0007 product-entropy derivation is specified, not just published.** host-spec §C.8 defines it - the three keyed BLAKE2b-256 layers over session secret, productId and caller key - at [`spec/C-account-derivation.md`](https://github.com/paritytech/host-spec/blob/adb3989208ae1c2107dbf0159611353e6989422c/spec/C-account-derivation.md?plain=1#L129-L147), which truapi's Rust implementation cites at a pinned commit. Worth pointing at the spec rather than at host-container: mirrors drift, and these five already disagree on whether an empty caller key is legal.

→ **Recommendation:** cite host-spec §C.8 instead of the implementation in the two places that name it - the changeset (`pending-changesets/terminal-derive-entropy.md`: *"byte-for-byte identical to the host's `host_derive_entropy`"*, *"pins the construction against `host-container`"*), which ships to the published changelog, and the module header comment.

product-sdk/packages/terminal/src/entropy.ts
ReinhardHatko: Done — and thanks, this one turned out to be load-bearing in a way neither of us expected.

Documenting the contract surfaced that the reference point you suggested is wrong. `requestResourceAllocation(session, adapter, resources, options)` takes no `productId` at all — it reads `adapter.appId` and forwards it as `callingProductId` (`host.ts:100,108`). And `appId` is a *local* identifier: host-papp only ever uses it as a storage namespace (`createLocalStorageAdapter`, `createUserSecretRepository`, `createAllowanceRepository`, `createDeviceIdentityStore`), plus terminal's own `getAes(appId) = blake2b(appId, 16)` for the allowance cache at rest.

So the two are deliberately different values — the package's own example pairs `createTerminalAdapter({ appId: "my-cli" })` with `getBulletinSigner(adapter, "my-cli.dot")` (`allowance.ts:77-79`). My first draft of the doc said "pass the same identifier you pass to `requestResourceAllocation`", which would have steered callers to `"my-cli"` while the format rule five lines above says `"my-cli.dot"`. Precisely the silent mismatch the paragraph exists to prevent.

The doc now points at the in-container deployment's identifier, says explicitly that it is **not** `adapter.appId`, and flags `requestResourceAllocation` as an unreliable reference point. Per-deployment rule and the `requestResourceAllocation` link to #262 are both in.

One small correction to your example: host-spec §C.7 gives local dev as `localhost:<port>` with **no** `.dot` suffix, so it would be `localhost:5173`, not `localhost:5173.dot`. Documented that way. (Worth noting `@parity/product-sdk-signer`'s `dappName` path cannot express it — `host.ts:551-553` appends `.dot` unconditionally when missing — which is a separate gap on that convenience path.)

product-sdk/packages/terminal/src/entropy.ts
ReinhardHatko: You were right, and the fix ended up better than either of us proposed.

First: I verified your vector before changing anything. `secret = bytes 0..31`, `productId "myapp.dot"`, `key [1]`, with layer 1 applied first → `4bafd6a34182959bad8914dcff88c6b6842d551d6f0067afbd407e9584223404`. Exact match. That also disambiguates the `blake2b256_keyed(rootAccountSecret, "product-entropy-derivation")` notation: message is the secret, key is the context string (the reversed orientation gives `50d7d4f8…`).

Rather than transcribe from the Rust and Swift test files, there's a canonical machine-readable suite: [`useragent-kit@92641ed conformance/rfc0007-product-entropy-v1.json`](https://github.com/paritytech/useragent-kit/blob/92641ed2cba46e39e9682230b91252cd21dfa396/conformance/rfc0007-product-entropy-v1.json). Its own description names the consumers: `crates/host-encoding` (Rust, canonical), `sdk-android`, `sdk-swift`, plus external hosts. All 8 vectors are now in the test block and **8/8 pass**. They start from raw BIP-39 entropy, so each case applies layer 1 first — the one-line step you described.

I did **not** add the `@novasamatech/host-container` devDependency + `entropy.interop.test.ts`, for the reason you gave yourself in the split comment: nothing terminal already depends on exports the derivation, so it would trade the novasama import we are shedding for a new one. Pinning directly against the Rust/Kotlin/Swift vectors is strictly stronger than pinning against one TS mirror — it catches drift in the mirror too. Happy to add the interop test as well if you still want belt-and-braces.

Golden vector kept, as you asked, relabelled to say honestly what it does: pins layers 2+3 from a session-supplied `rootEntropySource`, which is the exact shape terminal sees.

product-sdk/packages/terminal/src/entropy.ts
ReinhardHatko: Added, both halves. The `@remarks` block now reads:

> This reads nothing account-identifying, and a terminal may hold several paired sessions (see `waitForSessions`). Passing the wrong one yields valid entropy for a *different* wallet: the intended user cannot decrypt, and the other wallet's holder can. Pin the identity with `sessionRootPublicKey` before deriving long-lived keys.
>
> The return value is raw key material — do not log it or persist it unwrapped.

Written with plain backticks rather than `{@link}` on purpose: neither symbol is imported here, so a TSDoc link would emit an unresolved-reference warning in the typedoc build and render as literal text — degrading exactly the two cross-references that matter most in that paragraph.

product-sdk/packages/terminal/src/entropy.ts
ReinhardHatko: Added — both bounds now asserted as *accepted*, not just rejected outside. Also added the missing `rootEntropySource` wrong-length case you noted on line 60; it previously only covered the missing-field path.

On the cross-platform divergence: the conformance suite settles it, and in our favour. It lists `key-empty` under `invalid`, with the rationale spelled out:

> Empty keys must be rejected: some BLAKE2b implementations silently degrade an empty MAC key to unkeyed mode

and host-spec §C.8 is normative on it — *"`caller_key` MUST be 1–32 bytes. Zero-length and >32 bytes MUST be rejected with an error returned to the product."*

So this implementation, truapi's Rust and host-container are conformant, and the iOS/Android leniency you found is a bug on their side rather than a genuine spec ambiguity. That flips which end needs to move — worth an upstream issue, but nothing to change here.

Confirmed your `@noble` point too: it accepts keys of length 1..64 (and rejects 0 on its own, with a worse message). So the `> 32` half of the check really is the only thing enforcing the host's bound. Comment on that test now says so.

product-sdk/packages/terminal/src/entropy.ts
ReinhardHatko: Added, and it earned its keep almost immediately — a `/code-review` pass on the updated branch argued for exactly the change you predicted: take a `TerminalAdapter` and default the scope from `adapter.appId`, to match `requestResourceAllocation` / `getBulletinSigner`. I declined it and cited your reasoning.

Your call is right, and documenting the `productId` contract made the evidence concrete: `appId` is only ever a *local* storage namespace in host-papp, and the package's own example deliberately pairs `appId: "my-cli"` with productId `"my-cli.dot"`. Defaulting would bake the mismatch in rather than remove it — and unlike an allocation request, the wrong scope here returns 32 well-formed bytes that simply never decrypt.

The JSDoc now says so explicitly:

> Required and positional **by design** — do not give this a default. Unlike an allocation request, a wrong product scope here fails silently: it yields well-formed entropy that simply cannot decrypt anything.

product-sdk/packages/terminal/src/entropy.ts
ReinhardHatko: Taken, all of it. The `NOTE` is now a decision rather than an open question:

> The scheme has parallel implementations in five languages — host-container (TS), host-encoding (Rust, canonical), sdk-swift, sdk-android and brevity-flutter (Dart) — so the shared artifact is the spec rather than a package: Swift, Kotlin and Dart cannot import a TS one. […] Cite the spec, not any one implementation: mirrors drift.

followed by the three-way split as a stated follow-up (`blake2b256Keyed` → `crypto`, pure derivation → `keys` beside `product-account.ts`, session wrapper unchanged). I confirmed terminal already depends on `@parity/product-sdk-keys`, so that move adds no dependency, and that `crypto/src/hashing.ts:25` and `utils/src/hashing.ts:25` are byte-identical copies with no keyed variant in either — so the canonical-home question is real.

I have **not** opened the issue, since you offered to and you have the fuller reasoning. The header records the decision but links nothing yet — please do open it and I will add the link.

Also took the iOS clarification: the header now says layer 1's input is raw BIP-39 entropy, 16 or 32 bytes, not the 64-byte PBKDF2 seed. The conformance vectors bear that out directly — they carry both 16- and 32-byte roots (`zero-root-16-bytes` is BIP-39 entropy for a 12-word mnemonic).

One correction to the count, for the record: that is five implementations across *five* languages, not three — Dart and Kotlin/Swift are all distinct. My first draft repeated "three" and a review pass caught it.

ReinhardHatko:
Thanks @Imod7 — all eight points addressed in 38250be. Replies inline; summary of the three non-inline ones, plus two things your review turned up that changed the answer.

### 1. `dist` ships the test suites — fixed

Confirmed exactly as you described, and it was worse than the comment implied. A fresh build shipped **14** `describe` blocks, the golden-vector hex and the error strings into `terminal/dist/index.js`; `signer` (which has the flag) ships **0**; and `terminal` was the *only* package in `packages/*` without it. `treeshake: true` added. After: 14 → 0 describe blocks, both vector hexes gone, runtime code intact, and the test-only `hexToBytes` import stripped too.

One pointer: the terminal fix already exists as your own `6f972cb chore(terminal): treeshake the build`, on `fix/terminal-statement-store-endpoints` rather than `fix/tsup-treeshake-toplevelawait`. I added the identical line here so this PR is self-contained — trivial conflict to resolve whichever lands first.

### 2. "Closes #254" overstates — fixed

Now `Refs #254`, with the cut stated explicitly in the PR body: the encryption helpers (`getEncrSecret` / `getEncrPublicKey` / `createSharedSecret`) are not in this PR and the issue stays open.

### 3. Cite the spec, not the implementation — done, and it resolved more than citation

Module header and changeset now cite host-spec §C.8. Reading it settled two things a mirror could not have:

- **The empty-key question is not an ambiguity.** §C.8 is normative that zero-length and >32-byte keys MUST be rejected, and the conformance suite lists `key-empty` under `invalid` with the reason (*"some BLAKE2b implementations silently degrade an empty MAC key to unkeyed mode"*). So the iOS/Android leniency is non-conformance, not a live spec disagreement. Details in the boundary thread.
- **There is a canonical vector suite**, which replaced the self-referential golden hex entirely — 8/8 passing, shared with the Rust, Kotlin and Swift implementations. Details in the golden-vector thread.

### Two corrections your review produced indirectly

- Documenting the `productId` contract exposed that `requestResourceAllocation` takes no `productId` and forwards `adapter.appId` — a purely local storage namespace. My first draft of the doc pointed there and would have induced the exact silent mismatch the paragraph warns about. Corrected; see that thread.
- The "five implementations in three languages" framing is five languages, not three (Dart and Kotlin/Swift are distinct). Fixed in the header.

### Left open deliberately

The three-way split issue — you offered to open it and write up the reasoning, so I have not duplicated it. The module header records the decision and is ready for the link.

Also unfiled, and outside this PR: `@parity/product-sdk-signer` soft-degrades `connect()` to `[]` accounts on *any* `getProductAccount` failure (`providers/host.ts:558-575`), with a warning attributing it to an unregistered identifier. paritytech/polkadot-android-community#123 is a counterexample — a host that never answers `ProductSubtreeRequest` produces a byte-identical outcome, pointing developers at their dappName instead of at the peer. Worth its own issue if you agree.

product-sdk/packages/terminal/src/entropy.ts
ReinhardHatko: Correction to my reply above — I went to write the upstream issue and found I had this wrong.

I said the iOS/Android leniency was "a bug on their side." It is not. Both **do** reject empty caller keys; they just validate at a different layer than we do.

The derivation functions really do check only the upper bound, as you found:

```kotlin
// sdk-android ProductEntropy.kt
require(key.size <= PRODUCT_ENTROPY_MAX_KEY_SIZE) { ... }
```
```swift
// sdk-swift ProductEntropy.swift
guard key.count <= productEntropyMaxKeySize else { throw ... }
```

But the lower bound is enforced one level down, in the shared Rust primitive they both bind to:

```rust
// crates/host-encoding/src/statement_store.rs
pub fn blake2b_256_keyed(key: &[u8], data: &[u8]) -> Result<[u8; 32], String> {
    // Reject empty keys explicitly — some Blake2b implementations silently
    // degrade to unkeyed mode for key.len()==0, which would be a security bug.
    if key.is_empty() {
        return Err("blake2b key must not be empty".into());
    }
```

and Kotlin reaches it through UniFFI rather than reimplementing it:

```kotlin
// sdk-android Encoding.kt
fun blake2b256Keyed(key: ByteArray, data: ByteArray): ByteArray =
    uniffi.useragent_mobile.blake2b256Keyed(key, data)
```

Swift follows the same wrapper-over-UniFFI shape. Worth noting the canonical Rust `derive_product_entropy_from_source` *also* only checks `> MAX` at the derivation layer and leans on the primitive for the lower bound — so on this axis we are the outlier for checking both bounds up front, not them for checking one.

Net: no cross-platform divergence in observable behaviour, and no upstream issue needed. The only real difference is error quality — they surface a generic "blake2b key must not be empty" from the primitive where we surface a product-entropy-specific message. A nit at most.

What stands unchanged from that reply: both bounds are normative, the conformance suite lists `key-empty` under `invalid`, and our explicit `> 32` check is still the only thing enforcing the upper bound here given `@noble` accepts 1..64. The boundary tests you asked for are in.

product-sdk/packages/terminal/src/entropy.ts
ReinhardHatko: Nudge on this one, since the branch is now current with `main` and #260 is otherwise ready for approval.

The module header records the decision but still links no issue — the "open question in merged code stays open" problem you flagged. If you open it, I will add the link before this merges; if you would rather I write it up from your comment, say so and I will, with attribution.

Two things from the merge that are worth folding into it:

- `treeshake: true` for `terminal` had already landed on `main` (your `6f972cb`), so my line in `38250be` was a duplicate — the merge collapsed it to one. Nothing to do, just noting the fix is yours and already in.
- The `crypto` vs `utils` question is still live and unchanged: `crypto/src/hashing.ts:25` and `utils/src/hashing.ts:25` are byte-identical `blake2b256` implementations and neither has a keyed variant. `keys` already imports from `crypto`, which is the argument for making `crypto` canonical.

Also relevant to the scope: the canonical Rust (`crates/host-encoding`) keeps the keyed-BLAKE2b primitive separate from the derivation and enforces the empty-key rejection in the primitive — the same split you proposed here. Worth citing as precedent.

PR #261, opened 2026-07-20
feat(tx): surface pre-inclusion validity failures as typed TxValidityError

ReinhardHatko:
## What
A pre-inclusion validity/submission failure (e.g. `InvalidTransaction::Payment` — submitter can't pay or isn't authorized) previously surfaced as an opaque error — either a base `TxError` wrapping raw JSON, or a `TxDispatchError` whose message was the placeholder `"unknown error"` (that field only exists for an *included-and-failed* tx).

## Change
- New `TxValidityError extends TxError` (inherits the `SdkError` marker), carrying the raw failure payload on `.reason` + a human-readable `.formatted`.
- polkadot-api's `InvalidTxError` subscription rejection is now detected and surfaced as `err(TxValidityError)` with the decoded reason (`{ type: "Invalid", value: { type: "Payment" } }` → `"Invalid.Payment"`).
- `withRetry`'s `isNonRetryable` treats `TxValidityError` as non-retryable (these are deterministic for the same account state).

Implements proposal cluster **E2** (error classification: opaque tx failures). Changeset included (tx minor + umbrella minor).


product-sdk/pending-changesets/tx-validity-error.md
```diff
@@ -0,0 +1,32 @@
+---
```
Imod7, line 1: Please move the changeset in `pending-changesets` folder

product-sdk/packages/tx/src/submit.ts
```diff
@@ -45,6 +47,33 @@ function buildTxResult(
...
+/**
+ * Classify an `ok: false` tx event: with a `dispatchError` it's an on-chain
+ * dispatch failure; without one the failure happened *before* inclusion
+ * (validity/submission) — `dispatchError` only exists for an
+ * included-and-failed transaction, so nothing more specific is available on
+ * the event itself.
+ */
+function classifyFailure(event: { dispatchError?: unknown }, formatted: string): TxError {
+    return event.dispatchError == null
+        ? new TxValidityError(event, "validity/submission failure (no dispatch error)")
```
Imod7: **[nit]** `classifyFailure` discards its `formatted` arg in the no-`dispatchError` branch and hardcodes a near-duplicate of `formatDispatchError`'s string. Not a mismatch (both say the same thing) - just two literals to keep in sync. Consider reusing `formatted` or a shared constant.

product-sdk/packages/tx/src/submit.ts
```diff
@@ -129,7 +159,7 @@ export async function submitAndWatch(
                                     formatted,
                                     block: event.block,
                                 });
-                                settleErr(new TxDispatchError(event.dispatchError, formatted));
+                                settleErr(classifyFailure(event, formatted));
```
Imod7, line 160: **[nice to have]** `classifyFailure` is called from the best-block (L162) and finalized (line 213) paths, where the tx *was* included - but it emits `TxValidityError`, whose message asserts *"failed before inclusion."* For that defensive branch a neutral message would avoid a misleading log. (The genuinely pre-inclusion path at L242 is correctly labelled.)

product-sdk/packages/tx/src/errors.ts
```diff
@@ -45,6 +45,36 @@ export class TxDispatchError extends TxError {
...
+ * These failures carry no `dispatchError` (that field only exists for a
+ * transaction that was included and then failed to dispatch), so they are
+ * reported as a distinct type instead of a {@link TxDispatchError} with a
+ * placeholder string. The raw failure payload is preserved on
+ * {@link TxValidityError.reason} for programmatic inspection.
+ */
+export class TxValidityError extends TxError {
+    /**
+     * Raw underlying failure payload: the decoded validity error from
+     * polkadot-api's `InvalidTxError` (e.g. `{ type: "Invalid", value:
```
Imod7: This looks like it is breaking the docs CI build.

The `{ type: "Invalid", value: { type: "Payment" } }` example is split across two lines inside the backticks. The docs builder can't parse a code snippet that wraps, so it crashes on api/tx/index.mdx.

Fix: put the example on one line:
```
* polkadot-api's `InvalidTxError`
* (e.g. `{ type: "Invalid", value: { type: "Payment" } }`), or the failed
* tx event when the node reported failure without any payload.
```

ReinhardHatko: Fixed — the `InvalidTxError` example is now on a single line so the docs generator can parse the inline code span. The `docs: Verify Generate & Build` check is green again.

product-sdk/packages/tx/src/submit.ts
ReinhardHatko: Good catch. Rather than only softening the message, I reclassified this branch. `classifyFailure` runs only on *included* events, so an included failure with no `dispatchError` is now a `TxDispatchError` (neutral message, still non-retryable), and `TxValidityError` is reserved for the genuine pre-inclusion path (the `InvalidTxError` subscription-error handler). This is a slightly bigger diff than a message-only tweak — flagging so it is expected. Two tests updated accordingly.

product-sdk/packages/tx/src/submit.ts
ReinhardHatko: Addressed — the no-`dispatchError` branch is now a `TxDispatchError` (see the reply on the L160 thread), so `formatted` is no longer discarded into a duplicated literal and there is no second string to keep in sync with `formatDispatchError`.

Imod7 (approved):
LGTM! 
If we could just update the changeset bc it still lists the `no-dispatchError` path as producing `TxValidityError`; This is now changed it to `TxDispatchError`.

PR #262, opened 2026-07-20
feat(terminal): explicit productId on the host allocation APIs

ReinhardHatko:
## What
The terminal `/host` allocation APIs keyed everything on `adapter.appId` — both the wire `callingProductId` and the slot-cache namespace. When a CLI's storage `appId` differs from the product id it signs as, the wallet mints PGAS and auto-maps the **wrong** on-chain account (it signs as the product account but funds/maps the appId-derived one). Consumers worked around it with a `{ ...adapter, appId: productId }` spread.

## Change
Add an optional `productId` (default `adapter.appId`, strictly additive) to `requestResourceAllocation`, `getCachedAllocation`, `ensureSlotAccountSigner` and `createSlotAccountSigner`, threaded from one resolved variable to **both** the wire `callingProductId` and the cache namespace (so they can't drift). Mirrors the explicit `productId` the signer-side helpers already take.

Implements proposal cluster **C2**. Changeset included (terminal minor + umbrella minor).

### DIFF
--- a/product-sdk/pending-changesets/terminal-allocation-explicit-product-id.md
+++ b/product-sdk/pending-changesets/terminal-allocation-explicit-product-id.md
@@ -17,3 +17,11 @@ auto-mapped on the wrong on-chain account, and brings the allocation side in
 line with the signer/read side (`createSessionSignerForAccount`,
 `getBulletinSigner`), which already takes an explicit `productId`. Consumers
 can delete their `{ ...adapter, appId: productId }` spread workarounds.
+
+> **Warning:** thread the **same** `productId` through **all four** allocation
+> APIs — `requestResourceAllocation`, `getCachedAllocation`,
+> `ensureSlotAccountSigner`, and `createSlotAccountSigner`. Deleting the
+> `{ ...adapter, appId }` spread without passing `productId` everywhere silently
+> reintroduces the wrong-account PGAS mint (allowance minted / auto-mapped on
+> the account derived from `adapter.appId` instead of your product's account),
+> which is exactly the footgun this change closes.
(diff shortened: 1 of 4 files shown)


product-sdk/pending-changesets/terminal-allocation-explicit-product-id.md
```diff
@@ -0,0 +1,19 @@
+---
```
Imod7, line 1: Please move the changeset in `pending-changesets` folder

product-sdk/pending-changesets/terminal-allocation-explicit-product-id.md
```diff
@@ -0,0 +1,19 @@
...
+"@parity/product-sdk-terminal": minor
+"@parity/product-sdk": minor
```
Imod7, line 3: The changeset tells consumers to *delete* the `{ ...adapter, appId }` workaround but does **not** warn that `productId` must then be threaded to all four calls. That guidance, followed naively, silently reintroduces the wrong-account PGAS mint. Consider adding a relevant note to the changeset.

Imod7 (commented):
LGTM! Just 2 comments on the changeset

product-sdk/pending-changesets/terminal-allocation-explicit-product-id.md
ReinhardHatko: Added a warning to the changeset: consumers must thread the same `productId` through all four allocation APIs — deleting the `{ ...adapter, appId }` spread without it silently reintroduces the wrong-account PGAS mint.

Imod7 (approved):

PR #263, opened 2026-07-20
feat(signer,terminal): typed AllowanceExpiredError at the sign boundary

ReinhardHatko:
## What
When a paired session's statement-store allowance lapses, the terminal session signers surfaced it as a generic `Error`, so consumers had to string-match console output to detect it.

## Change
- New `AllowanceExpiredError extends SignerError` (inherits the `SdkError` marker) in `@parity/product-sdk-signer`, re-exported from terminal.
- Thrown at terminal's sign boundaries (`createTransaction` / `signRaw`) when the failure is a statement-store `NoAllowanceError` (deliberately thrown, per PAPI's `PolkadotSigner` contract).
- Adds `@parity/product-sdk-signer` as a terminal dependency (verified cycle-free).

Implements the **in-repo half** of proposal cluster **B1**; the 240s-hang root cause is upstream in `@novasamatech/host-papp`.

### Known follow-ups (non-blocking, from review)
- In the real host-papp path both sign calls collapse to a flat `new Error(msg)`, so production detection currently relies on the message substring (`instanceof`/cause-walk are dead there) — depends on the wallet forwarding statement-store's exact wording. Worth hardening + recentering the tests on that shape.
- terminal now exports two adjacent names (`AllowanceExpiredError` new, `AllowanceError` from host-papp) — a naming-clarity consideration.

Changeset included (signer + terminal minor + umbrella minor).


product-sdk/pending-changesets/allowance-expired-error.md
```diff
@@ -0,0 +1,27 @@
+---
```
Imod7, line 1: Please move the changeset in `pending-changesets` folder

product-sdk/packages/terminal/src/index.ts
```diff
@@ -17,6 +17,11 @@ export {
     INCOMPLETE_SESSION_MESSAGE,
 } from "./signer.js";
 export type { ProductAccountRef } from "./signer.js";
+// Thrown by the session signers' signTx/signBytes when the paired session's
+// allowance has lapsed (chain-side NoAllowanceError). Authored in
+// @parity/product-sdk-signer — the SDK's signer error taxonomy — and
+// re-exported here so terminal consumers can catch it without a second import.
+export { AllowanceExpiredError, SignerError } from "@parity/product-sdk-signer";
```
Imod7: Re-exporting from `@parity/product-sdk-signer` (the barrel) makes terminal eagerly load `HostProvider` → `@parity/product-sdk-host` → `@novasamatech/host-api` (an optional peer npm won't auto-install) at import time - so raw-Node/npm terminal consumers without that peer crash on import. terminal was previously host-free. Please import from an errors-only subpath instead: add `"./errors"` to signer's `exports` (pointing at a host-free `dist/errors.js`) and import `AllowanceExpiredError`/`SignerError` from `@parity/product-sdk-signer/errors` in both `index.ts:24` and `signer.ts:47`.

ReinhardHatko: Done as suggested — added a host-free `./errors` subpath to signer (its errors module has type-only imports, so `dist/errors.js` pulls in no host code); terminal now imports `AllowanceExpiredError` / `SignerError` from `@parity/product-sdk-signer/errors` in both `index.ts` and `signer.ts`. Verified terminal builds and resolves the subpath without loading `host-api`.

Imod7 (approved):

PR #264, opened 2026-07-20
feat(cloud-storage): Bulletin allowance status read-back

ReinhardHatko:
## What
The SDK could grant Bulletin storage authorizations but exposed no first-class read-back of current allowance state; consumers hand-rolled the extent/expiry math.

## Change
New `getBulletinAllowanceStatus(api, slotAddress)` → `Result<BulletinAllowanceStatus, CloudStorageAuthorizationError>`, composing the existing `checkAuthorization` quota read with a `System.Number` current-block read (run in parallel) to derive `remainingBlocks` and `usable`. `BulletinAllowanceStatus extends AuthorizationStatus`.

Implements proposal cluster **C1a**. Changeset included (cloud-storage minor + umbrella minor).

### DIFF
--- a/product-sdk/packages/cloud-storage/src/authorization.ts
+++ b/product-sdk/packages/cloud-storage/src/authorization.ts
@@ -156,7 +156,11 @@ export async function getBulletinAllowanceStatus(
     const result: BulletinAllowanceStatus = {
         ...status,
         remainingBlocks: Math.max(0, status.expiration - currentBlock),
-        usable: status.authorized && currentBlock < status.expiration,
+        usable:
+            status.authorized &&
+            currentBlock < status.expiration &&
+            status.remainingTransactions > 0 &&
+            status.remainingBytes > 0n,
     };
 
     log.debug("getBulletinAllowanceStatus", {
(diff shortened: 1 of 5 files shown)


product-sdk/pending-changesets/bulletin-allowance-status.md
```diff
@@ -0,0 +1,15 @@
+---
```
Imod7, line 1: Please move the changeset into the `pending-changesets` folder

product-sdk/packages/cloud-storage/src/authorization.ts
```diff
@@ -99,6 +99,76 @@ export async function checkAuthorization(
...
+        // address and cause; the caller decides whether to report it.
+        return err(new CloudStorageAuthorizationError(slotAddress, error));
+    }
+    if (!statusResult.ok) return statusResult;
+    const status = statusResult.value;
+
+    const result: BulletinAllowanceStatus = {
+        ...status,
+        remainingBlocks: Math.max(0, status.expiration - currentBlock),
+        usable: status.authorized && currentBlock < status.expiration,
```
Imod7: `usable` only checks expiry, so a not-yet-expired allowance with `remainingTransactions === 0` / `remainingBytes === 0n` still reports `usable: true` - but the chain rejects such a store with `TooManyTransactions` / `PermanentAllowanceExceeded` (transaction-storage pallet). So quota is a hard gate too, and the changeset's "expiry - the chain's only hard gate" is inaccurate. A consumer gating a store on `usable` would submit a doomed tx. Please fold quota into `usable` (`&& remainingTransactions > 0 && remainingBytes > 0n`) and note callers must still size-check `remainingBytes` against their payload.

ReinhardHatko: Fixed — `usable` now also requires `remainingTransactions > 0 && remainingBytes > 0n` (kept the authorized + expiry checks). Corrected the changeset wording (expiry is not the only hard gate) and noted callers must still size-check `remainingBytes` against their payload. Added two tests for the not-expired-but-quota-exhausted cases.

Imod7 (approved):
LGTM! 
Would be nice to also update in `product-sdk/packages/cloud-storage/src/types.ts`:L41 the comment so it reflects the updated code.

PR #265, opened 2026-07-20
feat(contracts): account-mapping read probe + dry-run origin validation

ReinhardHatko:
## What
Two contracts DX improvements (2 commits):

**C1b — `isContractAccountMapped` read probe.** Extracts the inline pallet-revive mapping check into an exported `isContractAccountMapped(runtime, address)` → `Result<boolean, ContractError>` (read-only: no signer, no tx, no wallet prompt), reused internally by `ensureContractAccountMapped`.

**E3 — dry-run/tx origin validation.** A non-SS58 origin (commonly the account's H160) used to fail deep in PAPI's `AccountId` codec as a bare `Invalid checksum`. Now validated up front with a typed `ContractInvalidOriginError` carrying an H160-aware hint; `.tx()`/`.prepare()` return `err(...)`, `.query()` throws (no err channel).

Implements proposal clusters **C1b** and **E3** (HANDOFF). Changesets included (contracts minor + umbrella minor).


Imod7 (approved):
LGTM! 
It would be great if we could also update the skills docs in `product-sdk/skills/product-sdk-contracts/`.

PR #266, opened 2026-07-21
fix(host): preserve TrUAPI routing IDs

pgherveou:
## Summary

- preserve the TrUAPI transport subscription ID through the PAPI ChainHead adapter
- buffer follow events emitted synchronously while the transport subscription starts
- preserve localhost, loopback, and subdomain.localhost product identifiers during signer connection
- add patch changesets and regression coverage for both fixes

## Why

The core registers ChainHead state under its opaque transport ID, such as p:17. Returning a synthetic follow_0 ID caused every PAPI follow-up request to address a subscription the core did not know.

The signer fallback also changed localhost:3000 into localhost:3000.dot, making a locally served product request an account belonging to a different product identifier.

## Testing

- pnpm --filter @parity/product-sdk-host test (87 passed)
- pnpm --filter @parity/product-sdk-signer test (98 passed)
- pnpm check
- local Dotli E2E against the Product SDK 0.19 playground: all 45 regular runnable cases passed; the additional sequential-allocation probe passed on retry after one remote-authority timeout


product-sdk/.changeset/preserve-local-product-id.md
```diff
@@ -0,0 +1,5 @@
+---
```
TarikGul, line 1: Changesets need to go in pending-changesets

product-sdk/.changeset/preserve-truapi-follow-id.md
```diff
@@ -0,0 +1,5 @@
+---
```
TarikGul, line 1: Same here needs to go in pending-changesets

TarikGul (approved):

TarikGul:
In this case its fine the changesets were'nt in pending since this commit is the only one since the last release

PR #267, opened 2026-07-21
ci(changesets): enforce pending staging

pgherveou:
## Summary

- reject release-ready changesets from non-release PRs
- allow dedicated `chore(release):` PRs to promote pending changesets
- document the enforced pending-to-release flow

## Why

PR #266 moved its changesets from `pending-changesets/` into `.changeset/` before merge. Because those release-ready files landed on `main`, the release workflow immediately published the patch as `@parity/product-sdk@0.19.1`; opening another release PR would have duplicated that release.

This follow-up turns the two review nits into a CI rule so feature PRs cannot bypass the dedicated release PR again.

## Testing

- `pnpm check`
- `pnpm changeset status --since=origin/main`
- parsed `.github/workflows/product-sdk-ci.yml` as YAML
- regression simulation flags the two `.changeset/` additions from #266
- current follow-up diff passes with no release-ready changesets

Refs #266

### DIFF
--- a/.github/workflows/product-sdk-ci.yml
+++ b/.github/workflows/product-sdk-ci.yml
@@ -34,7 +34,31 @@ jobs:
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
         with:
+          fetch-depth: 0
           persist-credentials: false
+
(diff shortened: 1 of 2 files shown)


TarikGul:
This is a good idea IMO I think we should reopen

PR #271, opened 2026-07-31
feat(host): bump truapi@0.6.0

valentinfernandez1:
## What

Bump `@parity/truapi` from v0.5.1 to [v0.6.0](https://github.com/paritytech/truapi/releases/tag/%40parity%2Ftruapi%400.6.0).

The release represents product-account derivation indexes as **tagged** `DerivationIndex` selectors (RFC 0022):

```typescript
// a plain numeric index, expanded by hosts to the internal 32-byte index
{ tag: "Left", value: number }
// a raw 32-byte index
{ tag: "Right", value: HexString }
```
It also adds the sr25519 account.signVrf API (RFC 0023).

## What changed
- Ergonomic surfaces keep plain numbers: `getProductAccount(id, index)` and `ProductAccount.derivationIndex` are unchanged, and the host adapter wraps them as Left at the wire boundary. 
- Signer now re-exports `DerivationIndex` and `ProductProofContext` from host instead of keeping structural copies.

### DIFF
--- a/product-sdk/packages/host/src/accounts.ts
+++ b/product-sdk/packages/host/src/accounts.ts
@@ -27,6 +27,7 @@ import { AccountId, type PolkadotSigner } from "polkadot-api";
 
 import type {
     ContextualAlias as WireAlias,
+    DerivationIndex,
     HostAccountConnectionStatusSubscribeItem,
     HostAccountCreateProofResponse as WireRingVRFProof,
     HostRequestLoginResponse,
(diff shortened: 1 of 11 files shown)


pgherveou (approved):

valentinfernandez1:
The e2e failures are expected: the test host (`@parity/host-api-test-sdk` 0.11.0) decodes the wire with pre-0.6.0 codecs, so any request carrying a derivation index (`createTransaction`, `signRaw`) misparses after the new tag byte and times out. Nothing in this PR can fix that; the SDK encoding comes from truapi's generated client.

PR #282, opened 2026-08-08
ci: enforce the pending-changesets convention and align the docs

Imod7:
### Description

Adds a CI check that fails a pull request when it puts a changeset file into
`product-sdk/.changeset/` instead of `product-sdk/pending-changesets/`, whether
directly or inside a subdirectory.

The check allows the one case where a file legitimately arrives in `.changeset/`:
a release PR promoting a changeset that was already parked in `pending-changesets/`.

### Changes

`.github/workflows/changeset-location.yml` (new file)

A single job that checks out the repository and compares the pull request against
its base commit, using a three-dot diff so that adding a changeset and moving it in
a later commit ends up green. It collects two kinds of offender:

- **Unpromoted changesets.** Any `.md` added directly under `product-sdk/.changeset/`,
  except `README.md`. `--no-renames` makes a legitimate promotion appear as an add
  under `.changeset/` plus a delete of the same basename under `pending-changesets/`,
  and those cancel each other out. Whatever add is left over was authored straight
  into `.changeset/`. Deletes are matched at any depth, so promoting a changeset
  parked in a `pending-changesets/` subdirectory still counts as a promotion.
- **Anything inside a `.changeset/` subdirectory**, of any extension, never cancelled
  by a promotion delete. There is no legitimate promotion into a subdirectory.

Failures produce one inline annotation per file in the Files changed tab, plus a
copy-paste block of remediation commands, deduplicated so several files sharing one
offending directory print a single `git rm -r`. Cause and fix differ by depth, so the
messages do too: `git mv` for a direct file, `git rm -r` for a directory.

The workflow deliberately has no `paths:` filter, unlike the other workflows in this
repository. A path filtered workflow reports no status on pull requests that do not
match its filter, and a required check that never reports leaves the pull request
waiting forever. This one always runs, so it can safely be marked required in the
branch ruleset.

`CLAUDE.md`, `product-sdk/RELEASES.md`, `product-sdk/pending-changesets/README.md`

All three said the pull request that closes the work promotes its own changeset. They
now say a separate `chore(release):` pull request does it, name the CI check, and warn
that the filename must not change during the move. No behaviour change, these are the
instructions people follow.

### Known limitation

A release pull request passes only if it purely promotes. Authoring a new changeset inside one fails and needs a separate pull request to park the file first. 

### Follow up, not in this PR

Someone with admin rights needs to mark `product-sdk: Changeset location` as required in Settings, Rules, Rulesets, on the `main` branch ruleset. Until that happens the check reports red but does not block merging. Copy the check name from a completed run rather than typing it, because a ruleset entry that matches no job leaves every pull request stuck on "Expected, waiting for status to be reported".

### DIFF
--- a/.github/workflows/changeset-location.yml
+++ b/.github/workflows/changeset-location.yml
@@ -1,9 +1,16 @@
 name: "product-sdk: Changeset location"
 
-# Guards the pending-changesets/ staging convention. Anything under
-# `.changeset/` is picked up by the release workflow on the next push to
-# `main`, so a changeset parked there by an in-flight PR gets published the
-# moment any *unrelated* PR merges. See product-sdk/pending-changesets/README.md.
+# Guards the pending-changesets/ staging convention.
+# See product-sdk/pending-changesets/README.md.
+#
+# A changeset directly under `.changeset/` ships on the next push to `main`, so
+# one parked there by an in-flight PR goes out when any unrelated PR merges.
+#
+# A subdirectory of `.changeset/` never ships (@changesets/read does a
+# non-recursive readdir) but is read as a legacy v1 changeset: `changeset
+# version` throws ENOENT on the missing `sub/changes.md`, breaking every
+# release until someone deletes it -- or publishes the directory unreviewed,
+# if `changes.md` and `changes.json` are both in it.
 #
 # Deliberately NOT path-filtered on `product-sdk/**` like the other workflows:
 # a filtered workflow reports no status on PRs that don't match, which leaves a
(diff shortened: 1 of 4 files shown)


.github/workflows/changeset-location.yml
```diff
@@ -0,0 +1,78 @@
...
+          # --no-renames makes a legitimate promotion show up as an add under
+          # .changeset/ plus a delete of the same basename under
+          # pending-changesets/. The awk pass cancels those against each other;
+          # whatever add is left over was authored straight into .changeset/.
+          #
+          # `sort` because array iteration order in awk is unspecified, and the
+          # runners use mawk, which has no asorti.
+          mapfile -t offenders < <(
+            git diff --no-renames --name-status "$BASE_SHA"...HEAD | awk '
+              $1=="A" && $2 ~ /^product-sdk\/\.changeset\/[^\/]+\.md$/ {
```
TarikGul, line 67: Should we check subdirectories as well?

TarikGul (approved):

.github/workflows/changeset-location.yml
Imod7: Good point! Fixed in [e3c9d33](https://github.com/paritytech/product-sdk/pull/282/changes/e3c9d33df100b3104c6243eae2f041569b1e703b)

Issue #286, opened 2026-08-10
Personhood and individuality support in product-sdk (tracking)


UtkarshBhardwaj007:
Two products are building on People chain personhood right now: humanity-spa and dim2-spa (Jollity). Both need the same things from the SDK, and both currently either hand-roll them or cannot express them at all from product code.

The Humanity team has proposed solving part of this with dedicated host calls instead: https://github.com/paritytech/humanity-spa/blob/develop/host-extensions.md. That approach would weld product domain logic (membership state, game phases, prize claims) into every host, so each rules change becomes a host release on four platforms. It is the same coupling problem RFC-0024 is currently removing for personhood keys. Everything in that document decomposes into a small set of SDK primitives plus typed domain helpers, which would work on any host and can be shared between Humanity and Jollity.

Goal of this tracker: everything listed in host-extensions.md should be expressible product-side through product-sdk.

- [x] #287 Personhood status helpers over the People chain
- [x] #288 Wrap account.signVrf (RFC-0023) in the accounts surface
- [x] #289 RFC-0024 ring VRF key management surface
- [x] #290 AsPerson transaction extension support in product-sdk-tx
- [x] #291 Shared game and prize domain helpers for Humanity and Jollity
- [x] #301 isDotNsAvailable() reports every name as available
- [x] #302 Account to username read over Resources.Consumers
- [x] #303 dotNS registry surface: default Asset Hub addresses have no code on chain
- [x] #304 Upgrade @parity/truapi to 0.9.x (unblocks #289)

Rough dependency order: 287 and 288 are standalone. 289 waits on the truapi JS client shipping the RFC-0024 wire methods. 290 needs proofs from the accounts surface but the extension building itself can start from the chain metadata. 291 builds on all of the above.

Context on host support: Android merged RFC-0024 natively (paritytech/polkadot-app-android-v2#969), the shared Rust core implementation is in review (paritytech/truapi#360), desktop has an implementation branch, and the JS client bindings do not exist yet. The manifest RFC (paritytech/truapi#206) is still open, so cross product proof allowlisting cannot be expressed yet. Details in the individual issues.

Imod7:
**Closing this.** Every sub-issue is closed, the work is published, and both consumers now track their own migration. The remainder is tracked where it can be fixed, which is not this issue.

[...]

PR #292, opened 2026-08-11
feat(host,signer): wrap account.signVrf in the accounts surface

### LINKED ISSUE #288: Wrap account.signVrf (RFC-0023) in the accounts surface
The hosts implement account.signVrf (sr25519 VRF over a caller supplied Merlin transcript, RFC-0023), and the host package changelog notes it is "not yet wrapped by an SDK accessor". Products that need it today have to go through the raw getTruApi() client.

Concrete consumer: the airdrop lottery ticket on the People chain. Game.sign_up_with_account takes one VRF per scheduled draw, and for players who are not yet recognized the ticket is an sr25519 VRF over the pop:airdrop transcript signed by the product account. dim2-spa currently implements this client side only for the standalone dev path where a mnemonic is in memory, and cannot do it in hosted mode without the raw client.

Ask: add signVrf to AccountsProvider / SignerManager next to createRingVRFProof, with a typed request (account, transcriptLabel, items) and response (preOutput, proof), and the same Result error handling as the other account calls. The fake truapi client in host/src/testing.ts already models the method, so mock support mostly exists.

Part of #286

Imod7:
Closes #288
Part of #286

### Description

Adds `signVrf` to the accounts surface so products can produce an sr25519 VRF signature
over a Merlin transcript without going through the raw truAPI client.

### Changes

| File | Change | Impact |
|---|---|---|
| `packages/host/src/accounts.ts` | `signVrf` on `AccountsProvider` and its adapter, plus the `VrfTranscriptItem`, `VrfSignature` and `ProductAccountLookup` types | New public method. Callers pass and receive `Uint8Array`; the adapter handles hex encoding and the tagged derivation index |
| `packages/host/src/index.ts` | Export the three new types | Types are usable by name |
| `packages/host/src/testing.ts` | Test driving the shipped fake through the real adapter | Catches a fake that stops matching what the adapter decodes. This fake is the only way a product can test a VRF flow |
| `packages/signer/src/providers/host.ts` | `HostProvider.signVrf`, the internal provider interface, and type re-exports from host | Same call reachable from the signer, with the signer error types |
| `packages/signer/src/signer-manager.ts` | `SignerManager.signVrf` | Reachable from the object most products already hold |
| `packages/signer/src/index.ts` | Export the three types | Types are usable by name |
| `pending-changesets/accounts-sign-vrf.md` | Changeset, minor for host, signer and the umbrella | Release notes and version bumps |

Two earlier commits also route `getProductAccount` and `getProductAccountSigner` through
one account-reference type, so the tagged derivation index is built in a single place.

### Why these changes

The hosts have implemented `account.signVrf` for a while, but no SDK accessor exposed it.
Products that needed it had to call `getTruApi()` directly and hand roll hex encoding,
account addressing and error handling. That also pins the product to the truAPI wire
format, which the SDK is supposed to absorb.

The concrete consumer is the airdrop lottery ticket on the People chain, which cannot be
produced in hosted mode today.

### Breaking change

`AccountsProvider` is an exported interface and `signVrf` is a required member, so
anything that implements it, including hand written test doubles, must add the method.
Code that only calls the provider is unaffected. The fake in
`@parity/product-sdk-host/testing` already implements it.

### Testing

```bash
cd product-sdk
pnpm install
pnpm -r build
pnpm -r --no-bail test -- --run
pnpm typecheck
pnpm check
```

Unit test results:
```
 pnpm --filter "@parity/product-sdk-host"   test -- --run
 pnpm --filter "@parity/product-sdk-signer" test -- --run
```

| Package | Before | After |
|---|---|---|
| `@parity/product-sdk-host` | 87 passed | 92 passed |
| `@parity/product-sdk-signer` | 107 passed | 113 passed |

All other packages are unchanged and green. `pnpm typecheck` and `pnpm check` are clean.

### DIFF
--- a/product-sdk/packages/host/src/accounts.ts
+++ b/product-sdk/packages/host/src/accounts.ts
@@ -6,7 +6,8 @@
  * `getAccountsProvider()` returns the full accounts surface — user identity
  * (`getUserId` / `requestLogin`), the user's existing wallet accounts
  * (`getLegacyAccounts`), app-scoped product accounts (`getProductAccount` /
- * `getProductAccountAlias`), Ring VRF proofs (`createRingVRFProof`), connection
+ * `getProductAccountAlias`), Ring VRF proofs (`createRingVRFProof`), sr25519 VRF
+ * signatures over a caller-supplied Merlin transcript (`signVrf`), connection
  * status, and PAPI `PolkadotSigner` factories for both product and legacy
  * accounts.
  *
@@ -40,9 +41,12 @@ import type {
     VersionedHostAccountCreateProofError,
     VersionedHostAccountGetAliasError,
     VersionedHostAccountGetError,
+    VersionedHostAccountSignVrfError,
     VersionedHostGetLegacyAccountsError,
     VersionedHostGetUserIdError,
     VersionedHostRequestLoginError,
+    VrfSignature as WireVrfSignature,
+    VrfTranscriptItem as WireVrfTranscriptItem,
     scale,
 } from "@parity/truapi";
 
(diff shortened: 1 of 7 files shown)


TarikGul (approved):
Looks good - lets get the e2e test fixed.

Imod7:
> Looks good - lets get the e2e test fixed.

Thanks! On the e2e failures, I looked into these and I also run locally and they are not coming from this PR. I will dig more in the e2e failures and I will open a separate PR for this.

Issue #291, opened 2026-08-10
Shared game and prize domain helpers for Humanity and Jollity


UtkarshBhardwaj007:
humanity-spa has drafted six host calls that would move game and prize domain logic into the hosts: getMembership, getCurrentGame, getDailyPrizeStatus, claimPrize, subscribeClaimPrize, registerGame. See https://github.com/paritytech/humanity-spa/blob/develop/host-extensions.md.

I think the same surface belongs in product-sdk instead. Every one of those calls decomposes into chain reads and existing signing primitives, host side implementations would have to be maintained on four platforms, and dim2-spa needs nearly identical logic (same pallets, same recognition checks, same airdrop claim flow), so a shared package serves both products.

Sketch, on top of the other issues in this tracker:

- current game and phase: Game.Game, Game.GameSchedules, phase and deadline selection (getCurrentGame)
- registration state and registration: account based and alias based paths (registerGame); the alias path needs the AsPerson extension issue
- daily prize status: event selection from the game's airdrops_scheduled, Airdrop event state, winner lookup, claim eligibility per the runtime rules (getDailyPrizeStatus). dim2-spa has a complete implementation of this read model already (packages/core/src/chain/airdrop.ts) which can seed the package
- claim and claim watching: build, submit and watch Game.claim_airdrop, resume after reload (claimPrize / subscribeClaimPrize)
- membership state: covered by the personhood status helpers issue
- no popups: the host-extensions doc requires no signing prompts for these flows, which is what an AutoSigning allowance already gives a product, so no new host mechanism is needed for that

Open question: package layout. One product-sdk-individuality package with people, game, score and airdrop modules probably beats several tiny ones.

Blocked by #287, #288 and #290 for the full surface. The read only parts (current game, prize status) have no blockers and can start immediately.

Part of #286

tommyldev:
There are a few runtime changes being worked on atm e.g decoupling prize eligibility from the game registration and instead tying it to personhood. I expect the pallet interface to change a bit. We should have a clearer picture by end of week

TarikGul:
cc: @Imod7

Imod7:
Closing this. Four of the six host calls shipped as asked, one shipped in a different shape, and
`registerGame` shipped for the account path only. What is left needs changes in `individuality`
and the hosts, not here.

#### What shipped

| Host call | Where | Shipped as |
|---|---|---|
[...]

Issue #289, opened 2026-08-10
RFC-0024 ring VRF key management surface


UtkarshBhardwaj007:
RFC-0024 (Proof of Personhood as a Product) makes ring VRF member keys explicit and product owned: products register their keys, list another product's keys by handle, and pass an explicit key handle to get_account_alias / create_account_proof, plus a new ring_vrf_sign. Spec: https://github.com/paritytech/truapi/blob/feature/people-rfc/docs/rfcs/0024-personhood-as-product.md (acceptance tracked in paritytech/truapi#324).

Host status: Android merged a native implementation (paritytech/polkadot-app-android-v2#969). The shared Rust core implementation is in review (paritytech/truapi#348, landing via paritytech/truapi#360). Desktop has an implementation branch. So hosts are arriving, but no SPA can call any of it: the JS client has no bindings for the new methods yet, and product-sdk has no surface.

Ask, once the truapi JS client ships the wire methods:

- wrap registerRingVrfKey, listRingVrfKeys (both disclosure levels) and ringVrfSign
- extend getProductAccountAlias / createRingVRFProof with the key handle parameter
- add a selection by ring helper: the RFC explicitly places "select the entry whose declared rings contain the requested ring, never surface a raw index" in the product SDK
- typed errors for KeyNotRegistered, KeyNotInRing, NotAllowlisted

Consumers lined up: humanity-spa (owns the peopl.dot keys under this model), dim2-spa (consumes them for score and prize flows), and a personhood faucet product for testing.

Note the cross product proof path also depends on the manifest RFC (paritytech/truapi#206) for the owner allowlist; reads and listing work without it.

Blocked by:

- a @parity/truapi release cut from current truapi main. The JS client is code-generated from the Rust API (truapi js codegen, src/generated is gitignored), so no hand-written bindings work is missing; but the latest npm release (0.8.0, published 2026-08-10) predates the RFC-0024 merge (paritytech/truapi#360, merged 2026-08-12), so no released client can encode the new calls. No release PR is open for the next version yet
- paritytech/truapi#206 (manifest RFC) for the cross product proof path only; registration, listing and own key use do not need it. See paritytech/truapi#373 for verified interim behavior: listing and get_account_alias work cross product, create_account_proof and ring_vrf_sign return NotAllowlisted with no prompt on core based hosts

[...]

Issue #287, opened 2026-08-10
Personhood status helpers over the People chain


UtkarshBhardwaj007:
Products keep re-deriving the user's personhood state from raw storage. humanity-spa needs the full membership card state (see the getMembership() call in https://github.com/paritytech/humanity-spa/blob/develop/host-extensions.md), and dim2-spa needs the same recognition checks for prize eligibility and its membership UI.

What is involved, all public storage on the People chain:

- Score.Participants (score, recognition, streak, attendance_history, reached_personhood), keyed by account or by score context alias
- Score.PersonhoodThreshold and Score.AbsenceGraceRatio for progress and the caution window
- Resources.Consumers(account).credibility for Lite vs full Person
- People.AccountToAlias / People.AccountToPersonalId for bound aliases
- the documented probe for "is this user a person": call createRingVRFProof or getProductAccountAlias and treat the typed NotMember error as the answer

Proposal: a typed read layer (new package, e.g. product-sdk-individuality, or a module in chain-client) that resolves a user (username, account, or product account) to one membership state along the lines of NotEnrolled / Lite / Candidate / MembershipReady / Member / Caution / Suspended, plus the raw values (score, threshold, misses, window) for UIs that need them.

There is also a footgun in the current SDK worth fixing or flagging as part of this: deriveContextAlias in sdk/src/identity/product-account.ts is a plain blake2b of parent pubkey and context, publicly linkable, not a ring VRF alias, and its RingLocation type does not match the protocol one. deriveAnonymousAlias / createRingProof / verifyRingProof in the same module just throw.

Part of #286

Imod7:
Done in #316, with #294 and #300 before it. In as asked: `Score.Participants` by account or alias,
`PersonhoodThreshold` and `AbsenceGraceRatio`, the seven-state union named as proposed, the raw
values alongside the state, and the `deriveContextAlias` footgun with its three stubs. Five asks
differ or sit elsewhere.

| Asked here | Delivered | Why |
|---|---|---|
| Lite from `Resources.Consumers().credibility` | `PeopleLite.LitePeople` presence | One lookup, and `credibility.tag` is not a standing check: a demoted person keeps `Person` and their full username. Credibility is surfaced from `lookupUsername` in #311 instead, with `lastUpdate` |
[...]

PR #293, opened 2026-08-11
feat(sdk): DotNS registry reads + writes in identity

### LINKED ISSUE #301: isDotNsAvailable() reports every name as available
The three dotNS resolution functions in @parity/product-sdk/identity are stubs, and one of them is dangerous rather than just unimplemented.

resolveDotNs() and reverseDotNs() throw "not yet implemented" (packages/sdk/src/identity/dotns.ts:112 and :139, both marked deprecated), which is fine: a caller finds out immediately. isDotNsAvailable() (dotns.ts:157) swallows that throw:

```ts
const record = await resolveDotNs(name).catch(() => null);
return record === null;
```

so it resolves to true for every input, including names that are taken. Verified against the published 0.21.0 dist and against live chain state: isDotNsAvailable("headlesszcvqzs.77.dot") returns true while Resources.UsernameOwnerOf has an owner for that name on paseo-people-next. A product wiring an availability check or a claim CTA on this silently offers taken names, and neither the type checker nor a smoke test catches it, because the signature and the happy path both look correct.

Runnable repro: dim2-spa apps/web/scripts/verify-uj6-people-names.mjs (deduction D7) demonstrates it against the live chain with the published package.

Ask: until real resolution lands, make isDotNsAvailable() throw like its two siblings, or return a typed error. A stub that lies is worse than a stub that throws. The tg-dotns-registry-impl branch implements the real reads; its default addresses have a separate issue.

Part of #286

TarikGul:
Closes #301
Part of #303
Part of #286

### Description

Implements the DotNS registry surface in `@parity/product-sdk/identity`, replacing the exported-but-throwing `resolveDotNs` / `reverseDotNs` / `isDotNsAvailable` stubs. Everything returns `Result<T, DotNsError>`; contract addresses default to the deployed set and are all overridable.

**Reads.** `resolveDotNs` hashes under the deployment's TLD, reads `registry.owner` and `registry.resolver`, then `resolver.addressOf` only when the pointer is the forward resolver. Three outcomes: `ok(null)` unregistered, `ok({ name, owner })` registered with no forward address record, `ok({ name, owner, address })` resolves (H160, not SS58). `isDotNsAvailable` asks `registrarController.available(label)` plus `PopRules.classifyName`, so a governance-reserved label is not reported as free. `reverseDotNs` calls `reverseResolver.nameOf`.

**Writes** return prepared calls the caller submits with their own signer. `setDotNsRecord` gives a `BatchableCall[]`: `setResolver` first when the pointer needs moving, then `setAddress`. `prepareDotNsRegistration` gives the commit call, secret, timing window, and a `prepareRegisterCall()` thunk for after `minCommitmentAge` (register consumes the commitment, so it cannot be built up front). Priced as `max(priceWithoutCheck(label, owner).price, transferFloor(label, payer, owner))`, matching what `register` charges.

### The TLD is per network, and read from the chain

**The part worth reviewing closely.** DotNS fixes its TLD when `DotnsProtocolRegistry.initialize` runs, with no setter: `.paseo` on Paseo Asset Hub Next V2, `.dot` on Previewnet, operator-chosen elsewhere. An earlier revision of this branch hardcoded `.dot`, so on the SDK's own default chain **every name read as unregistered** with no error anywhere, because the wrong root yields a node the chain has never written to.

The SDK now reads `protocolRegistry.tld()` once per runtime and derives the node itself, so one read answers the question and `tldNode()` is only a cross-check. A deployment older than `dotns` `b4096968` has no getter; there the TLD was a compile-time `.dot`, so an absent getter falls back to `.dot` with a warning. Any other failed read is an error, never a guess.

Contract addresses are the opposite: CREATE3-deterministic and identical on every network, verified on both. Hence `PASEO_ASSETHUB_DOTNS` becomes **`DOTNS_ADDRESSES`**, since the old name implied a per-network address table that does not exist.

### Behaviour notes for callers

- **A foreign suffix is refused** with reason `"TldMismatch"`, not hashed under ours. `alice.dot` and `alice.paseo` are separate registrations that may have different owners, and the identity layer cannot link a name across deployments in either direction, so translating silently could return a stranger's address. The label-to-alias binding is chain-local, and from individuality v0.12.0 the personhood context carries the network suffix, so the alias is per-network too.
- **A bare *multi-label* name is no longer suffixed.** `bob.alice` used to become `bob.alice.dot`; it is now refused, since telling it apart from `alice.dot` would need a hardcoded list of every network's TLD. A bare single label still works: `alice` gives `alice.paseo`.
- **Lite-person names must be passed flattened.** The contract strips the dots before hashing, so `alice.42` is stored as `alice42`. Resolve `alice42.paseo`.
- **`namehash(name, tld)` requires the root**, as do the three name helpers. A default correct on one deployment is the defect this branch fixes.
- **New failure mode:** `isDotNsAvailable` and `prepareDotNsRegistration` were never affected by the rooting bug, but now depend on the TLD read for validation, so a protocol-registry read failure fails a call that previously succeeded.
- **Breaking for anyone who imported the old stubs**, which took no options and threw. Pre-1.0, so `minor`. New exports: `DotNsTld`, `dotNsTld`, `DOT_TLD`, `stripSuffix`, `isConsistentDotNsTld`, and the `"TldMismatch"` / `"InvalidTld"` reasons.

### Verified against live chains

Unit tests could not have caught any of the three defects fixed here, so both deployments were exercised directly:

| | Next V2, `.paseo` | Previewnet, `.dot`, no getter |
|---|---|---|
| `resolveDotNs("dim2<own>")` | `{ name, owner: 0x4eD4A9b9... }` | `{ name, owner: 0x4eD4A9b9... }` |
| `resolveDotNs("dim2")` bare | same record | same record |
| `resolveDotNs("dim2<foreign>")` | `err TldMismatch` | `err TldMismatch` |
| `resolveDotNs("nosuchname9x")` | `ok(null)` | `ok(null)` |

The refusals point in opposite directions on the two chains, which is what distinguishes a per-network fix from relocated hardcoding.

142 unit tests, up from 83, each change also checked by reintroducing the defect and confirming the suite goes red (17 such checks, all caught). That mattered: the original bug survived three review passes because `dotns-namehash.test.ts` validated `namehash` against a reference implementation that reimplemented the same `.dot` rooting, so the tests could not have failed. The deterministic live values are pinned in tests (the `.paseo` tldNode, the `dim2.paseo` node, the content and PoP resolver addresses); owners are not, being mutable state.

**Not covered:** no write was submitted on chain; no name with a forward address record was found, so `ok({ name, owner, address })` is unit-test-only; `reverseDotNs` and lite-person names were not exercised live. There is no automated live test, so nothing here detects rot.

### Linked issues

```
Closes #301
Part of #303
Part of #286
```

- **`Closes #301`**, "isDotNsAvailable() reports every name as available". Fully resolved. Its ask was "until real resolution lands, make it throw or return a typed error", and this lands the real reads instead, and availability now asks `registrarController.available(label)` plus `PopRules.classifyName` rather than inferring from a zero resolver. That also covers the aside in #303, which flagged the same resolver-zero inference.
- **`Part of #303`**, not Closes. The issue is "the default Asset Hub contract addresses have no code on chain". The titled defect is gone: the defaults are now the CREATE3 set #303 itself names (`registry 0xf34054fd…`, `registrar 0x4f06E818…`), verified live on Next V2 and Previewnet. But neither of its two asks is implemented: no gateway walk at runtime, and no startup verification against `DotnsProtocolRegistry` that fails loudly on a mismatch. This PR does read one value from that contract at runtime (the TLD), which is the same direction, so it narrows #303 without answering it. Leaving it open is the tracking.
- **`Part of #286`**, the personhood and individuality tracking issue. This PR is the dotNS implementation behind two of its sub-issues, so the tracker can follow it from here.

Not linked, having checked: **#302** is a People-chain `Resources.Consumers` read, a different surface in the same file and untouched here; **#287**, **#289**, **#291** are personhood helpers, ring VRF keys and game helpers.

Worth noting that #303 independently corroborates the central finding of this PR before it was made: it records `owner(namehash("dim2.paseo"))` as owned on Next V2 and `owner(namehash("dim2.dot"))` as owned on Previewnet, which is the per-network TLD in one line, measured by someone else on different scripts.


product-sdk/packages/sdk/src/identity/dotns-registry.ts
```diff
@@ -0,0 +1,844 @@
...
+    const label = stripSuffix(normalized, tld.suffix);
+    const controllerAddr = opts.registrarControllerAddress ?? DOTNS_ADDRESSES.registrarController;
+    const popRulesAddr = opts.popRulesAddress ?? DOTNS_ADDRESSES.popRules;
+    const secret = `0x${bytesToHex(randomBytes(32))}` as HexString;
+    const registration = { label, owner: args.owner, secret, reserved: args.reserved ?? false };
+    // register compares msg.sender against the owner; msg.sender is derived
+    // from the SS58 origin, so derive it the same way to predict the branch.
+    const payer = ss58ToH160(origin);
+
+    try {
```
valentinfernandez1: ss58ToH160(origin) is above the try on the next line, so it throws rather than returning an err.

product-sdk/packages/sdk/src/identity/dotns-registry.ts
```diff
@@ -0,0 +1,844 @@
...
+     */
+    tld?: DotNsTld;
+}
+
+/** Arguments for {@link prepareDotNsRegistration}. */
+export interface RegisterDotNsArgs {
+    /** The name to register, e.g. `"alice.paseo"` — or the bare label `"alice"`. */
+    name: string;
+    /** The account that will own the registered name. */
+    owner: string;
```
valentinfernandez1, line 143: Nothing states what owner is. I assume H160

valentinfernandez1 (approved):
Just some small nits but over all LGTM

product-sdk/packages/sdk/src/identity/dotns-registry.ts
Imod7: Fixed in [628b0a0](https://github.com/paritytech/product-sdk/pull/293/changes/628b0a045197d9764caca37ed552f85d489ecb05)

product-sdk/packages/sdk/src/identity/dotns-registry.ts
Imod7: Fixed in [5ef3db0](https://github.com/paritytech/product-sdk/pull/293/changes/5ef3db0d992c50cf5302c3a483eaf3284e34ff61).

PR #294, opened 2026-08-11
refactor(sdk): deprecate the context-alias helpers and drop the ring stubs

Imod7:
Part of #286
Refs #287, which stays open for the personhood read layer.

### Description

`@parity/product-sdk/identity` exported five alias functions. Three raised an error on every
call, whatever you passed in. This PR deletes those three, so calling code now fails to build
instead of failing at runtime.

The other two work, but return an address that can receive funds and can never send them: the key
is a blake2b hash of the parent public key, and a hash has no matching private key, so nothing can
ever sign for it. Both are now deprecated, name the release that removes them (0.23.0), and list
what to use instead. Their output is unchanged, so no caller changes behaviour.

### Changes

| File | Change | Impact |
|---|---|---|
| `packages/sdk/src/identity/product-account.ts` | Deleted `deriveAnonymousAlias`, `createRingProof`, `verifyRingProof`. Deprecated `deriveContextAlias`, `verifyContextAlias` | Calls to the deleted three become compile errors. The other two are unchanged |
| `packages/sdk/src/identity/types.ts` | Deleted `AnonymousAliasInfo` and `RingLocation`. Deprecated `ContextAliasInfo` | Removes a `RingLocation` shaped `{ringIndex, memberIndex}`, the opposite of the protocol type `{chainId, junctions}` |
| `packages/sdk/src/identity/index.ts` | Dropped five exports, fixed the module docstring | Subpath no longer advertises Ring VRF helpers that do not exist |
| `packages/sdk/src/identity/product-account.test.ts` | New, 12 tests | Locks the derivation output so the deprecation cannot alter it |
| `packages/signer/src/providers/host.ts` | One doc comment | It claimed to match the product-sdk `RingLocation` shape. That was the opposite shape, and this PR deletes it |
| `pending-changesets/deprecate-context-alias.md` | New | `@parity/product-sdk` minor, `@parity/product-sdk-signer` patch |

DotNS is untouched. `resolveDotNs`, `reverseDotNs`, `resolvePeopleUsernameOwner` and the name
helpers work as before, and the subpath itself is not deprecated.

### Why these changes

Funds sent to a `deriveContextAlias` address are unrecoverable, and nothing surfaces until value
arrives, because the address validates like any other. The old docs made it worse by printing the
address in an example with no warning. This is the first time the SDK says so.

The three deleted functions could never work, so deleting them turns a guaranteed runtime error
into a compile error. A `@deprecated` tag adds nothing to a function that already throws.

The two working helpers are deprecated instead, because a caller may be using the value as a
plain identifier where the address property is irrelevant. Those callers are not broken and get a
release to move. For the same reason the output is left alone: the same function returning
different bytes would break them silently, with no compile error. Removal at a minor is in
policy, since `RELEASES.md` makes any pre-1.0 breaking change a minor.

The signer change follows from deleting identity's `RingLocation`, which a comment there pointed
at. It takes a patch because that comment ships in the published `.d.ts`, and that patch cascades
a version bump to five dependent packages containing no change, which is expected.

### Testing

From `product-sdk/`:

```bash
pnpm install
pnpm --filter "@parity/product-sdk-signer" test
```

| Suite | Before | After |
|---|---|---|
| `@parity/product-sdk` | 1 file, 6 tests passed | 2 files, 18 tests passed |
| `@parity/product-sdk-signer` | 7 files, 113 tests passed | 7 files, 113 tests passed |
| build, typecheck, check | pass | pass |

### Related issues

#287 is a sub-issue of the #286 tracker and this PR does half of it. The substantive reason is that the three deleted functions are stubs for what #289 builds for real: `deriveAnonymousAlias` was a placeholder for ring VRF aliases, and identity's `RingLocation` was a wrong-shaped duplicate of the protocol type #289 uses. Leaving them in place would mean two same-named types of opposite shape once that lands. Neither `Part of` nor `Refs` is a closing keyword, so nothing auto-closes here.

### DIFF
--- a/product-sdk/packages/sdk/src/identity/index.ts
+++ b/product-sdk/packages/sdk/src/identity/index.ts
@@ -3,8 +3,12 @@
 /**
  * @parity/product-sdk/identity
  *
- * Identity utilities: DotNS name resolution, context-alias derivation,
- * and Ring VRF anonymous aliases.
+ * Identity utilities: DotNS name resolution, plus the deprecated context-alias
+ * helpers.
+ *
+ * For real Ring VRF aliases and proofs, use `SignerManager`'s
+ * `getProductAccountAlias` and `createRingVRFProof` from
+ * `@parity/product-sdk-signer`.
  */
 
 // DotNS utilities
@@ -19,21 +23,13 @@ export {
 } from "./dotns.js";
 export type { PeopleUsernameChain, PeopleUsernameQueryApi } from "./dotns.js";
 
-// Context alias utilities
-export {
-    deriveContextAlias,
-    verifyContextAlias,
-    deriveAnonymousAlias,
-    createRingProof,
-    verifyRingProof,
-} from "./product-account.js";
+// Context alias utilities (deprecated)
+export { deriveContextAlias, verifyContextAlias } from "./product-account.js";
 
 // Types
 export type {
     DotNsRecord,
     ContextAliasInfo,
-    AnonymousAliasInfo,
-    RingLocation,
     VerificationResult,
     OnChainIdentity,
 } from "./types.js";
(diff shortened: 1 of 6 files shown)


TarikGul (approved):
Nice job!

PR #295, opened 2026-08-11
feat(host): export subscribeConnectionStatus

Imod7:
### Description

Adds a function to `@parity/product-sdk-host` that reports whether the app's channel to the host
container is up. You pass a callback, it runs immediately with the current status and again on every
change, and it returns a function to stop listening. The status is `"connecting"`, `"connected"` or
`"disconnected"`.

The signal already existed in `@parity/truapi`, reachable only by importing that lower level package
directly. This publishes it from the SDK.

```ts
import { subscribeConnectionStatus } from "@parity/product-sdk-host";

const unsubscribe = subscribeConnectionStatus((status) => setStatus(status));
```

### Changes

`packages/host/src/transport.ts` adds `subscribeConnectionStatus`, the `HostConnectionStatus` type,
and `emitConnectionStatus` for tests. It also holds `"disconnected"` when the underlying signal
reports `"connecting"` after the channel has closed, and notifies subscribers when `setTruApiClient`
injects or clears a client.

`packages/host/src/index.ts` exports the function and the type.

`packages/host/src/testing.ts` re-exports `emitConnectionStatus` and adds it to `FakeHost`, so a
product can test its offline and reconnecting states. Breaking for anyone who writes their own object
of type `FakeHost`, because the member is required.

`pending-changesets/host-subscribe-connection-status.md` is the changeset, minor for
`@parity/product-sdk-host` and `@parity/product-sdk`.

### Why these changes

coin-flip needs this to finish its migration and currently runs its own hook over
`@novasamatech/host-api-wrapper`. Landing it first means coin-flip never takes a direct dependency on
`@parity/truapi`, and the signature matches its existing hook, so the change there is one import.

The raw signal is wrong after a disconnect, so this corrects it rather than passing it through.
`@parity/truapi` 0.7.0 never clears its cached client when the pipe closes, so the next subscriber
reads `"connecting"` from a dead client and every existing subscriber receives that value with no way
back. A route change after the host went away showed a spinner that never ended instead of an offline
state. An upstream fix should follow, and this correction can be removed then.

### Testing

Before: 92 tests pass in the host package, 5 of them in `transport.ts`.
After: 102 pass, 15 in `transport.ts`.

Run from `product-sdk/`:

```bash
pnpm check
pnpm --filter "@parity/product-sdk-host" test
pnpm --filter "@parity/product-sdk-host" build
pnpm --filter "@parity/product-sdk" build
pnpm typecheck
```


product-sdk/packages/host/src/index.ts
```diff
@@ -53,6 +53,11 @@ export type {
     RemotePermission,
 } from "./truapi.js";
 
+// Host-channel connection status — the transport's own signal. Distinct from
+// @parity/product-sdk-signer's ConnectionStatus, which tracks a signer provider.
+export { subscribeConnectionStatus } from "./transport.js";
```
TarikGul, line 62: Conflicts here, just need a `git pull origin main`

product-sdk/packages/host/src/transport.ts
```diff
@@ -85,6 +100,100 @@ export async function getClient(): Promise<TrUApiClient | null> {
...
+ * for the host, `"connected"` once the channel is established, `"disconnected"`
+ * outside a host container or after the channel closes.
+ *
+ * Not the same concept as `@parity/product-sdk-signer`'s identically-shaped
+ * `ConnectionStatus`, which tracks a signer provider rather than the transport.
+ */
+export type HostConnectionStatus = ConnectionStatus;
+
+/**
+ * Correct one defect in the sandbox's status signal: `@parity/truapi` 0.7.0 never
```
TarikGul: But we are on 0.9.0, so did the truapi not fix this yet?

TarikGul (approved):
Good to go, just some comments and conflicts need to be fixed.

product-sdk/packages/host/src/index.ts
Imod7: Yes, thanks! Merged main & resolved conflicts in [`015cebf`](https://github.com/paritytech/product-sdk/commit/015cebf50b33faa4a1b561ed11be51775b084c68)

product-sdk/packages/host/src/transport.ts
Imod7: No it has not been fixed. Updated the comments and text in [`b9a783d`](https://github.com/paritytech/product-sdk/commit/b9a783df974afb219d7f5c144a703d754989533b) to cover the newer truapi versions that also have this defect and I will open a follow-up PR in host-rust-core to fix this.

PR #298, opened 2026-08-14
feat: consume chain.getChainInfo discovery, bump truapi@0.9.0

valentinfernandez1:
## What

`@parity/product-sdk-host` gains `getHostChainInfo()`, a cached facade over truapi's `chain.getChainInfo` that resolves chain roles (`AssetHub`, `Bulletin`, `People`) to genesis hashes from the host's own configuration instead of the SDK hard-coding them. `getChainAPI()` can now be called with **no argument** inside a discovery-capable container, deriving the environment by matching the discovered asset hub genesis against the bundled descriptors. Calls that pass an environment behave exactly as before on hosts predating discovery; the zero-arg form throws there, and outside a container, asking for an explicit environment.

Cross-validated against dotli.dev, which implements the host side of the `supported-chain` genesis query, by linking this branch into `host-playground` locally and adding a method there that calls `chain.getChainInfo`.


product-sdk/packages/chain-client/src/presets.ts
```diff
@@ -126,6 +130,85 @@ type PresetDescriptors = {
...
+function validateDescriptorGenesis(
+    descriptors: Record<keyof typeof CHAIN_IDENTIFIERS, ChainDefinition>,
+    discovery: HostChainDiscovery,
+): void {
+    for (const [key, identifier] of Object.entries(CHAIN_IDENTIFIERS)) {
+        const hostGenesis = discovery.chains[identifier];
+        const genesis = descriptors[key as keyof typeof CHAIN_IDENTIFIERS]?.genesis;
+        if (!hostGenesis || !genesis) continue;
+        if (genesis.toLowerCase() !== hostGenesis.toLowerCase()) {
+            throw new GenesisMismatchError(key, genesis, hostGenesis);
```
Imod7: This throws on the first chain that disagrees, so one stale descriptor makes `getChainAPI` fail entirely and the chains that do match become unreachable. That undoes `unsupportedChainApi` in `clients.ts`, which exists so an unservable chain does not tank the whole client, and which has a test pinning it. Tested: descriptors 0.5.1 records paseo-individuality being re-genesised on its own with the asset hub untouched, and reproducing that here, both call forms reject and `createChainClient` is never reached. Please throw only when `key === "assetHub"`, the chain the environment was derived from, and warn for the others; I tried it and the 13 existing preset tests stay green.

product-sdk/packages/host/src/accounts.ts
```diff
@@ -194,38 +186,48 @@ export interface AccountsProvider {
...
-     *   is not VRF-scoped, so it covers other signing by that account too.
-     *
-     * Hosts predating the call reject it through the error channel.
+     * Register a ring-VRF key owned by the calling product at `index` within
+     * its ring-VRF domain, declared for `ring`. Returns the member's public
+     * key. The registered key is then addressed by a `keyHandle` of
+     * `{ dotNsIdentifier: <calling product>, derivationIndex: index }`.
+     */
+    registerRingVrfKey(
+        index: DerivationIndex,
```
Imod7: These take raw wire shapes, against the convention stated at the top of this file, in the comment starting "The account/alias shapes come from `@parity/truapi`'s generated specs": bytes decoded to `Uint8Array`, derivation index kept as a plain `number` and wrapped into `Index` at the boundary. `ProductAccount` and `toWireProductAccountId` both follow it. `main` already answered this too: `signVrf` takes `account: ProductAccountLookup`, which a `ProductAccount` satisfies, and wraps it with `toWireProductAccountId`. As written a caller holding a `ProductAccount` cannot pass it as a `keyHandle`. Please take `number` here and `ProductAccountLookup` for `keyHandle` on the three methods that take one, and align the results too: `registerRingVrfKey` returns hex while `ringVrfSign` right below decodes with `fromHex`.

product-sdk/packages/host/src/testing.ts
```diff
@@ -193,6 +193,9 @@ export function createFakeTruApiClient(options?: CreateFakeTruApiClientOptions):
                     ringIndex: 0,
                     ringRevision: 0,
                 }),
+            registerRingVrfKey: () => okAsync(publicKey),
```
Imod7, line 230: Thanks for adding the three new account methods to the fake. The chain domain needs the same: `getChainAPI` now calls `chain.getChainInfo` on every invocation and the fake has `chain: notModeled("chain")`, so it throws. I ran it, `getHostChainInfo` resolves `null` and logs "not modeled by the fake", which means `getChainAPI()` with no argument always fails in consumer tests and `getChainAPI("paseo")` logs that warning on every call. Please model `chain.getChainInfo`, ideally with an option for which chains the fake host serves.

product-sdk/packages/host/src/chain-discovery.ts
```diff
@@ -0,0 +1,279 @@
...
+                ),
+            ),
+        );
+        const outcomes = await Promise.race([
+            probe,
+            new Promise<"timeout">((resolve) => {
+                timer = setTimeout(() => resolve("timeout"), PROBE_TIMEOUT_MS);
+            }),
+        ]).finally(() => clearTimeout(timer));
+        if (outcomes === "timeout") {
```
Imod7, line 123: Returning `null` here caches the timeout as a stable answer, so a host that supports discovery but started slowly is recorded as pre-discovery for the life of the client. I tested it with a host answering at 3.5 seconds: the second call still resolves `null` and `getChainInfo` ran once, so `getChainAPI()` throws "pass an explicit environment" for the rest of the session. Please return `TRANSIENT_FAILURE` instead, so the existing eviction path re-probes. The trade-off is that a genuinely legacy host then pays the deadline per call, which is bounded since `createChainClient` caches by fingerprint.

product-sdk/packages/chain-client/src/presets.ts
```diff
@@ -126,6 +130,85 @@ type PresetDescriptors = {
...
+            );
+        }
+        return env;
+    }
+    const discovered = discovery.chains.AssetHub?.toLowerCase();
+    let matched: Environment | null = null;
+    if (discovered) {
+        const candidates = env ? [env, ...PROBE_ORDER.filter((e) => e !== env)] : PROBE_ORDER;
+        for (const candidate of candidates) {
+            const genesis = await assetHubGenesis[candidate]();
```
Imod7: There is no error handling around this, so a descriptor chunk that fails to load takes down the whole call, including for environments that have nothing to do with it. `PROBE_ORDER` starts with paseo, so a devnet host running `getChainAPI()` loads the paseo bundle first. Please `catch` and `continue` here, with a warning, so a broken bundle is just a candidate that does not match.

Imod7 (commented):
Flagging that `product-sdk/pending-changesets/deprecate-context-alias.md` has a "Use instead" row pointing at `SignerManager.getProductAccountAlias(context, location)` and says the ring VRF operations are "host-backed so the host selects the member key within the ring". This PR falsifies both. Since #294's code is already on `main`, any release cut after this merges publishes that advice against a signature it does not contain, whichever wave promotes the changeset. Only those two clauses need touching; the rest of the file is historical record and should stay frozen.

peetzweg:
Looking forward for this getting in and published!

valentinfernandez1:
Will hold this until #299 lands on main

Imod7 (approved):

PR #299, opened 2026-08-16
feat(host): add registered ring VRF key selection

### LINKED ISSUE #304: Upgrade @parity/truapi to 0.9.x (first release with the RFC-0024 wire shapes)
@parity/truapi 0.9.0 was published on 2026-08-13, the first release cut after the RFC-0024 core merge (paritytech/truapi#360, merged 2026-08-12). It carries the new wire shapes: key_handle is a required field on get_account_alias and create_account_proof, the old host-selects-the-key shape is gone, and registerRingVrfKey / listRingVrfKeys / ringVrfSign exist. This is the release cut #289 has been blocked on.

The workspace catalog still pins "@parity/truapi": ^0.7.0 (pnpm-workspace.yaml), so every package resolves a client that mis-encodes alias and proof calls against any host built from current truapi main (and against Android since paritytech/polkadot-app-android-v2#969).

Ask:

- bump the catalog to ^0.9.0
- adapt the two shipped wrappers (getProductAccountAlias, createRingVRFProof) to the required key_handle shape, since the old encoding does not exist anymore
- then #289 proper can add the new RFC-0024 surface on top

truapi main has moved again since the cut (allowance slot pooling across personhood collections, typed login failures), so expect another release soon; ^0.9.0 keeps us on the current wire either way.

Part of #286

replghost:
Closes #304
Part of #289, #286

## Summary
- Update `@parity/truapi` to 0.9 and migrate derivation selectors to `Index` / `Raw`.
- Expose registered ring-VRF key listing plus ring-based opaque-handle selection.
- Require explicit key handles for contextual alias and ring-proof requests across host, signer, and manager APIs.

## Verification
- `pnpm build`
- `pnpm typecheck`
- `pnpm test`
- `pnpm check`
- `pnpm format:check`


product-sdk/packages/host/src/accounts.ts
```diff
@@ -180,11 +244,17 @@ export interface AccountsProvider {
         dotNsIdentifier: string,
         derivationIndex?: number,
     ): ResultAsync<ProductAccount, scale.CallErrorValue<VersionedHostAccountGetError>>;
-    /**
-     * Derive the contextual alias for a proof context and ring. The host
-     * selects the member key within the ring — no per-account addressing.
-     */
+    /** List an owner's registered ring-VRF keys. */
+    listRingVrfKeys(
```
Imod7, line 264: TrUAPI 0.9 adds three account methods, `registerRingVrfKey`, `listRingVrfKeys` and `ringVrfSign`, and this wraps one of them. A product that has never registered a key gets `[]` here, so the alias and proof calls cannot be made at all, where before this change the host picked a key and they worked. Could you either wrap registration too, or note in this doc comment and in the changeset that it is not wrapped yet and `getTruApi().account.registerRingVrfKey` is the interim path.

product-sdk/pending-changesets/truapi-09-ring-vrf-keys.md
```diff
@@ -0,0 +1,20 @@
...
+
+**Update TrUAPI to 0.9 and require registered ring-VRF key handles.**
+
+`AccountsProvider`, `HostProvider`, and `SignerManager` now expose
+`listRingVrfKeys(owner, disclosure?)`. The returned `RegisteredRingVrfKey` entries carry opaque
+`RingVrfKeyHandle` values. `findRingVrfKeyHandle(keys, ring)` selects a handle by declared
+`RingLocation`, so products do not hard-code another product's derivation index.
+
+`getProductAccountAlias` and `createRingVRFProof` now require that handle as their first argument.
+This is a compile-time breaking change. It matches TrUAPI 0.9, where the host no longer chooses a
```
Imod7, line 16: There is a second compile-time break this does not mention. The signer package's local `RingLocation` declared `chainId` as a plain `string` so callers could pass plain strings, and it is now a re-export of the host package's, so it is `` `0x${string}` ``. I reproduced it: a `const chainId: string` assigned into a `RingLocation` now fails with `TS2322`. Inline literals still infer fine, so it only bites consumers whose chain id comes from configuration. One sentence covers it, plus one for the internal `AccountsProvider` gaining a required `listRingVrfKeys`, which affects anyone passing their own `loadAccountsProvider`.

Imod7 (approved):

PR #300, opened 2026-08-17
feat(individuality): add the personhood read layer

Imod7:
### Description

Adds `@parity/product-sdk-individuality`, a new published package that reads a person's personhood state on the individuality chain from a DotNS username, at one pinned finalized block. Ports the derivation and the read from humanity-spa.

Part of #286. Refs #287, does not close it. The eligibility half is held behind #291, and #291 is not the only gap: five items from the issue body sit outside this PR, mapped item by item in the issue analysis.

Two of those are deliberate divergences from the issue text rather than omissions, worth stating so a reviewer diffing the PR against the issue does not read them as oversights:

- **The Lite signal comes from `PeopleLite.LitePeople` presence, not `Resources.Consumers(account).credibility`.** The issue names the latter. Both exist in the descriptor, so this is a choice: it follows the reference implementation in humanity-spa, which is the code this ports and the only version verified against the runtime. The alias is read from both pallets' `AccountToAlias` for the same reason.
- **A user is resolved by DotNS username only**, not by account or product account. `People.AccountToPersonalId` is also not read. Both are additive and neither blocks the seven-state answer.

### Changes

New package, `product-sdk/packages/individuality/`:

- `src/types.ts`: the seven-state `PersonhoodState` union (`NotEnrolled`, `Lite`, `Candidate`, `MembershipReady`, `Member`, `Caution`, `Suspended`), wrapped by `UsernameUnowned | Resolved`, plus `FinalizedSnapshot`, `AbsenceGracePolicy`, `PersonhoodParticipant` and `PersonhoodInputs`. Imports nothing, so both other modules can sit on it.
- `src/derive.ts`: the pure state machine, no I/O and no chain types. 15 tests.
- `src/decode.ts`: raw PAPI values to domain shapes. Rejects unknown `streak` and `recognition` variants, and enforces the runtime's grace invariants (`window <= 8`, `allowedMisses < window`) so a wrong byte order fails loudly instead of making `Caution` unreachable. 19 tests.
- `src/read.ts`: the batch. Pins one finalized block, resolves `Resources.UsernameOwnerOf`, runs six parallel reads, then a conditional alias-keyed `Score.Participants` lookup. Returns a `Result`. 18 tests.
- `src/errors.ts`: `ProductIndividualityError` and `IndividualityDecodeError`, both implementing the `SdkError` marker. 3 tests.
- `src/index.ts`: the public surface, exporting the derivation separately from the read.
- `package.json` at `0.0.0` so the changeset publishes `0.1.0`. Depends on `@parity/product-sdk-errors`, `@parity/result` and `polkadot-api`.
- `tsconfig.json`, `tsconfig.typecheck.json`, `tsup.config.ts`, `vitest.config.ts`: copied from `packages/statement-store`.

Umbrella wiring, mirroring `cloud-storage`: dependency and `./individuality` export in `packages/sdk/package.json`, the build entry in `packages/sdk/tsup.config.ts`, and `packages/sdk/src/individuality/index.ts`.

Also in `packages/sdk`: `src/individuality/contract.test.ts`, type-level assertions that a real `getChainAPI` client for both `paseo` and `devnet` still satisfies the read's structural chain type, plus a negative control that fails if that type ever stops constraining. The assertions gate CI through `pnpm typecheck`, and the one runtime test only stops vitest reporting an empty suite. It lives in the umbrella package rather than beside the type because the same assertions are vacuous inside the individuality package, see Risks.

Docs and release:

- `skills/product-sdk-individuality/SKILL.md`: quick start, the state table, the traps.
- `README.md`, `product-sdk/README.md`, `CLAUDE.md`: register the package and skill in the three hand-maintained lists.
- `pending-changesets/individuality-read-layer.md`: minor for the package and the umbrella.

### Why these changes

humanity-spa already had this logic, verified against the runtime, so anything else needing the same answer would have had to copy it. This lifts it into the SDK, keeps the derivation importable on its own for #291, and takes an already connected chain client so the caller picks the chain. That last choice matters beyond tidiness: the original plan resolved the chain itself, and with `chain.getChainInfo` unreachable at the pinned truapi version it called for a static chain table behind a seam. An injected client needs neither, which is why #298 landing requires no change here.

### Testing

```
cd product-sdk
pnpm install
pnpm check
pnpm build
pnpm typecheck
pnpm -r --no-bail test
```

Before: the package did not exist on `main`, so nothing covered this. The source vectors lived in humanity-spa, 12 for the derivation and 18 for decode and assembly.

After: 55 tests pass in `packages/individuality` (15 derive, 19 decode, 18 read, 3 errors) and 19 in `packages/sdk`, up from 18 with the contract assertion. `pnpm check` clean over 296 files. `pnpm build` and `pnpm typecheck` clean.

### Risks

- **`Score.PersonhoodThreshold` is a `u8`, and nothing can guard that.** PAPI types `u8` and `u32` alike as `number`, so the wrong width typechecks and passes tests. Checked against the committed metadata, and commented at the point of use.
- **`Score.AbsenceGraceRatio` byte order is `(allowed_misses, window)`.** The metadata tuple is anonymous, so the order comes from the pallet doc comment. This one is guarded: the decode enforces `allowedMisses < window`, so a swapped order throws instead of silently making `Caution` unreachable.
- **`Resources.UsernameOwnerOf` is read in two places**, here and at `resolvePeopleUsernameOwner` in `packages/sdk/src/identity/dotns.ts`. Deliberate: that one cannot pin a block and returns hex rather than the `SS58String` the next reads need.
- **The chain parameter is structural, not a descriptor.** Its compile-time guard sits in `packages/sdk` because the same assertion is vacuous inside the individuality package, where the descriptor types do not fully resolve. Both halves were tested before choosing the placement.
- **Not an authorization oracle.** A client-side read in a client-side library. A backend trusting "the SDK said Member" is trivially spoofed.
- **Chain targeting is unresolved.** The package reads whichever chain the caller's `getChainAPI(env)` selected, which is the right layering but means this PR does not pick one.

### DIFF
--- a/product-sdk/packages/individuality/src/decode.ts
+++ b/product-sdk/packages/individuality/src/decode.ts
@@ -13,7 +13,7 @@
  * 2. **`Score.PersonhoodThreshold` is a `u8`.** PAPI maps both `u8` and `u32` to
  *    `number`, so a width mistake typechecks *and* passes tests. It is read
  *    straight through rather than decoded here, so the note lives on
- *    `PersonhoodSnapshot.personhoodThreshold` in `derive.ts` and at the read
+ *    `PersonhoodInputs.personhoodThreshold` in `types.ts` and at the read
  *    site — but it belongs to the same class of trap as the one above.
  *
  * Unknown enum variants throw {@link IndividualityDecodeError} rather than
@@ -21,8 +21,7 @@
  * variant added by a runtime upgrade must fail loudly.
  */
 import { IndividualityDecodeError } from "./errors.js";
-import type { PersonhoodParticipant } from "./derive.js";
-import type { AbsenceGracePolicy } from "./types.js";
+import type { AbsenceGracePolicy, PersonhoodParticipant } from "./types.js";
 
 /** The attendance history is one byte, so the runtime caps the grace window at 8. */
 const GRACE_WINDOW_MAX = 8;
--- a/product-sdk/packages/individuality/src/decode.ts
+++ b/product-sdk/packages/individuality/src/decode.ts
@@ -24,6 +24,9 @@ import { IndividualityDecodeError } from "./errors.js";
 import type { PersonhoodParticipant } from "./derive.js";
 import type { AbsenceGracePolicy } from "./types.js";
 
+/** The attendance history is one byte, so the runtime caps the grace window at 8. */
+const GRACE_WINDOW_MAX = 8;
+
 /** `Score.AbsenceGraceRatio` serialized as `SizedHex<2>`: `0x` + four hex digits. */
 const GRACE_RATIO_PATTERN = /^0x[0-9a-fA-F]{4}$/;
 
@@ -40,10 +43,27 @@ export function decodeAbsenceGracePolicy(value: string): AbsenceGracePolicy {
     }
     // Byte order is (allowed_misses, window). The metadata tuple is anonymous,
     // so the order comes from the pallet doc comment, not the type.
-    return {
-        allowedMisses: Number.parseInt(value.slice(2, 4), 16),
-        window: Number.parseInt(value.slice(4, 6), 16),
-    };
+    const allowedMisses = Number.parseInt(value.slice(2, 4), 16);
+    const window = Number.parseInt(value.slice(4, 6), 16);
+
+    // The runtime guarantees `window <= 8` and `allowed_misses < window` for every
+    // grace tier, so both checks below reject impossible chain data rather than
+    // trusting it. They exist because the byte order above is unverifiable from
+    // the type: were the halves ever swapped, the decode would still succeed and
+    // return plausible numbers, and since the projection can return at most
+    // `window`, `misses > allowedMisses` would become unreachable and `Caution`
+    // would silently never fire. That turns the one trap this file cannot see
+    // into a loud failure.
+    if (window > GRACE_WINDOW_MAX) {
+        throw new IndividualityDecodeError("absence grace window exceeds the attendance history");
+    }
+    // `window === 0` is the legitimate no-grace tier and carries no useful
+    // allowance, so it is exempt from the ordering rule.
+    if (window !== 0 && allowedMisses >= window) {
+        throw new IndividualityDecodeError("absence grace policy violates allowed_misses < window");
+    }
+
+    return { allowedMisses, window };
 }
 
 /** The raw `streak` enum as PAPI decodes it: `Enum<{ Attended: u32; Absent: u32 }>`. */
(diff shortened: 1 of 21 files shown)


product-sdk/packages/individuality/src/derive.ts
```diff
@@ -0,0 +1,350 @@
...
+            // stays a plain member even when the personhood flag is unset. Testing
+            // the flag before the recognition variant gets this wrong.
+            return { tag: "Member", activeWeeks, lastAttendedGame: participant.lastAttendedGame };
+
+        case "Suspended":
+            return { tag: "Suspended" };
+
+        case "Recognized": {
+            if (!participant.reachedPersonhood) {
+                // Fail-safe: recognized without personhood is inconsistent state.
```
TarikGul, line 75: Can we add a default here that returns { tag: "Suspended" }, or an assertNever? This module is the "pure, never-throws" thing and #291 is going to call derivePersonhoodState directly with inputs that didn't come through our decoder — so an unexpected recognition value would fall through and return undefined, which isn't a valid state. Cheap to close and it future-proofs us if a new recognition variant lands.

product-sdk/packages/individuality/src/read.ts
```diff
@@ -0,0 +1,621 @@
...
+    };
+}
+
+if (import.meta.vitest) {
+    const { describe, expect, test } = import.meta.vitest;
+    const { unwrapOk, unwrapErr, isErrorOf } = await import("@parity/result");
+    const { IndividualityDecodeError } = await import("./errors.js");
+
+    const ALICE = "5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY";
+    const ALIAS = `0x${"ab".repeat(32)}`;
```
TarikGul, line 267: Our RPC-failure test only rejects getFinalizedBlock. Could we add one where one of the six parallel batch reads rejects, and assert it comes back as a clean typed error? The behaviour's correct by construction with the outer try/catch, but the actual parallel-read-fails path isn't covered and it's the kind of thing I'd want pinned.

product-sdk/packages/individuality/src/read.ts
```diff
@@ -0,0 +1,621 @@
...
+    }
+
+    const [
+        accountParticipant,
+        peopleAlias,
+        liteAlias,
+        litePerson,
+        personhoodThreshold,
+        absenceGraceRatio,
+    ] = await Promise.all([
```
TarikGul, line 216: Our RPC-failure test only rejects getFinalizedBlock. Could we add one where one of the six parallel batch reads rejects, and assert it comes back as a clean typed error? The behaviour's correct by construction with the outer try/catch, but the actual parallel-read-fails path isn't covered and it's the kind of thing I'd want pinned.

product-sdk/packages/individuality/src/decode.ts
```diff
@@ -0,0 +1,284 @@
...
+}
+
+/**
+ * Map a raw PAPI participant to the domain shape the derivation consumes.
+ *
+ * The recognition payload (a revision id) is discarded, the game-economy fields
+ * are dropped, and a missing `last_attended_game` becomes `null` rather than
+ * `undefined` so the domain type has one absent value, not two.
+ */
+export function toPersonhoodParticipant(raw: RawParticipant): PersonhoodParticipant {
```
TarikGul, line 142: Small thing, we validate the enum variants but pass score / streak.value / attendance_history straight through. Can we drop a one-line comment noting that's deliberate (enums validated, numerics trusted-by-descriptor)? Just so the asymmetry reads as a choice and not an oversight. No behaviour change needed.

TarikGul (approved):
All small nits, you can decide if they are important enough.
 Nice job

product-sdk/packages/individuality/src/derive.ts
Imod7: Fixed in [0983d0f](https://github.com/paritytech/product-sdk/pull/300/changes/0983d0f1bc1de414417cbde3009a3602e14b16da).

product-sdk/packages/individuality/src/read.ts
Imod7: Fixed in [1ee90c4](https://github.com/paritytech/product-sdk/pull/300/changes/1ee90c49c36340e90fcfcda1446257e3c138853f).

product-sdk/packages/individuality/src/decode.ts
Imod7: Added in [64afc7d](https://github.com/paritytech/product-sdk/pull/300/changes/64afc7d7bf2a85f169efeb7134b529cc3daad41c).

PR #309, opened 2026-08-19
feat(individuality): support the AsPerson transaction extension

### LINKED ISSUE #290: AsPerson transaction extension support in product-sdk-tx
On the People chain, acting as a person is not a normal signed origin. The transaction carries the AsPerson transaction extension with one of the AsPersonInfo variants: AsPersonalAliasWithProof (fresh ring VRF proof plus context), AsPersonalAliasWithAccount (plain signature by a previously bound alias account), AsPersonalIdentityWithProof / WithAccount, and AsPersonalAliasWithAccountRevised for ring revision changes.

Today only native host code can build these (the mobile apps construct them internally, for example when binding the score alias). A product can get the proof itself just fine via createRingVRFProof, but there is no way to put that proof into a transaction with product-sdk-tx, so flows like these are impossible from product code:

- People.set_alias_account, binding a contextual alias to a product account so later transactions can act as the person with a plain signature
- Game.sign_up_with_alias for recognized players
- any call that should run under a person origin instead of an account origin

Ask: support building and submitting transactions that carry the AsPerson extension, taking the RingVRFProof returned by the accounts surface as input, covering at least AsPersonalAliasWithProof and AsPersonalAliasWithAccount, with the message binding rule handled correctly (the proof signs a hash of the call implication, so build order matters).

This is the main missing piece between "the SDK can prove personhood" and "a product can act on it on chain".

No external blockers for the extension building itself: the AsPerson extension and its variants are in the live chain metadata, and the proof input already comes from createRingVRFProof. The key handle variant of the proof call (#289) can slot in later without changing the transaction side.

Part of #286

Imod7:
Closes #290

### Description

Adds `withAsPerson`, which wraps a `PolkadotSigner` so a call dispatches under a person origin on the People chain instead of an account origin.

### Changes

`packages/individuality/src/as-person-codec.ts` new. Reads the extension pipeline from the chain's own metadata and encodes `AsPersonInfo`. Encoding from metadata rather than a hand-written type matters because the deployed variant and the upstream `polkadot-sdk` one share a name with different field lists, so an upstream-derived encoder would emit plausible bytes with a field missing.

`packages/individuality/src/as-person-implication.ts` new. Builds the call implication and its
blake2-256 message, plus the extra binding the revised-alias variant needs.

`packages/individuality/src/as-person-signer.ts` new. `withAsPerson`. Sets `RestrictOrigins` to `true`, supplies `VerifyMultiSignature` as `Disabled` for the proof variant, hashes the implication, requests the proof, then writes `AsPerson` last.

`packages/individuality/src/errors.ts` adds `AsPersonError`. Thrown, not returned, because it happens inside `signTx` where there is no `Result` channel.

`packages/individuality/src/index.ts` exports `withAsPerson`, `AsPersonInfo`, `CreateRingVRFProof`, `RingVRFProof`, `AsPersonError`. The metadata-driven internals stay internal.

`packages/sdk/src/individuality/contract.test.ts` adds compile-time assertions tying the local `RingVRFProof` to the host's and pinning `People.set_alias_account`, each with a negative control.

`pnpm-workspace.yaml`, `packages/individuality/package.json`, `pnpm-lock.yaml` add `@polkadot-api/metadata-builders` at the version already in the tree.

`skills/product-sdk-individuality/SKILL.md`, `skills/product-sdk-transactions/SKILL.md` document the write path. `pending-changesets/as-person-extension.md` is the changeset, minor for the package and for the umbrella.

`packages/tx` is not touched.

### Why these changes

Only native host code could build this extension, so a product could obtain a ring VRF proof but had
no way to put it in a transaction. That blocked `Game.sign_up_with_alias` and any call needing a person origin.

Two things this handles that a hand-rolled extension does not. `RestrictOrigins` must be `true`, and
PAPI defaults it to `false`, which the origin-restriction pallet rejects before dispatch with no dispatch error to read. And for the proof variant `VerifyMultiSignature` must be supplied as `Disabled` so the host assembles an unsigned general transaction, which is what makes the origin `None`.

`AliasWithAccount` works today. The two proof variants are correct and tested but need paseo on
individuality v0.12.0, which derives the allowed proof contexts the same way the host does. Paseo runs `specVersion 1000032`, v0.12.0 is `1000035`.

Wrapping the signer rather than the submitter is deliberate: the extension value depends on the nonce and on a hash over every extension after it, and PAPI only resolves those immediately before calling `signTx`.

### Testing

```
pnpm install && pnpm -r build
pnpm --filter "@parity/product-sdk-individuality" test
pnpm --filter "@parity/product-sdk" typecheck
pnpm check
```

Before, on `main`: individuality 57 tests passing, umbrella 19.
After: individuality 127 passing, umbrella 19, `pnpm -r typecheck` 0 errors, `pnpm check` clean on 302 files.

Not run against a live chain. Every byte-level claim comes from the deployed metadata blob, the pallet source and PAPI's encoder run offline. The byte layout and ordering tests were checked by breaking the implementation and confirming they fail, eleven times.


TarikGul (approved):
My Pass looks good, some claude findings:

1. (nit) Doc proofread — ensure no stale AliasWithProof-without-createProof snippet in the JSDoc examples.
2. (nit) metadata-builders version coupling is implicit but guarded by the loud version check + real-blob tests.
3. (nit) PROOF_BYTES_MAX / CONTEXT_BYTES are constants not metadata-read — both justified, CONTEXT_BYTES held honest by a test.
4. AliasWithProof "not usable yet" is a chain-runtime limitation (paseo Individuality ≤ v0.11.2 allowed-context constants), not a code stub — fully wired, correctly documented.

Imod7:
> My Pass looks good, some claude findings:
> 
>     1. (nit) Doc proofread — ensure no stale AliasWithProof-without-createProof snippet in the JSDoc examples.
> 
>     2. (nit) metadata-builders version coupling is implicit but guarded by the loud version check + real-blob tests.
> 
>     3. (nit) PROOF_BYTES_MAX / CONTEXT_BYTES are constants not metadata-read — both justified, CONTEXT_BYTES held honest by a test.
> 
>     4. AliasWithProof "not usable yet" is a chain-runtime limitation (paseo Individuality ≤ v0.11.2 allowed-context constants), not a code stub — fully wired, correctly documented.

1. Checked, and nothing stale. The only `AliasWithProof` snippet is in the skill doc and it
passes `createProof`. The two JSDoc snippets use `AliasWithAccount`, which takes no proof. The
git history confirms the snippet was never written without it, so no change needed.
2. Agreed. The version check is at `as-person-codec.ts:101` and the real-blob tests at `:385`, so
a `metadata-builders` bump that changed lookup or builder behaviour fails against deployed bytes rather than slipping through. Keeping the caret range, consistent with the other `@polkadot-api` catalog entries.
3. Correct on both. The test at `as-person-codec.ts:510` pins `CONTEXT_BYTES` to the length the deployed metadata declares, so a runtime that widened it breaks the build instead of encoding the wrong width. `PROOF_BYTES_MAX` cannot be metadata-read at all: the `BoundedVec` bound is a type parameter with no type in the metadata, so it is a deliberate sanity ceiling rather than a claim about the chain.
4. That is the intended reading, thanks for confirming it. The variant is wired and tested, at
`as-person-signer.ts:253` and `:622`; the blocker is the runtime fixing allowed contexts as constants through v0.11.2. Commit `fb43d95` exists to make that legible, since "not usable yet"
otherwise reads like a stub. Nothing here changes when paseo upgrades. Tracked on #290.

PR #310, opened 2026-08-19
fix(host): preserve chain-head operation ordering

replghost:
## Problem

TruAPI request responses and follow-subscription events travel on independent frames. A fast chain-head body, call, or storage operation can therefore emit its completion event before the provider publishes the JSON-RPC `Started(operationId)` response. PAPI drops that unknown operation event and can wait indefinitely.

## Changes

- track chain-head operations per follow subscription and operation id
- buffer operation events until their `Started` response has been delivered
- replay buffered storage items and terminal events in arrival order
- release state on terminal events, explicit stop, unfollow, subscription interruption, and provider close
- add regression coverage for early call completion and ordered storage replay
- add the required patch changeset for `@parity/product-sdk-host`

No public API or consumer migration is required.

## Verification

- `pnpm --filter @parity/product-sdk-host test -- --run` — 106 passed
- `pnpm --filter @parity/product-sdk-host typecheck`
- `pnpm exec biome check packages/host/src/papi-provider.ts`
- `git diff --check`

### DIFF
--- a/product-sdk/packages/host/src/papi-provider.ts
+++ b/product-sdk/packages/host/src/papi-provider.ts
@@ -61,6 +61,38 @@ const JSON_RPC_METHOD_NOT_FOUND = -32601;
 /** A `chainHead_v1_followEvent` payload (loosely typed — consumed by PAPI's substrate-client). */
 type FollowEvent = { event: string } & Record<string, unknown>;
 
+type BufferedOperation = {
+    announced: boolean;
+    items: RemoteChainHeadFollowItem[];
+};
+
+/** Return the operation id carried by an operation follow item. */
+function followOperationId(item: RemoteChainHeadFollowItem): string | undefined {
+    switch (item.tag) {
+        case "OperationBodyDone":
+        case "OperationCallDone":
+        case "OperationStorageItems":
+        case "OperationStorageDone":
+        case "OperationWaitingForContinue":
+        case "OperationInaccessible":
+        case "OperationError":
+            return item.value.operationId;
+        default:
+            return undefined;
+    }
+}
+
+/** Whether this item closes its operation. */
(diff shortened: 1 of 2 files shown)


product-sdk/packages/host/src/papi-provider.ts
```diff
@@ -209,6 +242,65 @@ export function createHostPapiProvider(
...
+                sendFollowEvent(followSubscriptionId, convertFollowEventToJsonRpc(item));
+                return;
+            }
+
+            const operations = followOperations.get(followSubscriptionId);
+            if (!operations) return;
+
+            let operation = operations.get(operationId);
+            if (!operation) {
+                operation = { announced: false, items: [] };
```
valentinfernandez1, line 268: A missing entry here doesn't only mean the Started response hasn't landed. It also means the op already
ended (terminal item at :269, stopOperation at :490, failed start request), and both the spec and PAPI's
cancel path can still emit events after that, so those get buffered and never freed until unfollow.
Suggestion: keep a per-follow counter of start requests still waiting for a response. Only buffer while
that counter is above zero. If it's zero, no Started response can claim this event, so forward it and let
PAPI ignore it like it did before.

valentinfernandez1 (approved):
Just one small comment, you can address it here or on a follow up

product-sdk/packages/host/src/papi-provider.ts
Imod7: Implemented your suggestion in [`00ecc33`](https://github.com/paritytech/product-sdk/commit/00ecc331d2e4b19c101943b7cdd791091f9cc7c3). One caveat: a stray landing while an unrelated start is in flight still buffers, so it narrows the window rather than closing it.

PR #311, opened 2026-08-20
feat(individuality): add the account to username read

### LINKED ISSUE #302: Account to username read over Resources.Consumers (lookupUsername is referenced but does not exist)
The SDK can answer "who owns username X" (resolvePeopleUsernameOwner, the one implemented function in identity/dotns.ts) but not the direction products actually need on a results or profile screen: "what is this account's username". That read is Resources.Consumers(account), which carries both names and the credibility:

```
{ identifier_key, lite_username, full_username?, credibility }
```

Display name is full_username, else lite_username, and "eligible to claim the bare name" is exactly full_username being unset. Both humanity-design-draft (app/src/chain/identity.ts) and dim2-spa (apps/web/src/lib/membership.ts) hand-roll this decode today.

The SDK already believes it has this helper: auth/src/auth.ts:66 documents SessionAddresses.rootAddress as "the right input for lookupUsername", and skills/product-sdk-transactions/references/auth-api.md:57 repeats it. lookupUsername does not exist anywhere in the repo.

Ask:

- a typed Consumers read: account in, { liteUsername, fullUsername, credibility } out, over the same structural PeopleUsernameQueryApi pattern resolvePeopleUsernameOwner uses, so it works with any chain that has the storage shape
- while touching this surface: resolvePeopleUsernameOwner takes storage that yields SS58 and returns 0x hex. Documented and correctly typed, but asymmetric, so a Consumers round trip needs a manual conversion on the caller side. Returning SS58 to match the input side (or both forms) would remove that trap.

Part of #286

Imod7:
Closes #302.

### Description

Adds the account to username read: `lookupUsername(chain, { account })` reads `Resources.Consumers` and returns that account's lite and full usernames plus its credibility. Also makes chaining the two directions conversion free, by having `resolvePeopleUsernameOwner` return the SS58 the storage yields rather than `0x` hex, which is the form `Resources.Consumers` is keyed by.

### Changes

`packages/individuality/src/username.ts` (new). `decodeConsumerInfo`, `lookupUsername`, and three pure helpers: `displayUsername` (claimed name, else lite), `canClaimFullUsername`, `usernameBase`. Names decode strictly, so malformed bytes fail rather than becoming U+FFFD. An unknown credibility variant throws rather than reading as `Lite`. An account with no record is `ok(null)`, not an error. The chain is typed structurally as `ConsumersChain`, one storage entry, so a double does not have to implement the personhood read's six.

`packages/sdk/src/individuality/contract.test.ts`. Two type-level assertions that real paseo and devnet clients satisfy `ConsumersChain`, plus a negative control. The value shape was derived from the pallet and the codegen, so this is what makes it a checked fact: a descriptor regeneration that changes the entry fails `pnpm typecheck` instead of failing inside a product.

`packages/sdk/src/identity/dotns.ts`, `core/createApp.ts`, `identity/index.ts`. The resolver returns `SS58String`, and a new `accountIdToHex` is exported for callers that need hex. `signMessageWithDotNsIdentity` loses a conversion, SS58 to bytes instead of SS58 to hex to bytes, and its public `accountId` is still hex.

`packages/sdk/src/identity/dotns-owner.test.ts` (new). Five tests. The last chains both reads and asserts the storage key the owner reaches, which guards the conversion this PR removes.

Docs. `packages/auth/src/auth.ts` and `skills/product-sdk-transactions/references/auth-api.md` both pointed callers at a `lookupUsername` that existed nowhere; they now name the package. `skills/product-sdk-individuality/SKILL.md`, `CLAUDE.md` and `packages/individuality/{src/index.ts,package.json}` document the new direction. `pending-changesets/individuality-username-read.md` is minor for `@parity/product-sdk-individuality` and `@parity/product-sdk`, patch for `@parity/product-sdk-auth`.

### Why these changes

The SDK could answer "who owns username X" but not the direction a profile or results screen needs, "what is this account's username". Two apps and the host's Rust core each hand-rolled the same decode.

Four properties come from the pallet rather than the descriptor, so the compiler cannot show them, and each is a way to get this wrong. All four verified against `paritytech/individuality` at tag `v0.12.0`:

- A lite username is always present and always `<letters>.<digits>`. A full one is letters only, no dot.
- `fullUsername === null` is the chain's own precondition for claiming a bare name, which is what `canClaimFullUsername` reports.
- A demoted person keeps `Person` and their full username, so `credibility.tag` alone is not "in good standing".
- Neither is `demoted: false`. The chain sets that flag only when someone submits `demote_auth_expired`, and nothing does so automatically, so it also covers an authorization that expired days ago. `credibility.lastUpdate`, seconds since the epoch, is surfaced to compare against `PersonAuthDuration`.

### Breaking changes

Both are covered by the minor bump and both can break a consumer's build.

`resolvePeopleUsernameOwner` returns `SS58String` instead of `` `0x${string}` ``. The old hex return made every round trip carry a manual conversion. Callers needing hex use `accountIdToHex` from `@parity/product-sdk/identity`. `wallet.signMessageWithDotNsIdentity` is unaffected, its `accountId` is still hex.

`RawConsumerInfo` requires `last_update` on the credibility payload, so a product hand-building it as a test double needs the extra field.

### Testing

```bash
cd product-sdk
pnpm install
pnpm -r build
pnpm -r --no-bail typecheck
pnpm -r --no-bail test
pnpm check
```

Build before typecheck, the order CI uses: the contract test resolves the individuality package through its `types` field, so typechecking without a build asserts against the previous one.

Green: workspace typecheck 0 errors, `pnpm check` clean, all package suites passing. The new read adds 27 tests, and none of it existed to test before.

Nothing here has talked to a real chain. Every test uses a structural double that satisfies `ConsumersChain` by construction, so a live read against paseo is worth doing before merge.


product-sdk/packages/individuality/src/username.ts
```diff
@@ -0,0 +1,581 @@
...
+ *
+ * The chain also sends `identifier_key`, an opaque communication key with no
+ * bearing on names. Extra fields on the actual value are accepted; this is a
+ * structural type, not an exhaustive record of the storage entry.
+ */
+export interface RawConsumerInfo {
+    lite_username: Uint8Array;
+    full_username?: Uint8Array | undefined;
+    credibility: {
+        type: string;
```
valentinfernandez1, line 46: What is type here? I can make a guess but a more explicit comment would really help

product-sdk/packages/individuality/src/username.ts
```diff
@@ -0,0 +1,581 @@
...
+        value?: { alias: string; last_update: bigint; demoted: boolean } | undefined;
```
valentinfernandez1: Isn't the `undefined` case already covered by `value?`

product-sdk/packages/individuality/src/username.ts
```diff
@@ -0,0 +1,581 @@
...
+           * JSON-serializable and matches every other numeric in this package. A
+           * seconds timestamp is exact in a double until year 285 million.
+           */
+          lastUpdate: number;
+          demoted: boolean;
+      };
+
+/** The usernames registered for one account, decoded. */
+export interface ConsumerUsernames {
+    /** Always present. Letters, one dot, then digits, for example `bigtava.07`. */
```
valentinfernandez1: Can we switch the example to something generic so we dont have Tiago's username

product-sdk/packages/sdk/src/core/createApp.ts
```diff
@@ -321,16 +322,19 @@ function createWalletApi(signerManager: SignerManager): WalletApi {
...
-            const owner = accountIdHexToBytes(accountId);
             const signer = accountsProvider.getLegacyAccountSigner({
-                publicKey: owner,
+                publicKey: accountIdBytes(owner),
                 name: username,
             });
+            // The public result stays hex; only the resolver's return type changed.
+            // Converted outside the try, so an address failure is never reported
+            // as a signing failure.
+            const accountId = accountIdToHex(owner);
```
valentinfernandez1: Consider using
```typescript
    const publicKey = accountIdBytes(owner);
    const accountId = accountIdBytesToHex(publicKey);
    const signer = accountsProvider.getLegacyAccountSigner({ publicKey, name: username });
```

As right now accoutIdBytes runs twice, on the first invocation and on accountIdToHex()

valentinfernandez1 (commented):
I have some nits worth fixing. Additionally I am seeing a lot of bloat from code docs that just restates what has already been said

product-sdk/packages/sdk/src/core/createApp.ts
```diff
@@ -321,16 +322,19 @@ function createWalletApi(signerManager: SignerManager): WalletApi {
             const peopleApi = peopleClient.getTypedApi(
                 args.peopleChain,
             ) as unknown as PeopleUsernameQueryApi;
-            const accountId = await resolvePeopleUsernameOwner(username, peopleApi);
-            if (!accountId) {
+            const owner = await resolvePeopleUsernameOwner(username, peopleApi);
```
TarikGul, line 325: Traced this the whole way, `resolvePeopleUsernameOwner` going SS58 instead of hex only has one real caller (`signMessageWithDotNsIdentity`), and it's handled right: `accountIdBytes(owner)` for the signer, and the public `accountId` stays hex via `accountIdToHex`, so nothing consumer-facing changes. Nice touch moving the hex conversion outside the try so an address failure doesn't look like a signing failure.

product-sdk/packages/sdk/src/identity/dotns.ts
```diff
@@ -174,13 +180,11 @@ export function normalizeDotNsName(name: string, suffix: string): string {
 export async function resolvePeopleUsernameOwner(
     username: string,
     peopleApi: PeopleUsernameQueryApi,
-): Promise<`0x${string}` | null> {
+): Promise<SS58String | null> {
```
TarikGul, line 183: Just noting for the record: this is `@internal` but we do still re-export it from the `identity` subpath, so a return-type change is technically observable to anyone who ignored the tag. The `minor` on the changeset already treats it as breaking, which is the right call — no change needed, just flagging we're aware.

product-sdk/packages/individuality/src/username.ts
```diff
@@ -0,0 +1,581 @@
...
+ * `null` has to mean exactly `full_username.is_none()`, since that is what
+ * {@link canClaimFullUsername} reports. An empty name is `Some("")` on chain and
+ * the claim path rejects it, so reading it as absent would offer a claim the
+ * chain refuses. On-chain length validation makes it unreachable anyway.
+ */
+function optionalUsername(bytes: Uint8Array | undefined): string | null {
+    if (bytes === undefined) return null;
+    const username = decodeUsername(bytes);
+    if (username.length === 0) {
+        throw new IndividualityDecodeError("consumer record has an empty full username");
```
TarikGul, line 119: Good catch fixing the empty full-username case here — mapping `Some("")` to `null` would've let `canClaimFullUsername` offer a claim the chain rejects. And since `username.ts` wasn't exported before this PR, it never shipped, so the plain `minor` covers it.

TarikGul (approved):
Some comments overall awesome.

product-sdk/packages/individuality/src/username.ts
Imod7: `type` is how PAPI encodes a Rust enum: variant name in `type`, payload in `value`, so here `Lite`
or `Person` from the pallet's `Credibility`. The comment names both now. I also removed the copy of
that fact inside `decodeCredibility`, so it is stated once rather than twice.

Fixed in [1dc42e5](https://github.com/paritytech/product-sdk/commit/1dc42e56e9af7a762ca64e73a52ccc36bcd7ac24)

product-sdk/packages/individuality/src/username.ts
Imod7: You are right. Removed here and on `full_username` above. It would matter under
`exactOptionalPropertyTypes`, because PAPI types a unit variant as `{ type: "Lite"; value: undefined }`, a present undefined rather than an absent property. That flag is off, and the umbrella typecheck that holds the descriptor contract stays green without it.

Fixed in [1106773](https://github.com/paritytech/product-sdk/commit/11067735640be2de5a92bb1230dccb49ef6bc052)

product-sdk/packages/individuality/src/username.ts
Imod7: Changed to `example` and `example.07`, here and in the individuality skill. I kept the stem at six letters or more on purpose: `MinUsernameLength` is 6 on paseo, so a shorter placeholder would be a name the chain rejects. The rename also exposed a duplicated assertion in the `usernameBase` test, fixed in the same commit.

Fixed in [946fc96](https://github.com/paritytech/product-sdk/commit/946fc969041b548f8a955edc67ac7fa0e441d575)

product-sdk/packages/sdk/src/core/createApp.ts
Imod7: Taken as suggested, the bytes are bound once and the hex comes from them. One addition: `accountIdBytesToHex` was private, so I exported it from `dotns.ts`, which keeps it off the package's public subpath since `createApp` imports the module directly. `accountIdToHex` stays exported as the migration path for the SS58 return change.

Fixed in [d8fa6d1](https://github.com/paritytech/product-sdk/commit/d8fa6d191ca26d16e09554355bcfbedd9f39f2eb)

Imod7:
> I have some nits worth fixing. Additionally I am seeing a lot of bloat from code docs that just restates what has already been said

Three facts were each written out three times, the U+FFFD hazard, why `lastUpdate` is a timestamp rather than a boolean, and the lite username grammar, plus a few paraphrases of the line below. Each now sits once, where it is most likely to be broken: 54 comment lines out, 20 in. The pallet facts in the module header stay, since the types cannot tell you those.

Fixed in [3c71303](https://github.com/paritytech/product-sdk/commit/3c71303b6ed0385c194764455a2055b59a42a412)

product-sdk/packages/sdk/src/core/createApp.ts
Imod7: Thanks for tracing it. Worth flagging that the line you quoted moved after you looked: Valentin spotted that `accountIdBytes` ran twice, so the bytes are bound once now and the hex comes from those via `accountIdBytesToHex`. Your conclusion still holds, the public `accountId` is hex and the conversion stays outside the try.

Changed since you looked in [d8fa6d1](https://github.com/paritytech/product-sdk/commit/d8fa6d191ca26d16e09554355bcfbedd9f39f2eb)

product-sdk/packages/sdk/src/identity/dotns.ts
Imod7: Agreed on both counts, and the changeset calls the return-type change out as the reason for the minor. Since the last commit, `accountIdToHex` also carries a note saying it has no in-repo caller on purpose, because it is the migration path for anyone who was relying on the hex return, so nobody deletes it as dead code.

Documented in [d8fa6d1](https://github.com/paritytech/product-sdk/commit/d8fa6d191ca26d16e09554355bcfbedd9f39f2eb)

product-sdk/packages/individuality/src/username.ts
Imod7: Thanks. That one came out of the self-review: mapping `Some("")` to `null` made
`canClaimFullUsername` disagree with the chain, which checks `full_username.is_none()`. It is unreachable in practice, since `MinUsernameLength` is 6, but it made the helper's doc claim false. And agreed on the bump, nothing shipped with the old behaviour.

Landed in [276983c](https://github.com/paritytech/product-sdk/commit/276983c8b78353bc2255a168dead176d00c70221)

valentinfernandez1 (approved):

PR #313, opened 2026-08-20
Add the manual dev-publish workflow to main so it can be dispatched

BigTava:
## Summary

- Puts `product-sdk-dev-publish.yml` on `main`, which is where GitHub requires a `workflow_dispatch` workflow to live before the Run workflow button appears.
- The file is unchanged from the copy on `feat/worker-manager`, cherry-picked so the two stay identical.
- Once this lands, the workflow can be dispatched against any feature branch; it still refuses to cut a dev release from `main` itself.
- Nothing publishes as a result of merging this. The first cut needs an explicit dispatch with a PR number and iteration.


BigTava:
Closing this. We are keeping the workflow on the feature branch for now rather than landing it on main. Reopen if we change our minds.

PR #314, opened 2026-08-20
feat(identity): discover and verify DotNS addresses

### LINKED ISSUE #303: dotNS registry surface: the default Asset Hub contract addresses have no code on chain
The tg-dotns-registry-impl branch hardcodes default contract addresses in dotns-abis.ts (registry 0x4Da0d37aBe96C06ab19963F31ca2DC0412057a6f, registrarController 0xd09e0F1c1E6CE8Cf40df929ef4FC778629573651; the file already carries a TODO doubting them). Checked them with pallet-revive dry runs (state_call ReviveApi_call): none of them has contract code on paseo-asset-hub-next, and none on the previewnet Asset Hub either. Calls return Ok with empty data. The same holds for the "Paseo Asset Hub Next V2" list in paritytech/dotns DEPLOYMENTS.md and for the registry in dotns-sdk's cdm.json (0xf62c2ece...), so picking any documented list is fragile.

The live deployment is discoverable from chain state with public getters, no hardcoding:

1. DotnsGateway.DispatcherAddress (pallet storage on Asset Hub)
2. dispatcher.TARGET() (RootGatewayDispatcher.sol, public immutable)
3. popController.protocolRegistry() (DotnsPopController.sol)
4. protocolRegistry.get(bytes32("registry")), same for "registrar", "reverseResolver", "contentResolver", "popRules" (keys in DotnsConstants.sol)

Verified on both paseo-asset-hub-next and the previewnet Asset Hub: the walk lands on the same deterministic CREATE3 set on both chains (registry 0xf34054fd76BbF85f216cf9908226D5f0A72E50CA, registrar 0x4f06E818Ba3d987704fd91cf3d868E4b019106Ab, and so on), and probes against the discovered addresses behave: registry.owner(namehash("dim2.paseo")) is owned on next, owner(namehash("dim2.dot")) is owned on previewnet, registrar.available(labelhash) answers. Runnable proof: dim2-spa apps/web/scripts/verify-uj6-dotns.mjs (deduction D9).

Ask: resolve the addresses through the gateway walk at runtime, or at least verify the hardcoded defaults against DotnsProtocolRegistry at startup and fail loudly on a mismatch. Also worth a look while in there: the branch derives availability from resolver-zero rather than the registrar's own available(), and those are not the same question once reservations exist.

Part of #286

Imod7:
Closes #303
Part of #286

### Description

Adds a way to resolve the DotNS contract addresses from chain state instead of only trusting the table compiled into the SDK, and a startup check that reports when that table has gone stale. The pinned table stays the default, so nothing changes unless a product opts in.

### Changes

`packages/sdk/src/identity/dotns-addresses.ts` (new). Owns address resolution. `resolveDotNsAddresses` is the single entry point, `discoverDotNsAddresses` walks the deployment from chain state, `verifyDotNsAddresses` compares what a client calls against what the chain says. Also holds `DOTNS_REGISTRY_KEYS` and the shared `cachedPerRuntime` helper.

`packages/sdk/src/identity/dotns-registry.ts`. Every entry point now asks `resolveDotNsAddresses` instead of writing `opts.xAddress ?? DOTNS_ADDRESSES.x`. Ten of those expressions are gone. Adds the `addressSource` and `gatewayApi` options. `resolveTld` reuses the shared cache.

`packages/sdk/src/identity/dotns-abis.ts`. Adds `get(bytes32)` to the protocol registry ABI, plus ABIs for the gateway dispatcher and the PoP controller.

`packages/sdk/src/identity/dotns-errors.ts`. Two reasons, `AddressDiscovery` and `AddressMismatch`. A `switch` over `DotNsErrorReason` in consumer code is no longer exhaustive.

`packages/sdk/src/identity/index.ts`. Exports the four new functions and three new types.

Tests. New `dotns-addresses.test.ts` with 36 tests, and 8 added to `dotns-registry.test.ts` proving the discovered addresses reach every entry point.

### Why these changes

The addresses were pinned and nothing checked them. An earlier default set had no code deployed at any of its addresses, and every read came back as "unregistered" with no error anywhere. Issue #303 asked for either the gateway walk or a startup check. This does both, since the check is the walk plus a comparison.

Two things worth knowing before enabling discovery. The registry keys are not the field names: `registrarController` answers under `bytes32("controller")`, and `registrar` is a different contract at a different address. And discovery also selects the address the write helpers build calls against, so it decides where a signed transaction goes, not only where a read comes from.

### Testing

```
cd product-sdk
pnpm install && pnpm -r build
pnpm check
```

Before: `main` has no tests for address resolution, and 84 registry tests. The package suite is 143 tests.

After: 187 tests pass, 44 new. `pnpm check` clean, `tsc --noEmit` clean, all 38 packages build.

Also run against two live chains, driving the built package rather than a reimplementation. On Paseo Asset Hub Next V2 and on Previewnet the walk lands on all six pinned addresses and `verifyDotNsAddresses` reports no drift. That confirms the key mapping by measurement: `get(bytes32("controller"))` returns the registrar controller, and `get(bytes32("registrar"))` returns something else.

The same run also corrected a stale claim in the code. Previewnet reports TLD `.test`, not `.dot`, so four comments naming it as a `.dot` deployment were wrong and are fixed. No behaviour depends on it.

Not covered: there is no automated live test, so nothing detects rot. That is an existing tracked follow-up, and this change narrows it without closing it.

### DIFF
--- a/product-sdk/packages/sdk/src/identity/dotns-abis.ts
+++ b/product-sdk/packages/sdk/src/identity/dotns-abis.ts
@@ -238,6 +238,39 @@ export const DOTNS_PROTOCOL_REGISTRY_ABI: AbiEntry[] = [
         outputs: [{ name: "node", type: "bytes32" }],
         stateMutability: "view",
     },
+    {
+        type: "function",
+        name: "get",
+        inputs: [{ name: "key", type: "bytes32" }],
+        outputs: [{ name: "addr", type: "address" }],
+        stateMutability: "view",
+    },
+];
+
+/**
+ * `RootGatewayDispatcher.TARGET`, the PoP controller it forwards to. `immutable`,
+ * so a rotation means a new dispatcher and a new pallet value, which is why
+ * discovery starts at the pallet rather than here.
+ */
+export const DOTNS_ROOT_GATEWAY_DISPATCHER_ABI: AbiEntry[] = [
+    {
+        type: "function",
+        name: "TARGET",
+        inputs: [],
+        outputs: [{ name: "", type: "address" }],
+        stateMutability: "view",
+    },
+];
+
+/** `DotnsPopController.protocolRegistry`, the only hop that names the registry. */
+export const DOTNS_POP_CONTROLLER_ABI: AbiEntry[] = [
+    {
+        type: "function",
+        name: "protocolRegistry",
+        inputs: [],
+        outputs: [{ name: "", type: "address" }],
+        stateMutability: "view",
+    },
 ];
 
 /** `IPopRules.PopStatus`. Ordering is meaningful: a user meets a tier when `userStatus >= status`. */
@@ -258,11 +291,13 @@ export const POP_STATUS = { NoStatus: 0, PopLite: 1, PopFull: 2, Reserved: 3 } a
  * So extending this module to another network means resolving its TLD, which
  * `resolveTld` already does — not adding a second address table.
  *
- * Each address answers `DotnsProtocolRegistry.get(bytes32)` under its
- * `DotnsConstants` key (the name below, right-padded to 32 bytes) and has a
- * `Revive.AccountInfoOf` entry. They are pinned rather than resolved at runtime,
- * so re-verify after a redeploy; `protocolRegistry` is the address to walk from
- * if we ever resolve the rest dynamically.
+ * Five answer `DotnsProtocolRegistry.get(bytes32)` under the keys in
+ * `DOTNS_REGISTRY_KEYS`, two of which differ from the field name used here.
+ * `protocolRegistry` has no key and comes from the gateway walk. All six have a
+ * `Revive.AccountInfoOf` entry.
+ *
+ * Pinned, not resolved at runtime: re-verify after a redeploy, or have the
+ * client do it with `verifyDotNsAddresses`.
  */
 export const DOTNS_ADDRESSES = {
     registry: "0xf34054fd76BbF85f216cf9908226D5f0A72E50CA",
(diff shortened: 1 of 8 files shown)


product-sdk/packages/sdk/src/identity/dotns-abis.ts
TarikGul: Small one, this comment says the addresses are "the same on every network" and "verified on Paseo and Previewnet," but in this same PR you corrected the Previewnet TLD over in `dotns-registry.ts:15-16` (`.paseo` on Paseo, `.test` on Previewnet). Since you're already touching this block, worth reconciling the two Previewnet references so they don't drift.

product-sdk/packages/sdk/src/identity/dotns-addresses.ts
```diff
@@ -0,0 +1,317 @@
...
+            protocolRegistry.value,
+            DOTNS_PROTOCOL_REGISTRY_ABI,
+        );
+        const roles = Object.entries(DOTNS_REGISTRY_KEYS) as [
+            keyof typeof DOTNS_REGISTRY_KEYS,
+            HexString,
+        ][];
+        const found = await Promise.all(
+            roles.map(async ([role, key]) => {
+                const res = await registry.get.query(key, { origin });
```
TarikGul, line 261: Noticed the role walk inlines `registry.get.query(key, {origin})` at :258 instead of going through the `readAddress` helper at :175. Fine as-is, but a one-line comment on *why* `get` doesn't reuse `readAddress` (it takes a positional key and wants the role name in the error) would save the next person the double-take.

TarikGul (approved):

product-sdk/packages/sdk/src/identity/dotns-abis.ts
Imod7: Fixed in [82667a3](https://github.com/paritytech/product-sdk/commit/82667a3a2a4e1bef09b13ea25ac352537962f8d5)

product-sdk/packages/sdk/src/identity/dotns-addresses.ts
Imod7: Added in [851cfc7](https://github.com/paritytech/product-sdk/commit/851cfc74434b35ee72576d61eebf47f928307423)

valentinfernandez1 (approved):

PR #316, opened 2026-08-20
feat(individuality): read by account, and return the metrics

### LINKED ISSUE #287: Personhood status helpers over the People chain

Imod7:
Closes #287.
Part of #286. 

The eligibility half, dim2-spa's recognition checks, stays with #291.

### Description

`readPersonhoodState` now takes an account as well as a DotNS username, and every resolved result carries the numbers behind the state.

### Changes

`read.ts`. Options are a `username` or an `account`, never both. An account skips the `Resources.UsernameOwnerOf` lookup instead of adding a read, so it costs one round trip less and `UsernameUnowned` is unreachable on it. `selectInput` validates before the block is pinned, so both inputs, neither, or an empty string is an `err` that costs nothing.

`types.ts`, `index.ts`. New `PersonhoodMetrics` on every `Resolved` result: `score`,
`personhoodThreshold`, `misses`, `allowedMisses`, `window`, off the same pinned block.

`derive.ts`. `missesInWindow` extracted and exported, `projectedMisses` composes it, mirroring the pallet's `store_attendance` then `misses_in_window`.

`SKILL.md` and the changeset cover both, including that `metrics.misses` is the count already taken while `Caution.misses` projects one more absence.

### Why these changes

A profile screen holds an account, not a name, so the read made it find a username first and then spent a lookup turning it back. And the variants carry numbers only where the derivation needed them, so a screen rendering progress could not read a score off `Member`.

Left out on purpose: the `NotMember` probe, which answers only for the local user's own key and needs a registration that writes host-local state, and `AccountToPersonalId`, which the
specification says should not reach the client. Note the option type does not enforce either-or,
`{ username: maybeName, account: maybeAccount }` with both `string | undefined` compiles, so the runtime check holds the rule.

### Testing

```bash
pnpm install && pnpm -r build
pnpm --filter "@parity/product-sdk-individuality" test -- --run
pnpm --filter "@parity/product-sdk" test -- --run
pnpm typecheck && pnpm check
```

Ten new tests. 164 in the package and 148 in `@parity/product-sdk` after merging `main`, with
build, typecheck and check clean. Both guards were measured: `{ username, account }` used to return the account holder's standing after a block fetch and now returns an `err` with none, and the parity tests fail if the clamp or the window mask is removed.


TarikGul (approved):
🚀  Nice job

PR #319, opened 2026-08-21
feat(individuality): add the game sign-up

Imod7:
Part of #286.
Refs #291, does not close it.

### Description

Adds game sign-up to `@parity/product-sdk-individuality`: what the chain will accept from a player now, the airdrop VRFs the sign-up carries, and the call. Step 5 of #291's build order. Steps 6 and 7 need chain or host changes, see below.

### Changes

`individuality/src/signup-types.ts` The domain types. `AirdropVrfVariant` is read from the chain, never chosen. `SignUpBlocker` reports every cause rather than the first, and splits four that stop the extrinsic from four that stop only the draw entry.

`individuality/src/signup-vrf.ts` The Merlin transcript an `AirdropVrfs::Account` entry signs. Neither the label nor the domain prefix reaches metadata, so pinned byte vectors are the only guard. Imports no workspace package, so its tests run against an unbuilt tree.

`individuality/src/signup.ts` `readGameSignUpRequirement` reads the game, `Score.Participants` and `Game.Players` at one pinned block. `mintAccountAirdropVrfs` signs one VRF per draw, sequentially, through a structural signer so the package keeps no dependency on the host package. `signUpWithAccountTx` builds the unsigned call. Submission stays with `@parity/product-sdk-tx`.

`individuality/src/index.ts`, `as-person-signer.ts` Exports, plus a note on the two `withAsPerson` examples: they show the origin, which works, and the rest of that call cannot be assembled.

`sdk/src/individuality/contract.test.ts` Compile-time assertions that a real client satisfies `SignUpChain`, that devnet fails the intersection the read takes, and that devnet's argument is `airdrop` where paseo's is `airdrops`.

`skills/product-sdk-individuality/SKILL.md`, `README.md`, `packages/individuality/package.json` The sign-up section, placed before the write section so the blocked path is stated ahead of the example a reader would copy. The skill frontmatter and the package descriptions now name the game and sign-up surfaces, which they did not, so questions about them never routed here. Plus one changeset.

### Why these changes

Entering a user into a game and its draws meant building `Game.sign_up_with_account` and minting the VRFs by hand, and two products need the same logic. It also puts three easy-to-miss facts into types: the variant is the chain's choice and `Suspended` counts as not recognized, the entry count must equal `airdrops_scheduled` exactly or the sign-up fails with the deposit, and only sr25519 accounts can take this path because the pallet reads the account id as the public key and no chain read reveals the scheme.

Two parts of #291 are deliberately absent.

Entering the draws as a recognized player needs a ring-VRF proof at `blake2_256("pop:polkadot.network/airdrop" ++ event_id)`, and every context a host will sign under is `blake2b_256("product/" ++ productId ++ "/" ++ suffix)`, which the host computes itself. `sign_up_with_alias` also needs a statement-account proof over a bare `blake2_256` hash, and the host's `signRaw` always wraps what it signs in `<Bytes>`. Both need a chain or host change, so they surface as the `AliasVrfsUnavailable` and `AccountVrfsNeedAnAccount` blockers rather than as a call that fails on chain. That matters because a rejected sign-up costs a fee: `Pays::No` applies on success only, and the airdrop registration rides inside the same extrinsic.

Step 7, the `PeopleAirdrops` path, is untouched and still has no origin any product can produce.

### Testing

```
cd product-sdk
pnpm install
cd packages/descriptors && pnpm generate && cd ../..
pnpm -r build
pnpm --filter "@parity/product-sdk-individuality" test
pnpm --filter "@parity/product-sdk-individuality" build
pnpm -r typecheck
pnpm check
```

Build the individuality package before the umbrella typecheck, or it resolves a stale `dist`. That typecheck is the load-bearing check: it proves a real client satisfies `SignUpChain`, where the same assertion inside the package passes against a bogus contract. Confirmed by narrowing the interface, which makes it fail.

`pnpm generate` rewrites eight `chains/*/package.json` files. Not part of this branch, revert with `git checkout -- packages/descriptors/chains/`.

All green: individuality 377 tests, up from 350 on `main`. Umbrella 192, unchanged. 19 packages typecheck, `pnpm check` clean over 329 files.

The last two commits apply two rounds of local review, one Critical, four Moderate, four Optional, each with a test. The Critical is worth knowing when reading the diff: draw entry is gated on recognition and on the origin variant, and modelling only the first made an unrecognized person read as able to enter the draws when the chain rejects the whole sign-up.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -89,6 +89,6 @@ When the user's question matches a skill, invoke it via the `Skill` tool rather 
 - `product-sdk-contracts` — contract calls (queries, txs).
 - `product-sdk-cloud-storage` — cloud-storage chain client.
 - `product-sdk-statement-store` — statement store.
-- `product-sdk-individuality` — personhood / membership state reads, and the account to username read.
+- `product-sdk-individuality` — personhood / membership state reads, the account to username read, the game and its prize draws, game sign-up, and the AsPerson extension.
 - `product-sdk-utilities` — address, crypto, logger, local-storage, utils.
 - `migrating-to-product-sdk` — porting from legacy stacks.
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ TypeScript SDK for building products in the Polkadot ecosystem. Provides typed A
 | `@parity/product-sdk-contracts` | Typed contract interactions on Polkadot Asset Hub |
 | `@parity/product-sdk-cloud-storage` | Upload and retrieve data via Cloud Storage (currently backed by the Polkadot Bulletin Chain) |
 | `@parity/product-sdk-statement-store` | Publish/subscribe client for the Polkadot Statement Store |
-| `@parity/product-sdk-individuality` | Read a person's personhood state and the usernames an account holds, and dispatch calls under a person origin |
+| `@parity/product-sdk-individuality` | Read a person's personhood state, usernames, and the game and its prize draws, sign up for a game, and dispatch calls under a person origin |
 | `@parity/product-sdk-keys` | Hierarchical key derivation, session keys, and sr25519 product-account derivation |
 | `@parity/product-sdk-local-storage` | Key-value local storage with automatic host/browser backend detection |
 | `@parity/product-sdk-host` | Host container detection and storage access for Desktop/Mobile |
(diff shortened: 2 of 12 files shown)


TarikGul (approved):
Clean, nice job. Checked it against the runtime.

Issue #322, opened 2026-08-21
contracts-demo e2e: redeploy @t3rminal/bulletin-index — target contract reaped in Paseo AH genesis reset


TarikGul:
The `contracts-demo` e2e query tests target the `@t3rminal/bulletin-index` contract at `0xD35CFc6E9F07B42Cc524A9Fa4A001F6ac90586E2` on Paseo Asset Hub. That contract **no longer exists on-chain** — it was reaped in the genesis reset (cf. the "re-pin paseo chains after genesis reset" descriptor bump, #297).

**Evidence** (raw `ReviveApi.call` dry-run against live Paseo AH, specVersion `2000036`):

- `result.success: true`, `flags: 0`, `data: 0x` (0 bytes), `gas_consumed: 0`
- `Revive.AccountInfoOf` / `CodeInfoOf` for that address both return `undefined` — no code deployed there.

A call into a codeless address is a zero-gas no-op returning empty data, which the SDK correctly decodes to `undefined`. So the three deterministic-empty assertions (`getReportCount → 0`, `getAllDates → []`, `getCID → ""`) fail.

**This is stale chain state, not an SDK regression.** The `wrap.ts` "PAPI 2.x boundary" decode suite (66 tests) and all 162 contracts unit tests pass. It is *not* the class-based `Binary`/`FixedSizeBinary` regression the spec header warns about — that would fail with `Incompatible runtime entry RuntimeCall(ReviveApi_call)` before decode; here the call succeeds.

**Fix required:**

1. Redeploy `@t3rminal/bulletin-index` to Paseo Asset Hub.
2. Update the address in `examples/contracts-demo/src/main.ts` and `examples/contracts-demo/src/cdm.json`.
3. Un-skip the `query` describe block in `examples/contracts-demo/e2e/query.spec.ts` (skipped in the tracking PR, marked `TODO(contracts-demo-redeploy)`).

PR #323, opened 2026-08-21
test(e2e): skip contracts-demo query specs against a reaped contract

TarikGul:
The `contracts-demo` e2e query tests target `@t3rminal/bulletin-index` at `0xD35C…86E2`, which no longer exists on Paseo AH (reaped in the genesis reset). A `ReviveApi.call` dry-run there returns `success:true` with 0-byte data and 0 gas → the SDK correctly decodes `undefined` → the deterministic-empty assertions fail.

This is stale chain state, **not** an SDK regression, all 162 contracts unit tests, including the `wrap.ts` PAPI 2.x boundary decode suite, pass. Only `contracts-demo` fails in the whole e2e run; every other demo is green.

Skips the `query` describe block (mirroring the existing `cloud-storage` and `signing-rejected` skips) so the E2E workflow stops being blocked. Re-enable once the contract is redeployed and the address + `cdm.json` are updated.

Tracking: #322

### DIFF
--- a/product-sdk/examples/contracts-demo/e2e/query.spec.ts
+++ b/product-sdk/examples/contracts-demo/e2e/query.spec.ts
@@ -40,7 +40,17 @@ import { waitForAppReady } from "./helpers";
  */
 const FRESH_SHOP_KEY = ("0x" + randomBytes(32).toString("hex")) as `0x${string}`;
 
-test.describe("@parity/product-sdk-contracts via Host API — query", () => {
+// TODO(contracts-demo-redeploy): the @t3rminal/bulletin-index contract at
+// 0xD35CFc6E9F07B42Cc524A9Fa4A001F6ac90586E2 no longer exists on Paseo Asset
+// Hub — it was reaped in the genesis reset (see the "re-pin paseo chains after
+// genesis reset" descriptor bump). A ReviveApi.call dry-run to that address now
+// returns success:true with flags:0 and empty (0-byte) output data and 0 gas,
+// which the SDK correctly decodes to `undefined`, so these deterministic-empty
+// assertions (uint256 0, string[] [], string "") fail. This is stale chain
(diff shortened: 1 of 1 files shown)


product-sdk/examples/contracts-demo/e2e/query.spec.ts
```diff
@@ -40,7 +40,17 @@ import { waitForAppReady } from "./helpers";
...
+// TODO(contracts-demo-redeploy): the @t3rminal/bulletin-index contract at
+// 0xD35CFc6E9F07B42Cc524A9Fa4A001F6ac90586E2 no longer exists on Paseo Asset
+// Hub — it was reaped in the genesis reset (see the "re-pin paseo chains after
+// genesis reset" descriptor bump). A ReviveApi.call dry-run to that address now
+// returns success:true with flags:0 and empty (0-byte) output data and 0 gas,
+// which the SDK correctly decodes to `undefined`, so these deterministic-empty
+// assertions (uint256 0, string[] [], string "") fail. This is stale chain
+// state, not an SDK regression: the wrap.ts "PAPI 2.x boundary" unit tests that
+// cover this exact decode all pass. Re-enable once the contract is redeployed
+// and the address + cdm.json are updated. Tracking: <issue link>.
```
Imod7:
```suggestion
// and the address + cdm.json are updated. Tracking: #322
```

Imod7 (approved):
The test `storeDailyReport tx lands in best block via host signing` in
`product-sdk/examples/contracts-demo/e2e/submit.spec.ts` targets the same address as the query specs you are skipping, and it passes, because a call into an address with no code is a successful no-op.
It still covers host signing and submission, but it no longer covers anything about the contract, while the file header says it exercises the write path. Please either skip it alongside the query block or add a line saying what it does and does not cover right now, and add the file to the checklist on #322. Worth a quick look at `cloud-storage-demo` for the same shape.

PR #325, opened 2026-08-23
fix(signing): prefer signed V4 on dual-version runtimes

replghost:
## Summary

- prefer Extrinsic V4 whenever a runtime advertises it, so product-account signatures use the signed envelope
- retain the previous highest-version behavior for V5-only and unknown future runtimes
- apply the same policy to host and terminal product-account signers
- add focused version-selection regressions and a staged patch changeset

## Why

V5 General transactions have no envelope signature. Their authorization is defined by a runtime-specific transaction-extension pipeline, and runtime metadata does not describe whether the connected host can implement that pipeline.

Paseo Asset Hub advertises `[4, 5]`, but its V5 pipeline gives the current Epoca host no supported place to put a product-account signature. The previous `max(versions)` policy selected V5 and Epoca correctly rejected the transaction as unauthorized. Asset Hub still advertises and accepts V4, whose signed envelope carries the product-account signature.

The first revision attempted to infer host V5 support from a `VerifyMultiSignature` metadata entry. Review found that unsafe: the host encodes the highest transaction-extension pipeline version rather than necessarily version zero, and metadata presence alone does not prove how the host owns or supplies that authorization extension. This revision makes no capability inference. V5 capability negotiation belongs in the host protocol before dual-version runtimes select it here.

## Verification

- `@parity/product-sdk-host`: 19/19 files, 108/108 tests passed
- `@parity/product-sdk-terminal`: 12/12 files, 168/168 tests passed
- host and terminal typechecks passed
- host and terminal builds passed
- Biome and `git diff --check` passed
- live Paseo Asset Hub through Epoca light-client: 5/5 affected flows passed
  - `create-transaction`
  - `sign-batch-payload`
  - `contract-store-value`
  - `contract-store-value-if-person`
  - `contract-withdraw`

Latest live batch transaction: `0x2205febc276eae15c8fea270ad72299358ef90e4671d768375b4c83b6cdfe8ea`

### DIFF
--- a/product-sdk/packages/host/src/accounts.ts
+++ b/product-sdk/packages/host/src/accounts.ts
@@ -342,21 +342,40 @@ export interface AccountsProvider {
 }
 
 /**
- * Derive the host's extrinsic-extension version from SCALE-encoded metadata:
- * v4 → 0, otherwise the latest supported version. `unifyMetadata` normalizes
- * v14/v15 so `.extrinsic.version` is an array.
+ * Map metadata's supported extrinsic formats to the host wire protocol.
  *
- * Indirected through {@link deps} so the SCALE decode (which needs a real
- * metadata blob) can be stubbed in unit tests while the rest of the `signTx`
- * flow — genesis extraction, extension mapping, the host call — is exercised.
+ * V5 General transactions have no envelope signature, so prefer V5 only when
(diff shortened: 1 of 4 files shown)


TarikGul (approved):
Looks solid, thanks

PR #326, opened 2026-08-24
ci(docs): fix the Bulletin deploy domain

Imod7:
### Description

The docs Bulletin deploy fails on every release day. It passes a hardcoded `product-sdk.dot`, and `bulletin-deploy` now takes the suffix from the target environment, which uses `.paseo` names. Every step is gated on an umbrella tag at HEAD, so ordinary runs skip everything and report success, which hid it until `0.23.0`.

### Changes

- Pass the bare label `product-sdk` with `--env paseo-next-v2`, so the tool resolves the suffix and a moved default fails instead of publishing somewhere else.
- Pin the install to `bulletin-deploy@0.15.2`. Nothing bumps that automatically, which the comment records.
- Add a `bulletin-domain` job to `docs-verify.yml` that rejects a literal suffix or an install without a version, and add `docs-deploy.yml` to the path filters.

The published name changes from `product-sdk.dot` to `product-sdk.paseo`, at `product-sdk.paseo.li`. Nothing in the repo links to the old one. The first deploy registers the new name and pays its storage deposit.

No changeset. `RELEASES.md` exempts CI changes under `.github/`.

### Testing

The guard is the regression test: exit 1 on `main`, 0 here. The first commit adds it before the fix, so it is red on its own.

```bash
grep -nE 'bulletin-deploy .*\.(dot|paseo|test)\b' .github/workflows/docs-deploy.yml
grep -nE 'npm (install|i) (-g|--global) [^[:space:]]+' .github/workflows/docs-deploy.yml | grep -vE '@[0-9]'
```

Both are silent here. On `main` they print lines 119 and 111. The domain was checked against the tool's own `parseDomainName` from `0.15.2`: `product-sdk.dot` throws the error CI reported, `product-sdk` resolves to `product-sdk.paseo`.

### DIFF
--- a/.github/workflows/docs-deploy.yml
+++ b/.github/workflows/docs-deploy.yml
@@ -116,7 +116,7 @@ jobs:
               with:
                   timeout_minutes: 10
                   max_attempts: 3
-                  command: cd docs && bulletin-deploy ./out product-sdk.dot --js-merkle
+                  command: cd docs && bulletin-deploy ./out product-sdk --js-merkle
 
             - name: Rebuild static site for GitHub Pages
               if: steps.gate.outputs.skip != 'true'
--- a/.github/workflows/docs-deploy.yml
+++ b/.github/workflows/docs-deploy.yml
@@ -108,7 +108,8 @@ jobs:
 
             - name: Install bulletin-deploy
               if: steps.gate.outputs.skip != 'true'
-              run: npm install -g bulletin-deploy
+              # Pinned: an upstream change would otherwise surface only on a release day.
+              run: npm install -g bulletin-deploy@0.15.2
 
             - name: Deploy to Bulletin Chain
               if: steps.gate.outputs.skip != 'true'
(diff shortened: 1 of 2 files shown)


.github/workflows/docs-deploy.yml
```diff
@@ -108,15 +108,17 @@ jobs:
 
             - name: Install bulletin-deploy
               if: steps.gate.outputs.skip != 'true'
-              run: npm install -g bulletin-deploy
+              # Pinned: an upstream change would otherwise surface only on a release day.
+              # No automation bumps this, so check it if a release fails here.
+              run: npm install -g bulletin-deploy@0.15.2
```
TarikGul, line 113: Hmm this seems like a problem in the future since its hardcoded.

TarikGul: Bit of a tradeoff here I guess.

Imod7: Yes indeed. I can't bump this automatically but at least the comment is there and now when it goes stale it will fail clearly/loudly. I could still add the caret and switch to `^0.15.2` but after checking with Claude it looks like it doesn't help much.

TarikGul (approved):

PR #328, opened 2026-08-25
test(e2e): add retry diagnostics reporter to surface masked regressions

TarikGul:
The demo e2e suites run against a live public chain, so retries:1 is the right call for genuine infra flakiness (RPC blip, slow block, dropped websocket). But Playwright's retry is unconditional: it retries an assertion mismatch the same as a timeout, so a test that fails a real assertion on attempt 1 and passes on attempt 2 (nondeterministic chain state, ordering races) turns green with no trace — a regression hidden behind a retry.

Add a shared reporter (examples/e2e-retry-reporter.ts) that observes each first-attempt failure, classifies it as INFRA (timeout/connection/RPC) or ASSERTION (an expect mismatch), and prints an end-of-run summary. It specifically flags any test that first failed an ASSERTION and then went GREEN on retry — the case most likely to be masking a real bug. Retry behaviour is unchanged; this only adds visibility.

Wired into all 9 demo playwright.config.ts via a relative reporter path. No changeset: examples/ is not published.

### DIFF
--- a/product-sdk/examples/e2e-retry-reporter.ts
+++ b/product-sdk/examples/e2e-retry-reporter.ts
@@ -30,7 +30,17 @@ import type {
     TestResult,
 } from "@playwright/test/reporter";
 
-type FailureKind = "infra" | "assertion";
+/**
+ * - `infra`     — a transport/timeout failure with no assertion involved.
+ * - `wait`      — a web-first assertion (`toHaveText`, `toContainText`, …) that
+ *                 exhausted its own timeout while polling. It reads like an
+ *                 assertion (`expect(` is in the message) but it failed on
+ *                 *time*, not on a synchronous mismatch, so a retry is
+ *                 legitimate — this is the shape of a slow chain connection.
+ * - `assertion` — a synchronous `expect` mismatch. The one a retry must not
+ *                 mask, because it means the value was genuinely wrong.
+ */
+type FailureKind = "infra" | "wait" | "assertion";
 
 interface RetriedAttempt {
     /** The test itself, so its final `outcome()` can be read in `onEnd`. */
@@ -43,11 +53,15 @@ interface RetriedAttempt {
 }
 
 /**
- * Best-effort classification of a failed attempt. A `timedOut` status, or an
- * error message mentioning the transport / chain, is infra. Anything else —
- * most importantly an `expect(...)` mismatch — is treated as a real assertion
- * failure, which is the conservative choice: we would rather over-report a
- * masked regression than let one slip through as "just flaky".
+ * Best-effort classification of a failed attempt.
+ *
+ * The load-bearing distinction is between a *synchronous* `expect` mismatch (a
+ * real regression — `assertion`) and a *web-first* assertion that timed out
+ * while polling (a slow-connection flake — `wait`). Both carry `expect(` in the
+ * message, so `expect(` alone can't tell them apart. The discriminator is
+ * Playwright's `Call log:` section: it is present only for a polled/awaited
+ * assertion that exhausted its timeout, never for a synchronous mismatch.
+ * Verified across the timeout-vs-mismatch shapes.
  */
 function classify(result: TestResult, error: TestError | undefined): FailureKind {
     if (result.status === "timedOut") return "infra";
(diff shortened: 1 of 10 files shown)


product-sdk/examples/e2e-retry-reporter.ts
```diff
@@ -0,0 +1,172 @@
...
+        "expect(",
+        "\nexpected:",
+        "\nreceived:",
+        "tocontaintext",
+        "tohavetext",
+        "tobe(",
+        "toequal(",
+        "tohavelength",
+    ];
+    if (ASSERTION_SIGNALS.some((s) => haystack.includes(s))) return "assertion";
```
Imod7: Every web-first assertion (`toHaveText`, `toContainText`) carries `expect(` in its message, including one that failed only on time, so this line claims it before `INFRA_SIGNALS` runs. The last failing e2e run on `main` is exactly that: `Received: "connecting"`, `Timeout: 90000ms`, 182 polls, green on retry. It would have printed the masked-regression warning for a slow connection. The specs have 147 web-first assertions against 58 plain ones, so it is the common case.

An assertion that exhausted its timeout carries a `Call log:` section; a synchronous mismatch does not. Please add a third kind before this check:

```ts
if (haystack.includes("call log:")) {
    return haystack.includes("expect(") ? "wait" : "infra";
}
```

with its own tag, so the warning keeps only real mismatches. Checked by Claude on seven failure shapes. Worth dropping `"rpc"`, `"network"` and `"navigation"` from `INFRA_SIGNALS` too, since they would swallow "Balance decoding failed for network Paseo".

product-sdk/examples/e2e-retry-reporter.ts
```diff
@@ -0,0 +1,172 @@
...
+            console.log(
+                "     GREEN on retry. A retry is meant for infrastructure flakiness, not for",
+            );
+            console.log(
+                "     assertion mismatches — treat each of these as a possible regression the",
+            );
+            console.log("     retry masked, and investigate before trusting the green run.");
+        }
+        console.log("──────────────────────────────────────────────────────────────");
+        console.log(`  Run status: ${result.status}`);
```
Imod7: Nothing happens after this warning prints. Claude ran a suite whose only problem was a real assertion failure rescued by a retry: `1 flaky`, warning printed, exit 0. In CI that paragraph lands in one of nine `##[group]` sections, which GitHub collapses, in a green job.

`onEnd` can return a status that overrides the run result:

```ts
if (maskedAssertions.length > 0 && result.status === "passed") {
    return { status: "failed" };
}
```

The exit code went from 0 to 1 when I tried it. Do it after fixing the classification, though, otherwise it fails the build on every slow connection. Also move this `Run status:` line below the override, since it prints `passed` for a run you are about to fail.

product-sdk/examples/e2e-retry-reporter.ts
```diff
@@ -0,0 +1,172 @@
...
+
+class RetryDiagnosticsReporter implements Reporter {
+    private readonly retried: RetriedAttempt[] = [];
+
+    onTestEnd(test: TestCase, result: TestResult): void {
+        // Playwright reports each attempt separately. The one that matters for
+        // masking is the *first* failure (retry === 0): that's the failure a
+        // later retry could hide. We record it here; whether the retry rescued
+        // it is read from `outcome()` later, in `onEnd`.
+        if (result.retry === 0 && result.status !== "passed" && result.status !== "skipped") {
```
Imod7: This records every first-attempt failure, including ones never retried and ones meant to fail. Locally `retries` is 0, so two plain failures print "2 test(s) needed a retry" with both rows marked "still failed". And a `test.fail()` test, which Playwright counts under `1 passed`, is listed as an ASSERTION failure.

Add `test.outcome() !== "expected"` here, and count the retried ones separately in `onEnd`:

```ts
const retriedCount = this.retried.filter((r) => r.test.results.length > 1).length;
`${this.retried.length} test(s) failed first attempt, ${retriedCount} retried`
```

Claude checked `--max-failures` too; tests that never ran do not reach `onTestEnd`, so no phantom rows.

Imod7 (approved):

filvecchiato (approved):

Issue #318, opened 2026-08-21
Chain-read support for collectibles, credits & collections


piggydoughnut:
Below are our use cases for Scarcity and I am sure other apps can reuse those too in the future.

> [!WARNING]
> Currently the purses are not implemented. Most of the following functionality depends on purses being available.

**Implementable today:**
- `getItemDisplay(instance)` - read by instance. No purse/identity dependency.
- `getVerifiedArtwork(imageRef)` - IPFS fetch + verify; the gateway origin is a config question, not a blocker.
- `previewClaim(credit, collections)` - takes a credit hash + collection ids. 
- `getCollections()` - reads the registry.  (part of https://github.com/paritytech/product-sdk/pull/329)
- `getCollectionItems(id)` - reads a collection's catalogue.  (part of https://github.com/paritytech/product-sdk/pull/329)
- `getCredits`

The rest has to wait for purses etc.

### Get NFTs a user holds.  `getOwnedNfts(identity)` → `OwnedNft`[] 

- storage: `Scarcity.NftsByOwner` (map AccountId → Nft, one per purse)
- scans the purse subtree + resolves metadata for each NFT 

### Get metadata for an NFT or a batch of NFTs.`getItemDisplay(instance)` → `{ hash?, name?, imageUrl?, rarity? }`

- metadata: name + artwork + rarity + hash + others. 
- runtime `ScarcityApi.metadata_batch`
- batched variant `getItemDisplays(instances[])` 

### Fetch Artwork bytes + verification. `getVerifiedArtwork(imageRef)` → Blob | null

- fetch image by CID/digest, verify against on-chain hash.

## Credits

> [!WARNING]
>  Credit reads are scoped by the player's identity. AccountOrPerson: either an account or a person Alias. Needs clarification which one is used for awarding credits. Note: account-era and person-era credits live under separate keys with no on-chain link, so reading a full history may require querying both.

### Get a user's credits to show in the UI, `getCredits(identity)` → `Credit[]`

- All credits awarded to the player, newest first, each carrying its state (earned / claimable / claimed).
- People Chain — awards + proofs:
  storage `NftCredits.NftClaimCreditBlocks`, `NftClaimCreditAwards`;
  runtime `NftCreditsApi.nft_claim_credit_roots`, `nft_claim_credit_proofs`
[...]

PR #329, opened 2026-08-26
feat(nfts): collections

piggydoughnut:
Part of #318.

Adds `@parity/product-sdk-nfts`: catalogue reads of the `Scarcity` pallet on Asset Hub. Every value comes from storage, pinned to one finalized block per call.

## Pagination

`DEFAULT_PAGE_LIMIT` (100) and caps at `MAX_PAGE_LIMIT` (1000) - both exported.

| in | out |
|---|---|
| `limit`, `fromId` | `idCeiling`, `nextId` |

Pagination walks the sequential `u32` index space each read enumerates, which is sound because the runtime guarantees the shape: ids and indices come from counters that only move forward, and both `delete_collection` and `delete_item` document that identifiers are **never reused**.

## Shared options

Every read takes these. `getCollectionItems` adds `attributes`.

| Option | Type | Default | Notes |
|---|---|---|---|
| `limit` | `number` | `100` | Entries this page returns. Max limit is set to `1000`; a larger request is capped by max limit rather than fail, and `nextId` still reports where the page stopped. |
| `fromId` | `number` | `0` | Where the window starts. Take it from the previous page's `nextId`. |
| `at` | `FinalizedSnapshot` | — | Address a block a previous read already pinned, instead of pinning a new one. |
| `signal` | `AbortSignal` | — | Forwarded into every underlying pull.  |

**`at` is what makes a walk coherent.** Without it every call pins its own finalized block, right for unrelated questions, wrong for one question asked in pages, since a walk over its own snapshots is not a walk of any single chain state. Pass another result's `at` straight back in. It also makes two reads agree: the registry and the full listing, or a listing and a catalogue, at one block.

## API functions

Each pins its block first through `raw.assetHub.getFinalizedBlock()` (unless given `at`), then addresses storage at that hash, so all values in one result come from one block. Each returns a `Result`, per the SDK-wide error model.

### **`getCollections(chain, options?)`**
→ `Result<CollectionsResult, ProductNftsError>`

```ts
interface CollectionsResult {
    at: FinalizedSnapshot;
    collections: Collection[];    // { id, name, itemCount, owner, selection }
    idCeiling: number;            // exclusive end of the collection id space
    nextId: number | null;
}
```

Every collection on chain, ascending by id, a page at a time. `selection: null` means the collection exists but accepts no claims.

| Call | Shape | Count |
|---|---|---|
| `raw.assetHub.getFinalizedBlock()` | pin the block, unless `at` was given | 1 |
| `query.Scarcity.NextCollectionId.getValue(at)` | the id ceiling | 1 |
| `query.Scarcity.Collections.getValues(window)` | the window's records | 1 (+1 per stretch of deleted ids) |
| `query.NftClaims.CollectionMinters.getValues(page ids)` | `selection` for the page | 1 |
| `query.Scarcity.CollectionMetadata.getValues(page ids × "name")` | exact-key names | 1 |

### **`getClaimableCollections(chain, options?)`**
→ `Result<ClaimableCollectionsResult, ProductNftsError>`

```ts
interface ClaimableCollectionsResult {
    at: FinalizedSnapshot;
    collections: ClaimableCollection[];  // { id, name, itemCount, owner, selection }
    idCeiling: number;
    nextId: number | null;
}
```

The subset `getCollections` filters: collections registered to accept claims, ascending by id. 

| Call | Shape | Count |
|---|---|---|
| `raw.assetHub.getFinalizedBlock()` | pin the block, unless `at` was given | 1 |
| `query.Scarcity.NextCollectionId.getValue(at)` | the id ceiling | 1 |
| `query.NftClaims.CollectionMinters.getValues(window)` | which ids are registered | 1 (+1 per stretch of unregistered ids) |
| `query.Scarcity.Collections.getValues(page ids)` | the records, batched | 1 |
| `query.Scarcity.CollectionMetadata.getValues(page ids × "name")` | exact-key names | 1 |

### **`getCollectionItems(chain, id, options?)`**
→ `Result<CollectionItemsResult, ProductNftsError>`

```ts
type CollectionItemsResult =
    | { tag: "Found"; at: FinalizedSnapshot; idCeiling: number; nextId: number | null;
        collection: CollectionDetail }   // { id, name, itemCount, items }
    | { tag: "NotFound"; at: FinalizedSnapshot; id: number };
```

One page of a collection's item catalogue, items ascending by index. Applies **no** registry filter, so it reads a collection that accepts no claims just as well. A collection nobody created is not an error; it rides the `ok` channel as `NotFound`.

`idCeiling` counts every item ever defined here (indices are never reused); 
`collection.itemCount` counts the ones still alive. They diverge permanently once anything is deleted.

**Extra option:**

| Option | Type | Default | Notes |
|---|---|---|---|
| `attributes` | `boolean` | `false` | Return the open metadata. |

| Call | Shape | Count |
|---|---|---|
| `raw.assetHub.getFinalizedBlock()` | pin the block, unless `at` was given | 1 |
| `query.Scarcity.Collections.getValue(id, at)` | the record: id ceiling + the `NotFound` test | 1 |
| `query.Scarcity.CollectionMetadata.getEntries(id, at)` | the collection's defaults, to merge under | 1 |
| `query.Scarcity.ItemDefs.getValues(window)` | the window's definitions | 1 (+1 per stretch of deleted indices) |
| `query.Scarcity.ItemMetadata.getValues(page × 3 named keys)` | `name`, `image`, `rarity` — when `attributes` is off | 1 |
| `query.Scarcity.ItemMetadata.getEntries(id, at)` | every key of every item — when `attributes` is on | 1 |

**`attributes` is `Record<string, string> | null`, and `null` means "not fetched".** An empty object would claim the item carries no metadata, which is a different statement. The typed fields (`name`, `image`, `rarity`) are keys this package can name, so a page fetches them for its whole window with one exact-key read; the bag's keys are open by definition, so there is nothing to ask for by name and filling it means scanning the collection. Collection-level defaults are inherited either way.

## Other changes

`packages/nfts/src/chain.ts` — the client the reads take, typed structurally by the **six** storage entries they touch rather than by naming a descriptor, so no genesis hash is pinned to read a catalogue. It asks for exactly the accessors the reads use: making every read paged removed four of them. An app pruning its own descriptors must whitelist all six, `Scarcity.NextCollectionId` included.

`packages/nfts/src/paging.ts` — `DEFAULT_PAGE_LIMIT`, `MAX_PAGE_LIMIT`, `pageBounds`, and `fillByIdWindow`, shared by all three reads. They differ only in what makes an index interesting (a collection record, a minter entry, an item definition), so the density widening, scan budget and cursor semantics live in one place.

`packages/nfts/src/metadata.ts` — the decode convention. Metadata is untyped `Vec<u8>` → `Vec<u8>` in three layers, each overriding the last; a catalogue read merges the first two. Values decode as UTF-8 when the bytes are readable text and `0x`-hex otherwise. Exact-key reads take a plain `Uint8Array` — PAPI 2.x generates `[number, Uint8Array]` for this `Vec<u8>` key.

`packages/nfts/src/errors.ts` — `NftsChainEntryError` names the storage entry a read could not reach, carries it on `entry`, and keeps PAPI's error as the `cause`.

`packages/sdk/*` — the `@parity/product-sdk/nfts` subpath, plus `src/nfts/contract.test.ts`: compile-time assertions that a real `getChainAPI` client satisfies `NftsChain`, with devnet Asset Hub as a negative control. These run under `pnpm typecheck`, not vitest.

`examples/nfts-demo/*` — demo app plus Playwright specs. The panel for `getCollections` renders the ids with `selection: null` next to the registry's, because the gap between the two lists is the thing worth seeing on live data. It also walks the id space in pages of 2 with `at` pinned, and the spec asserts two things a type cannot: that names from a small-page walk match a single larger page (parameter bivariance means the contract check would accept a wrong exact-key type, and a wrong key silently returns `null` for every name), and that the whole walk touched exactly one block.

`skills/product-sdk-nfts/SKILL.md`, `CLAUDE.md`, `.claude-plugin/marketplace.json`, `README.md`, `.changeset/config.json` — the skill and its registration, the new package's row in the package table, and `@parity/product-sdk-nfts-demo` added to the changeset ignore list alongside the other demos. Plus one changeset.

## Notes

Nothing on chain declares the metadata keys or the value types, so `name`, `image` and `rarity` are a convention this package applies, not a schema it enforces; every other key is reachable through `attributes`. Unconfirmed with the pallet team.

`imageRef` reports the same bytes as `hex` and as `text` (`null` when unreadable): one deployment stores a 32-byte content digest there, another an ASCII CID, and nothing on chain says which, so the caller picks.

`transferability` is **not** returned. It traces to `pallet_nfts`' `CollectionSetting::TransferableItems` and has no source in `Scarcity`, not in `ItemDefs`, not in any metadata key the live chain carries.


product-sdk/packages/nfts/src/items.ts
```diff
@@ -0,0 +1,969 @@
...
+    try {
+        const { signal } = options;
+        const snapshot = await pinBlock(chain, signal, options.at);
+        const at = readAt(snapshot, signal);
+        const query = chain.assetHub.query;
+        const { limit, fromId } = pageBounds(options);
+
+        // The record carries the index ceiling, so unlike the collection listing
+        // this read pays nothing extra to learn where the space ends.
+        const record = query.Scarcity.Collections.getValue(id, at);
```
Imod7, line 186: Please validate `id` before it becomes a storage key. PAPI's `u32` encoder truncates rather than rejecting, so `getCollectionItems(chain, NaN)` returns collection 0's catalogue labelled `id: NaN` and `1.5` returns collection 1's, both on the `ok` channel as `Found`. `Number.isInteger(id) && id >= 0 && id < 2 ** 32`, returning `err`, closes it. `fromId` in `packages/nfts/src/paging.ts` line 93 needs the same check, so the predicate belongs there.

product-sdk/packages/nfts/src/paging.ts
```diff
@@ -0,0 +1,243 @@
...
+ */
+export function pageBounds(options: { limit?: number; fromId?: number }): {
+    limit: number;
+    fromId: number;
+} {
+    const asked = usable(options.limit) ?? DEFAULT_PAGE_LIMIT;
+    const from = usable(options.fromId) ?? 0;
+    return {
+        limit: Math.min(Math.max(0, Math.trunc(asked)), MAX_PAGE_LIMIT),
+        fromId: Math.max(0, Math.trunc(from)),
```
Imod7, line 93: `fromId` gets a floor but no ceiling, and the id space is `u32`. PAPI's encoder truncates, so `getCollections(chain, { fromId: 2 ** 32 })` on a two-collection chain returns those two as ids 4294967296 and 4294967297, with `nextId: null`.

product-sdk/packages/nfts/src/paging.ts
```diff
@@ -0,0 +1,243 @@
...
+    let cursor = Math.min(consume(firstIds, firstProbe, fromId + limit), Math.max(idCeiling, 0));
+
+    // Only while short. On a dense id space this never runs.
+    while (kept.length < limit && cursor < scanCeiling) {
+        const want = limit - kept.length;
+        const scanned = cursor - fromId;
+        // Widen by the density seen so far, so a sparse range converges in a
+        // couple of reads rather than one read per gap.
+        const density = Math.max(kept.length / Math.max(scanned, 1), 1 / SCAN_BUDGET_FACTOR);
+        const width = Math.min(Math.ceil(want / density), want * SCAN_BUDGET_FACTOR);
```
Imod7, line 156: This width is capped by the scan budget but not by `MAX_PAGE_LIMIT`, so one probe can ask for 15,000 keys at `limit: 1000` on a sparse space. Since `getValues` is `Promise.all(keys.map(getValue))`, that is 15,000 concurrent operations against a documented bound of 1,000. 
`MAX_PAGE_LIMIT`'s own doc needs rewording too, since it names one bound where there are three.

product-sdk/packages/nfts/src/items.ts
```diff
@@ -0,0 +1,969 @@
...
+        }
+
+        const defaultBag = collectionBag(defaultRows);
+        const decodedDefaults = decodeBag(defaultBag);
+        const indices = filled.kept.map(({ id: index }) => index);
+
+        // The one branch: named keys for the page, or the whole collection's item
+        // metadata when the open bag was asked for.
+        const overrides = options.attributes
+            ? await readAllKeys(query, id, at)
```
Imod7, line 219: `readAllKeys` runs even when the page found no items, so a page past the end still prefix-scans the whole collection to return nothing. `readTypedKeys` on the next line has that guard and this branch does not, so please skip the metadata read when `filled.kept` is empty. Separately, `attributes` is a per-page option with a per-collection cost, so a walk repeats the full scan every page. A `getCollectionAttributes(chain, id, { at })` read would let a caller pay it once.

product-sdk/packages/nfts/src/collections.ts
```diff
@@ -0,0 +1,1481 @@
...
+ *
+ * let page = first.value;
+ * const at = page.at;
+ * for (;;) {
+ *     for (const c of page.collections) {
+ *         console.log(c.id, c.name ?? "(unnamed)", c.selection ? "claimable" : "not claimable");
+ *     }
+ *     if (page.nextId === null) break;
+ *     const next = await getCollections(chain, { limit: 100, fromId: page.nextId, at });
+ *     if (!next.ok) break;
```
Imod7, line 486: This treats a failed page as the end of the walk, so a caller copying it renders a partial list as complete. PAPI holds a pinned block for one storage read only, so a long `at`-pinned walk will lose it. Please report the failure rather than break, and say in the `at` doc that a walk has to re-pin. Four more sites: `items.ts` line 167, `SKILL.md` lines 98 and 154, and the changeset line 64.

product-sdk/packages/nfts/src/index.ts
```diff
@@ -0,0 +1,171 @@
...
+export type {
+    CollectionDetail,
+    CollectionItem,
+    CollectionItemsResult,
+    FinalizedSnapshot,
+    ImageRef,
+    ItemSelection,
+    ClaimableCollection,
+    Collection,
+    RawBytes,
```
Imod7, line 169: Four types in `NftsChain`'s signature are not exported: `RawCollection`, `RawItemDef`, `RawMinter` and `RawMetadataEntry`. Importing them fails, and `RawCollection` gets the suggestion "Did you mean 'Collection'?", which is a different type and would give a wrong implementation. Adding them here is additive and clears four typedoc warnings.

product-sdk/packages/nfts/src/metadata.ts
```diff
@@ -0,0 +1,314 @@
...
+
+/**
+ * Is this decoded string safe to hand a UI as text?
+ *
+ * `TextDecoder` with `fatal` already rejects invalid UTF-8, but valid UTF-8 is
+ * not the same as readable: a digest can decode cleanly and still be control
+ * characters. Anything below U+0020 that is not tab/newline/carriage-return
+ * disqualifies it, as does a lone U+FFFD.
+ */
+function isText(decoded: string): boolean {
```
Imod7, line 91: The doc above asks whether the string is safe for a UI, but the check covers only C0 controls, U+007F and U+FFFD. C1 controls, bidi overrides and zero-width characters pass, so a name containing U+202E renders as a different name. Metadata is author-supplied, so please either narrow the doc or widen the check.

product-sdk/packages/nfts/src/collections.ts
```diff
@@ -0,0 +1,1481 @@
...
+ * records, registry entries and `name` rows — with nothing proportional to the
+ * chain anywhere in the read.
+ *
+ * **A page comes back full.** Ids of deleted collections are holes — no record,
+ * and the runtime requires the metadata gone before deletion — so the read walks
+ * past them until it has `limit` collections rather than handing back a short
+ * page. That costs an extra record read where holes appear and nothing else: a
+ * hole never gets a name or registry lookup. A page is short only at the end of
+ * the id space, or when a mostly-deleted range hits
+ * {@link SCAN_BUDGET_FACTOR}; `nextId === null` is the only end signal.
```
Imod7, line 464: `SCAN_BUDGET_FACTOR` is not exported, so this link does not resolve and `docs:extract` warns. Exporting it beside `DEFAULT_PAGE_LIMIT` and `MAX_PAGE_LIMIT` fixes the link and gives callers the second half of the paging bound.

product-sdk/packages/nfts/src/collections.ts
```diff
@@ -0,0 +1,1481 @@
...
+): Promise<Result<ClaimableCollectionsResult, ProductNftsError>> {
+    try {
+        const { signal } = options;
+        const snapshot = await pinBlock(chain, signal, options.at);
+        const at = readAt(snapshot, signal);
+        const query = chain.assetHub.query;
+
+        const { limit, fromId } = pageBounds(options);
+
+        if (limit === 0) {
```
Imod7, line 262: `fillByIdWindow` already handles `limit: 0` and returns `nextId: null`, so this shortcut and its twin in `readPage` duplicate the helper. `getCollectionItems` has none and is correct without one. Dropping both costs one `getValues([])`, which does no network work.

product-sdk/packages/nfts/src/collections.ts
```diff
@@ -0,0 +1,1481 @@
...
+        // should cost no further round trips.
+        const registered = registry.map(({ id, minter }) => ({
+            id,
+            selection: toItemSelection(minter.selection),
+        }));
+        const ids = registered.map(({ id }) => id);
+
+        // Both concurrently, for exactly the ids being returned — `chain.ts` has
+        // what a multi-key read costs.
+        const [records, names] = await Promise.all([
```
Imod7, line 293: `readPage` returns early when its window found nothing and this read does not, so an empty registry page still calls `Collections.getValues([])`. The same early return keeps the two in step. Separately, the three option interfaces each redeclare `limit`, `fromId`, `at` and `signal`; a shared `PagedReadOptions` would document them once.

Imod7 (commented):
The paging design is right, and resting it on the never-reused id space is the right reason. Three things worth settling before merge, each with the detail on its own line: unchecked ids, at `packages/nfts/src/paging.ts` line 93 and `packages/nfts/src/items.ts` line 186; and the unknown `ItemSelection` variant at `packages/nfts/src/collections.ts` line 208. One shape recommendation, on the `readAllKeys` line in `packages/nfts/src/items.ts`: keep `attributes` as an option and give it a named-key form.

The design fits the repo. The read signature, the `SdkError` model, the structural chain contract and the compile-time check in `packages/sdk/src/nfts/contract.test.ts` all match what `@parity/product-sdk-individuality` established, and the dependency set is leaner than the sibling's. Two things I expected to be divergences are not, the positional `id` argument and the `get*` naming. Two consistency points remain, neither blocking, with the detail on each line: `packages/nfts/src/index.ts` on a missing `fromPapi`, and `packages/nfts/src/chain.ts` on how wide `NftsChain` is.

BigTava:
jollity-next wants this package, and today it duplicates part of 318: `getCredits` exists twice already, as `packages/game-chain/src/chain/attestations.ts` here and `src/chain/pallets/credits.ts` in scarcity-stash, and our artwork loader skips the verification `getVerifiedArtwork` would give us. Worse, jollity-next still derives the collectable from the credit hash against a catalogue in the repo, which `scarcity-stash/docs/READ_PATH.md` recorded as wrong, since a claim mints with empty metadata and replaying the selection fails silently as a collection grows. Please land `getCredits`, `previewClaim` and `getVerifiedArtwork` beside the collection reads so both products call one implementation, and we will be the second consumer against paseo-next-v2.

product-sdk/packages/nfts/src/items.ts
BigTava: Fixed. getCollectionItems now refuses an id that is not a u32 with a new NftsIdError. Thanks.

product-sdk/packages/nfts/src/paging.ts
BigTava: Fixed. Thanks.

product-sdk/packages/nfts/src/paging.ts
BigTava: Fixed. Thanks.

product-sdk/packages/nfts/src/items.ts
BigTava: Fixed. Thanks.

product-sdk/packages/nfts/src/index.ts
BigTava: Fixed. Thanks.

product-sdk/packages/nfts/src/collections.ts
BigTava: Fixed. Thanks.

product-sdk/packages/nfts/src/collections.ts
BigTava: Fixed. Thanks.

product-sdk/packages/nfts/src/collections.ts
BigTava: Fixed, except limit and fromId stay per read because their docs differ on purpose. Thanks.

product-sdk/packages/nfts/src/collections.ts
BigTava: Fixed. Thanks.

product-sdk/packages/nfts/src/metadata.ts
BigTava: Widened to C1 and bidi controls, left the zero-width joiner in because emoji sequences need it. Thanks.

PR #330, opened 2026-08-26
fix(host): surface host response-decode failures as a typed error (#270)

### LINKED ISSUE #270: RangeError: Offset is outside the bounds of the DataView
#### Reproduction steps

1. App on product-sdk 0.19.0, opened inside Polkadot Desktop 0.6.24.
2. `const p = await getAccountsProvider(); await p.createRingVRFProof({ productId, suffix: "0x" }, { chainId, junctions: [] }, message).match(...)`.

#### Environment

- `@parity/product-sdk` 0.19.0
- `@parity/product-sdk-host` 0.14.0
- `@parity/truapi` 0.5.0
- Polkadot Desktop 0.6.24 (reported "up to date")

#### Current behaviour

```
RangeError: Offset is outside the bounds of the DataView
    at DataView.prototype.getUint8
    at ... scale-ts.js
    at ... scale.js
    at decodeResponse (client.js:153)
    at Object.resolve (client.js:255)
    at Object.deliver (transport.js:139)
    at p.onmessage (transport.js:262)
```

This worked on `0.18.x` and breaks after upgrading to `0.19.0`. No other host calls (statements, storage, `getProductAccount`, `getUserId`, `getProductAccountAlias`) regressed.

TarikGul:
Closes #270

## Problem

`getAccountsProvider()` calls could crash with an opaque `RangeError: Offset is outside the bounds of the DataView`, a stack naming neither the call nor the cause (reported on `createRingVRFProof`). It happens when the host and the product's `@parity/truapi` client are on different protocol versions, so a host reply is a frame the client's SCALE codec can't decode. The truapi client wraps each call with `ResultAsync.fromSafePromise`, which assumes its promise never rejects, but the decode runs inside that promise and can throw, so the failure escapes the `Result` channel entirely as an unhandled throw from the transport's message handler.

## Fix

A `guardDecode` boundary in the accounts adapter routes that thrown or rejected decode onto the `Result` err channel as a new `HostResponseDecodeError`, which names the failing call and preserves the original error as `cause`. Well formed responses and each call's own typed `Err` pass through untouched. Applied to all eight account lookup methods.

New exports: the `HostResponseDecodeError` class (extends `HostError`, so `isHostError` and `instanceof HostError` catch it) and the `WithDecodeError<E>` type alias. Every `AccountsProvider` lookup method's err type is widened to `WithDecodeError<...>`, so consumers matching on the err channel gain one additional case.


product-sdk/packages/host/src/errors.ts
```diff
@@ -144,6 +144,33 @@ export class HostCallFailedError extends HostError {
...
+ * that names neither the call nor the cause. This names the call and preserves
+ * the original error as {@link cause}, so a bug report has somewhere to start.
+ */
+export class HostResponseDecodeError extends HostError {
+    /** The host-API call whose response failed to decode, e.g. `"createRingVRFProof"`. */
+    readonly call: string;
+
+    constructor(call: string, cause: unknown) {
+        super(
+            `Failed to decode the host's response to ${call}. The host app and the @parity/truapi version this product is built against are likely on different protocol versions.`,
```
Imod7: This asserts a cause the code cannot know, and omits the one it has. `guardDecode` fires for every rejection: a closed channel and an adapter mapping bug both produce this exact text, with the real error left on `cause`. Please interpolate it and soften the claim:

```ts
`Could not process the host's response to ${call}: ${formatHostError(cause)}. The usual cause is a protocol-version skew between the host app and the @parity/truapi version this product is built against; a host channel that closed mid-call looks the same.`
```

`formatHostError` is already in this file, and this is how `HostCallFailedError` builds its message. The `toContain("protocol versions")` assertion below needs the new wording.

product-sdk/packages/host/src/accounts.ts
```diff
@@ -231,20 +232,37 @@ export type VrfSignature = { [K in keyof WireVrfSignature]: Uint8Array };
...
+ * cannot be decoded at all (a host/client protocol-version skew) — see
+ * {@link WithDecodeError}.
  */
+/**
+ * A call's declared `Err` channel, plus {@link HostResponseDecodeError}: any
+ * host reply can fail to decode if the host and the product's `@parity/truapi`
+ * client are on different protocol versions, so every decoded call can surface
+ * it in addition to its own typed errors.
+ */
+export type WithDecodeError<E> = E | HostResponseDecodeError;
```
Imod7: This went in between the `AccountsProvider` doc comment and the interface, so the interface has no doc comment now: the last comment before a declaration is the one that attaches. Confirmed on built output, where both the compiler and TypeDoc report no description for `AccountsProvider`, so the API reference page loses it. Please move `WithDecodeError` and its comment above the `Accounts provider handle` block. CI will not catch this either way.

product-sdk/packages/host/src/accounts.ts
```diff
@@ -392,105 +422,157 @@ function toWireProductAccountId({
...
+ * transport's message handler. Wrapping the call re-homes that throw as a
+ * typed {@link HostResponseDecodeError} that names the call, while ok values
+ * and the call's own typed `Err` values pass through untouched.
+ */
+function guardDecode<T, E>(
+    call: string,
+    result: ResultAsync<T, E>,
+): ResultAsync<T, E | HostResponseDecodeError> {
+    return ResultAsync.fromPromise(
+        Promise.resolve(result),
```
Imod7, line 450: `Promise.resolve` is load-bearing here, for a reason worth removing. `result` is already a `PromiseLike`, so `ResultAsync.fromPromise(result, ...)` would do, but the `okMatch` fake below is a hand-rolled object with no `then` that survives only because `Promise.resolve` passes it through. That also means the ten older adapter tests never reach the guard's real path. Making the fake `okAsync(value)` (already imported) fixes both, and all 20 tests pass. If you keep the fake, please say in a comment that `Promise.resolve` is what tolerates it.

product-sdk/pending-changesets/host-response-decode-boundary.md
```diff
@@ -0,0 +1,12 @@
+---
+"@parity/product-sdk-host": minor
+"@parity/product-sdk": minor
+---
+
+**Surface a clear error when a host reply can't be decoded, instead of an opaque `RangeError`.**
+
+When the host app and the `@parity/truapi` version a product is built against are on different protocol versions, a host call can return a frame the client's SCALE codec can't decode. The truapi client builds each call's `ResultAsync` with `fromSafePromise`, which assumes the underlying promise never rejects — but the decode runs inside that promise and *can* throw, so the failure escaped the `Result` channel as an unhandled `RangeError: Offset is outside the bounds of the DataView` from inside the transport's message handler, with a stack that named neither the call nor the cause (reported for `createRingVRFProof`).
```
Imod7: The last clause is not what `@parity/truapi@0.10.0` does. Its `client.js` resolves inside `try { p.resolve(payload.value) } catch (error) { p.reject(toError(error)) }`, so the throw is caught and becomes a rejection; it never leaves the message handler. What escapes is that rejection, because `fromSafePromise` installs no rejection handler. Since this text becomes the published changelog, could you reword it? The same sentence is in the `guardDecode` comment and in the `HostResponseDecodeError` class doc. Also, the PR description says eight lookup methods; the code guards ten.

Imod7 (approved):
The same hole is open outside the accounts adapter. In `product-sdk/packages/host/src/truapi.ts`, line 72:

```
    return result.match(
```

`match` on a rejected `ResultAsync` returns a rejected promise, so `mapHostResult` misses exactly what `guardDecode` catches, and every flat public operation folds through it: `requestPermission` rejects with a raw `RangeError` instead of returning `err(...)`, against its own signature. Adding `.catch((cause) => err(new HostResponseDecodeError(label, cause)))` fixes all of them, tests stay green. `unwrapHostResult` on line 52 has the same shape and covers the signer factories.

PR #331, opened 2026-08-26
fix(terminal): suppress benign DestroyedError teardown noise in destroy()

TarikGul:
## Problem

`adapter.destroy()` logs `Statement subscription error: DestroyedError: Client destroyed` to `console.error`, pure noise that every consumer (bulletin-deploy, playground-cli, d3pot) hand-rolls the same `/DestroyedError|Client destroyed/` filter to hide.

The line comes from an unconditional `console.error` inside `@novasamatech/statement-store`'s rpc adapter, fired from a detached rxjs finalizer when disconnect destroys the client while a statement subscription observable is still live, often after `destroy()`'s Promise has already resolved. A prior refactor (#65) replaced the old suppression with a "drain the tracked unsubscribe RPCs then disconnect" approach, but draining those RPCs does not cover that observable's error emission, so the line still escapes and there is now no suppression at all.

## Fix

`destroy()` runs the disconnect, plus a short tail past it, under a scoped `console.error` filter that drops only the benign teardown line and restores itself afterward. Every other `console.error` passes through untouched, so a genuine error during teardown is still visible. This restores and narrows the suppression #65 removed.

Also exports `isBenignTeardownError(error: unknown): boolean`, the predicate `destroy()` uses, so consumers wiring their own subscription `onError` handlers can drop the line without reinventing the regex.


product-sdk/packages/terminal/src/adapter.ts
```diff
@@ -225,13 +278,22 @@ async function teardown(
...
+    const restore = suppressBenignTeardownErrors();
     try {
         lazyClient.disconnect();
     } catch (e) {
         log.warn("lazyClient.disconnect threw during destroy", { error: e });
     }
+    // The finalizer that logs the benign error can fire a few microtasks after
+    // `disconnect()` returns, so keep the suppression up for a short tail, then
+    // restore. Unref so the timer never keeps a CLI process alive.
+    const timer = setTimeout(restore, TEARDOWN_SUPPRESS_TAIL_MS);
```
Imod7: Two overlapping destroys leak this patch permanently: the second wrapper installs over the first, and once both timers fire the real `console.error` is unreachable. Any spy installed inside the window does the same. The line is emitted synchronously inside `disconnect()`, so `finally { restore(); }` fixes it. Please add a test for overlapping destroys.

product-sdk/packages/terminal/src/adapter.ts
```diff
@@ -202,8 +205,52 @@ function wrapLazyClient(inner: LazyClient): TrackedLazyClient {
...
+ * was destroyed underneath it (`DestroyedError: Client destroyed`).
+ *
+ * Exported so consumers wiring their own subscription `onError` handlers, or
+ * their own `console.error` guards, can drop this one line without reinventing
+ * the regex. {@link createTerminalAdapter}'s own `destroy()` already suppresses
+ * it (see {@link suppressBenignTeardownErrors}); this is for code paths outside
+ * that window.
+ */
+export function isBenignTeardownError(error: unknown): boolean {
+    const text = error instanceof Error ? `${error.name}: ${error.message}` : String(error);
```
Imod7: This also matches `@parity/product-sdk-signer`'s own `DestroyedError`, a direct dependency, so a real signer error gets dropped, and the exported predicate carries that to consumers with no time limit. Only the message is unique:

```ts
const message =
    error instanceof Error ? error.message : typeof error === "string" ? error : "";
return /Client destroyed/.test(message);
```

Your existing assertions still pass. This also fixes the throw flagged on the `args.some` line.

product-sdk/packages/terminal/src/adapter.ts
```diff
@@ -202,8 +205,52 @@ function wrapLazyClient(inner: LazyClient): TrackedLazyClient {
...
+ * returning a restore function.
+ *
+ * The noise comes from an unconditional `console.error('Statement subscription
+ * error:', …)` inside `@novasamatech/statement-store`'s rpc adapter, fired from
+ * a detached rxjs finalizer when `disconnect()` destroys the client while a
+ * subscription is still live. Draining the tracked unsubscribe RPCs does not
+ * cover that observable's error emission, so the ordering-only approach still
+ * lets one line through — hence a scoped suppression.
+ *
+ * Deliberately narrow: only calls whose first argument matches the benign
```
Imod7: This says only the first argument is checked, but the code is `args.some(...)` over all of them, and has to be: the upstream call puts a plain string first, the error second. Matching the code to this comment would disable the feature. Also, "detached finalizer, after this function resolves" is wrong, it is synchronous. Same wording in the `teardown` comment, README and changeset.

product-sdk/packages/terminal/src/adapter.ts
```diff
@@ -78,11 +78,14 @@ export type TerminalAdapter = PappAdapter & {
...
      * `Promise.allSettled` before destroying the substrate-client request
-     * queue. No timing-based guesses; no global-state mutations of
-     * `console.error` or `process.on('unhandledRejection')`. Pending
-     * subscribes (where `onSuccess` hasn't fired yet) are cancelled by
-     * the underlying `getSubscribeFn` teardown via `cancelRequest()`,
-     * which is the in-band fast-path and doesn't surface as a rejection.
+     * queue. Draining those RPCs is not enough on its own: destroying the
+     * client while a statement-subscription observable is still live makes
+     * `@novasamatech/statement-store` log `Statement subscription error:
+     * DestroyedError` to `console.error` from a detached finalizer, often
```
Imod7: Worth noting which half this closes. With a subscription still established at `destroy()`, the same teardown raises an unhandled `Error: Not connected` and exits Node with code 1. `sessions.dispose()` misses it, because the auth subscription is released only by `abort()`. Not for this PR, but could you note the limit here and open a follow-up.

product-sdk/packages/terminal/src/adapter.ts
```diff
@@ -202,8 +205,52 @@ function wrapLazyClient(inner: LazyClient): TrackedLazyClient {
...
+ * lets one line through — hence a scoped suppression.
+ *
+ * Deliberately narrow: only calls whose first argument matches the benign
+ * pattern are dropped; every other `console.error` passes through untouched, so
+ * a genuine error during teardown is still visible.
+ */
+function suppressBenignTeardownErrors(): () => void {
+    const original = console.error;
+    const wrapper = (...args: unknown[]) => {
+        if (args.some(isBenignTeardownError)) return;
```
Imod7, line 257: `String()` on every argument means the wrapper throws where the real `console.error` prints fine, a null-prototype object for example, and `console.error` is usually called from a `catch`. The predicate rewrite suggested on the `const text = ...` line fixes this too. Worth a test that `console.error(Object.create(null))` does not throw.

product-sdk/packages/terminal/README.md
```diff
@@ -71,7 +71,7 @@ Creates a terminal adapter backed by the host-papp SDK.
 - `appId` -- the value you passed in (re-exposed so `createSessionSigner` can pull the productId from the adapter)
 - `sso` -- auth component (`.authenticate()`, `.abortAuthentication()`, status subscriptions)
 - `sessions` -- session manager (signing, disconnect)
-- `destroy()` -- disconnect the WebSocket and release resources. Idempotent. Suppresses `@novasamatech/statement-store`'s noisy `Statement subscription error` log for ~50 ms after the call.
+- `destroy()` -- disconnect the WebSocket and release resources. Idempotent. Drops `@novasamatech/statement-store`'s benign `Statement subscription error: DestroyedError` line for ~50 ms after the call — only that line; every other `console.error` passes through. The predicate it uses, `isBenignTeardownError`, is exported for consumers wiring their own subscription `onError` handlers.
```
Imod7: Two doc things. Every other exported function has its own `###` entry under `## API`; this one is only a clause here. And "consumers wiring their own subscription `onError` handlers" is not reachable: `PappAdapter` exposes only `sso`, `sessions`, `secrets`, `identity` and `allowance`. Their own `console.error` guard, the other half of that sentence, is the real reason. Same wording in the doc comment and changeset.

Imod7 (approved):

PR #332, opened 2026-08-27
fix(signer): preserve qualified dapp names

tommyldev:
## Summary

Host applications can now pass an already-qualified dotNS identifier such as `host-playground.paseo` through the signer SDK without it becoming the invalid `host-playground.paseo.dot`.

Bare labels retain the legacy `.dot` fallback, while `.dot`, `.paseo`, localhost, loopback, and localhost-subdomain forms remain complete identifiers.

## Validation

- Signer package: 146 tests passed
- Signer TypeScript typecheck passed
- Biome check passed
- Live Host Playground probe connected `host-playground.paseo` to its host product account and returned a 64-byte raw signature through PWallet's Rust core

### DIFF
--- a/product-sdk/packages/signer/src/providers/host.ts
+++ b/product-sdk/packages/signer/src/providers/host.ts
@@ -38,9 +38,9 @@ export interface HostProviderOptions {
      * not set, so `connect()` can still surface a usable account on hosts
      * that don't enumerate legacy accounts.
      *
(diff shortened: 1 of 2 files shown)


TarikGul (commented):
Ready to approve, just needs a changeset in ./pending-changesets

tommyldev:
Closing at the requester’s direction.

TarikGul (approved):

Issue #189, opened 2026-06-08
[tracking] Cloud Storage — migrate from host preimage lookup to bitswap retrieval (waiting on @parity/bulletin-sdk)


TarikGul:
### Summary

`@parity/product-sdk-cloud-storage` currently retrieves data via the host's `preimageManager`, which polls IPFS gateways. We want to migrate to a decentralized retrieval path once the upstream bulletin chain ships one.

This issue replaces and reframes #28, which incorrectly described the migration target as "direct chain queries." That target is not structurally available (see "Why the original premise was wrong" below).

### Current implementation

```
Upload: data → TransactionStorage.store → chain stores CID/hash + chunks via bitswap layer
Query:  CID → host preimageManager.lookup() → polls IPFS → returns bytes
```

Files involved (renamed from `packages/bulletin/` → `packages/cloud-storage/` in #59):
- `packages/cloud-storage/src/resolve-query.ts` — `QueryStrategy` is currently a single-variant union (`kind: "host-lookup"`) routing through `preimageManager`.
- `packages/cloud-storage/src/query.ts` — `queryBytes`, `queryJson`, `executeQuery`.

### Why the original premise was wrong

The bulletin chain pallet's `Transactions` storage holds **metadata only**, not bytes. Confirmed in `pallets/transaction-storage/src/lib.rs` on the upstream `main`:

```rust
pub struct TransactionInfo {
    chunk_root: H256,
    pub content_hash: ContentHash,
    pub hashing: HashingAlgorithm,
    pub cid_codec: CidCodec,
    size: u32,
    block_chunks: ChunkIndex,
}
```

There is no on-chain storage item that returns content bytes. The proposed `api.query.TransactionStorage.X → returns bytes` path does not exist and is not coming.

### What's actually planned upstream

Per the upstream SDK's own README (`paritytech/polkadot-bulletin-chain` → `sdk/typescript/src/index.ts`):

> *Data retrieval will be supported via the smoldot light client's `bitswap_block` RPC. This approach allows fully decentralized data retrieval directly from Bulletin validator nodes without relying on centralized gateways.*

The RPC has since been renamed to `bitswap_v1_get`. The work is tracked at https://github.com/paritytech/polkadot-bulletin-chain/pull/264.

The upstream README also explicitly **deprecates** IPFS gateway retrieval:

[...]

PR #333, opened 2026-08-27
feat(host): add getLocaleProvider for the host's selected language

kalininilya:
Adds `getLocaleProvider`, so a product renders in the language the user picked inside the host rather than inferring one from `navigator.language` — which reports the operating system's preference and is wrong whenever the two differ.

```ts
import { getLocaleProvider } from "@parity/product-sdk-host";

const provider = await getLocaleProvider();
const sub = provider?.subscribeLocale((locale) => {
  i18n.activate(SUPPORTED.has(locale.languageTag) ? locale.languageTag : "en");
});
```

`subscribeLocale` fires with the current locale and again on every change; the returned `HostSubscription` carries `unsubscribe` and `onInterrupt`. `getLocaleProvider` resolves to `null` outside a host container.

`languageTag` is a BCP 47 tag such as `"en"`, `"pt-BR"` or `"zh-Hans"`. The set is open — a host adds languages without an SDK release — so a product that ships no catalog entry for the tag it receives picks its own fallback.

The fake client models `locale` as not-modeled, so a test that reaches for it fails loudly rather than silently returning nothing.

Protocol counterpart: paritytech/host-rust-core#526.

## Blocked on

A `@parity/truapi` release carrying the `locale` domain. The catalog pins `^0.10.0`, which predates it, so CI will not go green until that lands.

## Verified

- `pnpm --filter @parity/product-sdk-host typecheck` against a local `@parity/truapi` built from paritytech/host-rust-core#526 — no locale errors. One unrelated error remains (`system.info` / `getProductContext` missing from the fake client), which reproduces on a clean tree with the same link: it is drift between host-rust-core `main` and the published 0.10.0, not this change.
- `pnpm --filter @parity/product-sdk-host test` — 20 files, 123 tests passed, including the new in-source `locale.ts` test.

### DIFF
--- a/product-sdk/packages/host/src/locale.ts
+++ b/product-sdk/packages/host/src/locale.ts
@@ -1,0 +1,76 @@
+// Copyright 2026 Parity Technologies (UK) Ltd.
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Higher-level wrapper for the host's locale subscription, backed by
+ * `truApi.locale.subscribe`.
+ *
+ * `getLocaleProvider` returns a handle whose `subscribeLocale(cb)` delivers a
+ * typed {@link LocaleInfo} — a `{ languageTag }` struct carrying a BCP 47 tag
+ * such as `"en"`, `"pt-BR"` or `"zh-Hans"` — and yields a
+ * {@link HostSubscription} (`unsubscribe` + `onInterrupt`).
+ *
+ * @module
+ */
+
+import type { HostLocaleSubscribeItem, TrUApiClient } from "@parity/truapi";
+
+import { getClient, subscribeWithInterrupt } from "./transport.js";
+import type { HostSubscription } from "./types.js";
+
+/**
+ * Host locale value. A `{ languageTag }` struct re-exported from
+ * `@parity/truapi`.
+ */
+export type LocaleInfo = HostLocaleSubscribeItem;
+
+/**
+ * Host locale provider handle. `subscribeLocale(callback)` receives a typed
+ * {@link LocaleInfo} on every change and returns a {@link HostSubscription}.
+ */
+export interface LocaleProvider {
+    subscribeLocale(callback: (locale: LocaleInfo) => void): HostSubscription;
+}
+
+/** Build a {@link LocaleProvider} over a TruAPI client's `locale` domain. */
+function adaptLocaleProvider(client: TrUApiClient): LocaleProvider {
+    return {
+        subscribeLocale(callback) {
+            return subscribeWithInterrupt(client.locale.subscribe(), callback);
+        },
+    };
+}
+
+/**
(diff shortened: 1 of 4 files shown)


duewarn:
Hey @valentinfernandez1 @decrypto21 @TarikGul @pgherveou  ! As https://github.com/paritytech/host-rust-core/pull/526 has been approved and merged, could you take a look at the implementation for the Product SDK? This functionality is included in the first release for mobile apps and the Funding modality.

valentinfernandez1:
Hi @kalininilya will take a look. There are some failing checks that you need to resolve. <

Also, how urgent is this? to know whether we need to create a truapi release asap to include the pr from host-rust-core or if it can wait till it lands alongide the next release.

product-sdk/pending-changesets/host-locale-provider.md
```diff
@@ -0,0 +1,31 @@
+---
+"@parity/product-sdk-host": minor
```
valentinfernandez1, line 2: The umbrella package should also be listed
```suggestion
"@parity/product-sdk-host": minor
"@parity/product-sdk": minor
```

product-sdk/packages/host/src/locale.ts
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * `getLocaleProvider` returns a handle whose `subscribeLocale(cb)` delivers a
+ * typed {@link LocaleInfo} — a `{ languageTag }` struct carrying a BCP 47 tag
+ * such as `"en"`, `"pt-BR"` or `"zh-Hans"` — and yields a
+ * {@link HostSubscription} (`unsubscribe` + `onInterrupt`).
+ *
+ * @module
+ */
+
+import type { HostLocaleSubscribeItem, TrUApiClient } from "@parity/truapi";
```
valentinfernandez1, line 15: Should wait till the change lands on a truapi release

valentinfernandez1 (changes requested):
Looks good, blocking until truapi release and fix on the changeset are fixed

duewarn:
@valentinfernandez1 thank you for review! These changes have not yet been merged on the hosts - this is expected to happen today or tomorrow. However, as soon as you release the new version of the Product SDK, it will be possible to integrate the new subscription into GetCash (Funding modality), which is included in the scope of the first release on 4 September.

product-sdk/pending-changesets/host-locale-provider.md
kalininilya: Ty, added

valentinfernandez1:
Ok I will do a release on truapi first so we can integrate this change.

valentinfernandez1:
Truapi release done, bumped by #353. I'm also bumping truapi on dotli with [#189](https://github.com/paritytech/dotli-community/pull/189) side so this should be available there on the next dotli release too.

valentinfernandez1 (approved):

PR #335, opened 2026-08-27
feat(descriptors,chain-client): re-add previewnet as a first-class environment

TarikGul:
## Summary

Re-adds `previewnet` as a first-class environment. `getChainAPI("previewnet")` resolves again, routing to the `previewnet.substrate.dev` endpoints for asset-hub, bulletin, and people.

## Why

Previewnet was dropped (#123) when its identity endpoints weren't secured for public use and its runtime matched paseo. Both have changed. The endpoints are secured now, and previewnet runs a Paseo runtime kept ahead of paseo-next-v2 (asset-hub `2000039` vs `2000036`, individuality `1000036` vs `1000032`), so products can build against upcoming runtime changes weeks early.

Complements #334, which documented the BYOD host dependency but did not re-add the preset.

## Changes

- `descriptors`: re-adds the `previewnet-asset-hub`, `previewnet-bulletin`, and `previewnet-individuality` subpath exports, generated fresh against the live endpoints with real (non-zero) `codeHash` values, so previewnet is covered by descriptor-drift detection like every other chain.
- `chain-client`: `previewnet` back in the `Environment` union.
- `cloud-storage`: `previewnet` back in `CloudStorageNetworks`.
- `host`: `BULLETIN_RPCS.previewnet` restored.

Purely additive. Paseo and production consumers are unaffected. Adapted to the current presets structure rather than reverting the drop.


product-sdk/packages/cloud-storage/src/networks.ts
```diff
@@ -28,6 +29,14 @@ export const CloudStorageNetworks = {
         genesisHash: "0x8cfe6717dc4becfda2e13c488a1e2061ff2dfee96e7d031157f72d36716c0a22",
         descriptor: paseoBulletinDescriptor,
     },
+    previewnet: {
+        genesisHash: "0x1144acd27f0e5b2c88da7dc12c111e396983dec036ccfb42da5bbb0dd7104e89",
```
valentinfernandez1, line 33: Not related to this PR but I am just wondering, since the host can provide the genesis hash with `getHostChainInfo()`, do we still need to keep them hardcoded?

valentinfernandez1 (approved):
🚀

product-sdk/packages/cloud-storage/src/networks.ts
TarikGul: I do know they are used in a few places, so lets see if we can. I would assume so - but might have some residual effects.

PR #341, opened 2026-08-28
feat(individuality): registration on readCurrentGame, fromPapi adapter

### LINKED ISSUE #340: individuality: consumer feedback from adopting readCurrentGame in humanity-spa
Part of #286. humanity-spa now reads the weekly game through `readCurrentGame` (paritytech/humanity-spa#58). Three things the integration had to work around, in order of value:

**1. `readCurrentGame` cannot answer "is this player registered?"**
The first thing a game UI asks about a running game is whether the current identity is in it. That is `Game.Players` under the account *and* alias keys, read at the same pinned block, with one hit meaning registered and a failed read meaning unknown rather than false. Today a consumer has to do that second read itself against `result.at` (which is exposed — good) and re-implement the reconciliation. Proposal: an optional `player` in `ReadCurrentGameOptions` and a `registration` field on the `Running` result:
```ts
readCurrentGame(chain, { player: { account, alias? } })
// Running → game.registration: { registered: boolean | null }  // null = a key read failed
```
`signup.ts` already models both key variants, so the shape exists.

**2. `GameChain` requires a chain-client-shaped object.**
`{ individuality: api, raw: { individuality: client } }` is the `getChainAPI` convention; a product that already owns a papi `client`/`api` (own provider selection, host bridge vs WS) needs an adapter (`toIndividualityChain` in humanity-spa). A `fromPapi(client, api)` helper would remove it without touching the current contract.

**3. `IndividualityDecodeError("unknown game state variant")` drops the variant.**
Including the offending value in the message is the difference between "runtime added a state" being obvious and being a debugging session.

We'll put up a PR for 1 and 2 unless there's a reason to shape them differently.

peetzweg:
Closes #340. `readCurrentGame` can now say whether a player is in the running game, and the reads accept a caller's own PAPI client.

**Why:** humanity-spa adopted `readCurrentGame` (paritytech/humanity-spa#58) and immediately had to do a second `Game.Players` read against `result.at` plus hand-build the chain object. Both are things every game UI needs.

**How:** `players?: AirdropRegistrant[]` on the read; the `Running` result carries `registration: Registered | NotRegistered | Unknown | Unchecked` read at the same pinned block. One person is keyed twice (account, alias) so any hit is `Registered`; a failed key read is `Unknown`, never `NotRegistered`, and doesn't fail the game read — the same rule humanity-spa applies today. The `Players` entry sits in a separate [`GamePlayersChain`](product-sdk/packages/individuality/src/game-read.ts) so the call without `players` keeps the narrower contract (overloads enforce it). [`fromPapi(client, api)`](product-sdk/packages/individuality/src/chain.ts) builds the chain-client shape without connecting or loading descriptors. `PlayerKey`/`playerKey` moved out of `signup.ts` into `player-key.ts`, shared by both reads.

Deliberately not done: the issue's point 3 (naming the variant in the decode error) — `game-decode.ts` has a test asserting chain data never reaches error messages, so that stays.

Changeset parked in `pending-changesets/` (individuality + umbrella minor). Umbrella contract test asserts paseo/previewnet satisfy `GamePlayersChain` and `fromPapi` satisfies every `*Chain`. Skill doc updated.

### DIFF
--- a/product-sdk/packages/individuality/src/chain.ts
+++ b/product-sdk/packages/individuality/src/chain.ts
@@ -1,0 +1,52 @@
+// Copyright 2026 Parity Technologies (UK) Ltd.
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Adapt a PAPI client the caller already holds to the chain shape every read in
+ * this package takes.
+ *
+ * The reads are typed against the `@parity/product-sdk-chain-client` convention —
+ * typed APIs keyed by chain name, raw clients under `raw` — so a product that
+ * resolves its own People-chain connection (its own provider, its own descriptors)
+ * would otherwise build that object by hand. `fromPapi` builds it, and nothing
+ * else: no connection is made and no descriptor is loaded.
+ *
+ * ```ts
+ * import { createClient } from "polkadot-api";
+ * import { fromPapi, readCurrentGame } from "@parity/product-sdk-individuality";
+ *
+ * const client = createClient(provider);
+ * const game = await readCurrentGame(fromPapi(client, client.getTypedApi(people)));
+ * ```
+ *
+ * The result is only as complete as `api`: a typed API generated from the
+ * individuality chain's metadata satisfies every `*Chain` interface here, and
+ * `tsc` checks each entry a read touches against it.
+ */
+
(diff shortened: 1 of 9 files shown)


TarikGul (approved):
Nice, super clean.

PR #343, opened 2026-08-28
feat(individuality): report candidate games remaining

tommyldev:
## Summary

Candidate personhood state now reports the number of consecutive attended games remaining before the current score threshold is reached. The derivation mirrors the runtime's streak-weighted score gains and resets an absent streak before projecting progress, so consumers no longer need to approximate progress from `score` and `personhoodThreshold`.

This adds a required `gamesRemaining` field to the exported `Candidate` variant and parks the corresponding minor changeset for a future release wave.

## Validation

- `pnpm check`
- `pnpm build`
- `pnpm typecheck`
- `pnpm test`

### DIFF
--- a/product-sdk/pending-changesets/individuality-candidate-progress.md
+++ b/product-sdk/pending-changesets/individuality-candidate-progress.md
@@ -1,5 +1,6 @@
 ---
 "@parity/product-sdk-individuality": minor
+"@parity/product-sdk": minor
 ---
 
 **Expose candidate progress as part of personhood state.**
--- a/product-sdk/pending-changesets/individuality-candidate-progress.md
+++ b/product-sdk/pending-changesets/individuality-candidate-progress.md
@@ -1,0 +1,9 @@
+---
+"@parity/product-sdk-individuality": minor
+---
+
+**Expose candidate progress as part of personhood state.**
+
+`derivePersonhoodState` now reports the consecutive attended games remaining on `Candidate`, accounting for streak-weighted score accrual and absence resets.
+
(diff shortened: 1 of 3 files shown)


product-sdk/pending-changesets/individuality-candidate-progress.md
```diff
@@ -0,0 +1,9 @@
+---
+"@parity/product-sdk-individuality": minor
```
TarikGul, line 2: The changeset only lists `@parity/product-sdk-individuality: minor`. But the umbrella package `@parity/product-sdk` re-exports individuality (`export * from "@parity/product-sdk-individuality"`), so per our umbrella policy in `RELEASES.md`, a minor bump on a constituent has to **also** list `@parity/product-sdk` at minor in the same changeset.

I confirmed with `pnpm changeset status`: as written, `individuality` bumps **minor** but the umbrella only cascades at **patch** — so the umbrella ends up a minor version behind the child whose new `gamesRemaining` surface it re-exports.

Fix is one line — add to the changeset frontmatter:

```
---
"@parity/product-sdk-individuality": minor
"@parity/product-sdk": minor
---
```

tommyldev: > a minor bump on a constituent has to **also** list `@parity/product-sdk` at minor in the same changeset.

Addressed in dbf4807 — the pending changeset now bumps both `@parity/product-sdk-individuality` and the umbrella `@parity/product-sdk` at minor.

TarikGul (approved):

PR #345, opened 2026-08-28
fix(sdk): remove localStorage.clear(), a silent no-op in production 

### LINKED ISSUE #344: LocalStorageApi.clear() is a silent no-op in production but empties in the test fake
`createApp().localStorage.clear()` does nothing. [`createApp.ts#L134-L137`](https://github.com/paritytech/product-sdk/blob/713ae3378b4aeddbbb3843454523e0465f7f5933/product-sdk/packages/sdk/src/core/createApp.ts#L134-L137) (`@parity/product-sdk@0.25.0`):

```typescript
clear: async () => {
    // LocalKvStore doesn't have clear - this is a no-op
    log.debug("clear() is not supported in container storage mode");
},
```

It resolves successfully and logs at debug, so callers get no signal.

The test fake does empty, and says so:

```typescript
async clear() {
    // Unlike the production adapter (a no-op in container mode), the fake
    // actually empties, so tests get real isolation.
    host.reset();
```

So a test asserting `clear()` empties storage passes, and the same code no-ops in production. That divergence is worth fixing regardless of what happens to `clear()` itself.

It also cannot be implemented as-is: `HostLocalStorage` exposes no enumeration (`keys` / `list` / `entries`), so there is nothing to iterate. `HostLocalStorage.clear(key: string)` is a single-key remove, which may be where the clear-all shape came from.

Options, in preference order:

1. Remove `clear()` from `LocalStorageApi` and drop it from the fake. `remove(key)` already works at every layer.
2. Add enumeration to `HostLocalStorage`, then implement `clear()` over it.
3. Keep it, but throw or return an err instead of resolving silently, and make the fake match.

Either way the fake and the production adapter should agree.

Docs side tracked in polkadot-developers/polkadot-docs#1746.

TarikGul:
Closes #344

## Problem

`createApp().localStorage.clear()` resolves successfully but does nothing in a host container — it only logs at debug. The `createFakeApp` test fake actually empties, so a test asserting `clear()` wipes storage passes against code that no-ops for real users.

It also can't be implemented as-is: the host localStorage protocol exposes no key enumeration — only per-key `read` / `write` and a single-key `clear(key)` through `@parity/truapi` → `HostLocalStorage` → `LocalKvStore` — so there is nothing to iterate for a clear-all.

## Fix

Remove `clear()` from `LocalStorageApi`, the production adapter, and the fake, so the two agree. `remove(key)` already works at every layer and stays.

Both `createApp` and the fake now build their `localStorage` through one shared `createLocalStorageApi` adapter, so the shipped code and the test fake can no longer drift — which is what let `clear()` no-op in production while the fake emptied.

## Migration

Use `remove(key)` to delete keys individually. There is no clear-all; if you need one, track the keys your app writes and remove them.

**Breaking for callers:** an untyped `app.localStorage.clear()` now throws `TypeError: clear is not a function` where it used to resolve.

**Breaking for implementors:** `clear` is gone from the exported `LocalStorageApi`, so anyone writing one inline — e.g. the `localStorage` override to `createFakeApp` — must delete their `clear` to compile.

### DIFF
--- a/product-sdk/packages/sdk/src/core/createApp.ts
+++ b/product-sdk/packages/sdk/src/core/createApp.ts
@@ -16,6 +16,7 @@ import type {
     CloudStorageApi,
     LocalStorageApi,
 } from "./types.js";
+import { createLocalStorageApi } from "./local-storage-api.js";
 import { configure, createLogger } from "@parity/product-sdk-logger";
 import { createLocalKvStore } from "@parity/product-sdk-local-storage";
 import { SignerManager } from "@parity/product-sdk-signer";
@@ -124,14 +125,8 @@ export async function createApp(config: AppConfig): Promise<App> {
         log.debug("Cloud Storage client disabled");
     }
 
-    // Create storage API adapter
-    const localStorageApi: LocalStorageApi = {
-        get: (key) => localKvStore.get(key),
-        set: (key, value) => localKvStore.set(key, value),
-        getJSON: <T>(key: string) => localKvStore.getJSON<T>(key),
-        setJSON: <T>(key: string, value: T) => localKvStore.setJSON(key, value),
-        remove: (key) => localKvStore.remove(key),
-    };
+    // Storage API adapter — shared with `createFakeApp` so the two cannot drift.
+    const localStorageApi: LocalStorageApi = createLocalStorageApi(localKvStore);
 
     // Create wallet API adapter
     const walletApi = createWalletApi(signerManager);
--- a/product-sdk/packages/sdk/src/core/createApp.ts
+++ b/product-sdk/packages/sdk/src/core/createApp.ts
@@ -131,10 +131,6 @@ export async function createApp(config: AppConfig): Promise<App> {
         getJSON: <T>(key: string) => localKvStore.getJSON<T>(key),
         setJSON: <T>(key: string, value: T) => localKvStore.setJSON(key, value),
         remove: (key) => localKvStore.remove(key),
-        clear: async () => {
-            // LocalKvStore doesn't have clear - this is a no-op
-            log.debug("clear() is not supported in container storage mode");
-        },
     };
 
     // Create wallet API adapter
--- a/product-sdk/packages/sdk/src/core/types.ts
+++ b/product-sdk/packages/sdk/src/core/types.ts
@@ -72,7 +72,14 @@ export interface WalletApi {
     createProof(message: Uint8Array): Promise<Uint8Array>;
 }
 
-/** Storage API exposed by the SDK */
+/**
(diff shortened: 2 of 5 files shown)


product-sdk/packages/sdk/src/core/createApp.ts
```diff
@@ -131,10 +131,6 @@ export async function createApp(config: AppConfig): Promise<App> {
         getJSON: <T>(key: string) => localKvStore.getJSON<T>(key),
         setJSON: <T>(key: string, value: T) => localKvStore.setJSON(key, value),
         remove: (key) => localKvStore.remove(key),
```
Imod7: This file has no tests, so the adapter that ships is never run by one while the fake is. That is what let `clear()` drift, and the five remaining members are wired the same way. Please extract this object into a `createLocalStorageApi(localKvStore)` factory and add an in-source test driving one assertion set through both it and `createFakeApp().localStorage`. The same idea would help the wallet and cloud-storage fakes.

product-sdk/pending-changesets/remove-localstorage-clear.md
```diff
@@ -0,0 +1,11 @@
+---
+"@parity/product-sdk": minor
```
Imod7, line 2: Please cover the two breaks it does not mention yet.
- Implementors: anyone who writes their own `LocalStorageApi` inline, such as the `localStorage` override passed to `createFakeApp`, stops compiling until they delete their `clear`. `pending-changesets/README.md` has a "Breaking for implementors" wording to copy.
- Untyped callers: `app.localStorage.clear()` used to resolve and now throws `TypeError: app.localStorage.clear is not a function`.

product-sdk/packages/sdk/src/core/types.ts
```diff
@@ -82,10 +82,15 @@ export interface LocalStorageApi {
...
     /** Set a JSON value by key */
     setJSON<T = unknown>(key: string, value: T): Promise<void>;
-    /** Remove a value by key */
+    /**
+     * Remove a value by key.
+     *
+     * There is no `clear()`: host localStorage exposes no key enumeration
+     * (only per-key read / write / remove), so a clear-all cannot be
+     * implemented — it was a silent no-op in a host container. Remove keys
+     * individually with {@link remove}.
```
Imod7: The `{@link remove}` sits inside `remove`'s own doc comment, so it links back to itself. And a reader looking for a clear-all reads the interface doc, not `remove`'s, so the explanation belongs on `LocalStorageApi`, whose comment is currently just "Storage API exposed by the SDK". Moving it there fixes both.

Imod7 (approved):
Better make the line `Closes #344` in the PR description because the parenthesis break it and it will not close the corresponding issue.

PR #346, opened 2026-08-30
feat(individuality): decode draw event ids

Imod7:
Related to #286

### Description

Adds `parsePeopleAirdropsEventId(eventId)`. It returns the draw index inside a `PeopleAirdrops` event id, or `null` for anything else.

The package could build these ids but not read them back. Callers who list `Airdrop.Events` get ids they never built, from two schedulers mixed together, so they had to reimplement the 32 byte layout to tell them apart. humanity-spa does that today.

### Changes

The function sits in `airdrop-ids.ts` beside the encoder it inverts, so one test asserts they round-trip. Exported from the package index, with a note in `SKILL.md` on the two directions and which one a caller needs.

Two decisions worth flagging. It returns `null` rather than throwing, because `Airdrop.Events` is shared and a foreign id is ordinary input. And the `0x` prefix is matched case-sensitively while the body is not, which looks inconsistent but is deliberate: PAPI hands us the id and decodes it again on the next read, and it accepts either case in the body but only a lower-case `0x`. Lower-casing the whole string would accept `0X...` and report an index for an id that then reads back as different bytes.

### Testing

17 cases in the existing in-source block: round-trips through `u64::MAX`, a `Game` id and a foreign base returning `null`, an upper-case body decoding, and malformed inputs. One case is a bare hex literal, the same 32 bytes humanity-spa's stub backend already hardcodes, pasted in rather than derived. No dependency, just the value. A vector built with this module's own encoder would pass even if encode and decode drifted together, which is the failure this layout is exposed to since nothing in metadata describes it.

```bash
pnpm install && pnpm -r build && pnpm -r test && pnpm check
```

411 tests over 20 files in this package, zero typecheck errors, biome clean. Build the siblings first: without `pnpm -r build` the in-source tests cannot resolve `@parity/result`, on `main` as well as on this branch.

### DIFF
--- a/product-sdk/packages/individuality/src/airdrop-ids.ts
+++ b/product-sdk/packages/individuality/src/airdrop-ids.ts
@@ -156,19 +156,20 @@ const PEOPLE_AIRDROPS_PREFIX_HEX = toEventId(
     baseBytes(PEOPLE_AIRDROPS_EVENT_ID_BASE, PEOPLE_AIRDROPS_BASE_BYTES),
 ).slice(2);
 
-/** One shape check for prefix, length and hex digits, so the three cannot disagree. */
-const EVENT_ID_PATTERN = new RegExp(`^0x[0-9a-f]{${EVENT_ID_BYTES * 2}}$`);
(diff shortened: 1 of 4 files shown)


TarikGul (commented):
Nice job, clean

TarikGul (approved):

PR #347, opened 2026-08-31
feat(individuality): product proof contexts and personhood ring locations

peetzweg:
First piece of the lite-personhood sign-up flow (#354): pure helpers for the proof contexts and ring locations that dim2 and humanity each hand-roll today.

Why: every context a host signs under is blake2b-256("product/" ++ productId ++ "/" ++ suffix) with the RFC-0024 Index/Raw expansion, and two of the personhood product's five contexts never reach metadata, so consumers have to derive them client-side.

What: productContext / contextSuffixBytes / personhoodContext(tld, name) as pure functions; peopleRing / litePeopleRing build the two RingLocations, structurally compatible with product-sdk-host without depending on it; readScoreContext checks Score.score_context is the product derivation of peopl.<network suffix>/Index(0) and answers NotProductDerived on the ok channel, so proof flows stop before the chain rejects with nothing local to read. Product ids are always full DotNS ids with no TLD default; a tld must be a single lower-case label inside the runtime's 16-byte bound, and a suffix that is not valid UTF-8 errors instead of becoming replacement characters in a product id.

The suffix source is typed, not probed. individuality-community#20 deleted the Score.Suffix constant in favour of Root-settable NetworkSuffix.NetworkSuffix storage, and that pallet is testnet-only, so there are three narrow contracts: NetworkSuffixChain (storage, read at a pinned block), LegacySuffixChain (the constant, previewnet only), or neither, where the caller passes tld. readScoreContext is overloaded across them, so a chain that can resolve no suffix fails to compile rather than at runtime. Separate contracts rather than one with optional members, because an optional member constrains nothing structurally and the umbrella guard would stop separating environments. runScoreContextRead is the throwing variant, for a composed read that already pinned a block.

Tests pin previewnet's constants (spec 1000036), both collection ids and the Index(0) vector. contract.test.ts asserts the three contracts per environment and both RingLocation directions, with the negatives checked by flipping them. 424 package tests, umbrella typecheck and tests clean. The individuality skill documents the new surface.

Stack: #350 consumes ScoreContextChain and calls readScoreContext inside a composed read, so it needs a suffix contract (or a tld) and runScoreContextRead with its own snapshot. #352 references readScoreContext in JSDoc only.

### DIFF
--- a/product-sdk/packages/individuality/src/contexts.ts
+++ b/product-sdk/packages/individuality/src/contexts.ts
@@ -125,6 +125,9 @@ export const PERSONHOOD_CONTEXT_INDEX = {
 
 export type PersonhoodContextName = keyof typeof PERSONHOOD_CONTEXT_INDEX;
 
+/** `indiv_support::context::MAX_NETWORK_SUFFIX_LENGTH`. */
+const MAX_TLD_BYTES = 16;
+
 /**
  * The personhood product's context for `name` on the network issuing `.<tld>`
  * names: `productContext("peopl.<tld>", Index(PERSONHOOD_CONTEXT_INDEX[name]))`.
@@ -133,13 +136,18 @@ export type PersonhoodContextName = keyof typeof PERSONHOOD_CONTEXT_INDEX;
  * paseo networks), so the same context name on two networks is two different
  * values — which is why it is a parameter and never a default.
  *
- * @param tld - a single DotNS label, without the leading dot.
- * @throws ProductIndividualityError on a TLD that is empty or not a single
- *   label: `"peopl" + ".te.st"` and `"peopl.te" + ".st"` would collide, so a
- *   composite here is always a caller bug.
+ * @param tld - a single lower-case DotNS label, without the leading dot.
+ * @throws ProductIndividualityError on a tld the runtime cannot represent, or a
+ *   composite one: `"peopl" + ".te.st"` and `"peopl.te" + ".st"` would collide.
  */
 export function personhoodContext(tld: string, name: PersonhoodContextName): Uint8Array {
-    if (tld.length === 0 || tld.includes(".") || tld.includes("/")) {
+    if (
+        tld.length === 0 ||
+        tld.includes(".") ||
+        tld.includes("/") ||
+        tld !== tld.toLowerCase() ||
+        utf8ToBytes(tld).length > MAX_TLD_BYTES
+    ) {
         throw new ProductIndividualityError("personhood context tld must be a single dotns label");
     }
     return productContext(`${PERSONHOOD_PRODUCT_NAME}.${tld}`, {
@@ -251,8 +259,12 @@ if (import.meta.vitest) {
             expect(contexts.size).toBe(tlds.length);
         });
 
-        test.each(["", "te.st", "te/st"])("rejects the tld %j", (tld) => {
+        test.each(["", "te.st", "te/st", "TEST", "a".repeat(17)])("rejects the tld %j", (tld) => {
             expect(() => personhoodContext(tld, "score")).toThrow(ProductIndividualityError);
+        });
+
+        test("accepts a tld at the 16-byte bound", () => {
+            expect(() => personhoodContext("a".repeat(16), "score")).not.toThrow();
         });
     });
 }
--- a/product-sdk/packages/individuality/src/contexts.ts
+++ b/product-sdk/packages/individuality/src/contexts.ts
@@ -1,0 +1,258 @@
+// Copyright 2026 Parity Technologies (UK) Ltd.
+// SPDX-License-Identifier: Apache-2.0
+/**
+ * Product-scoped ring-VRF proof contexts (RFC-0004 / RFC-0022 / RFC-0024).
+ *
+ * A host never lets a product choose its proof context: it derives it from the
+ * product's identity, so a product can prove personhood only in its own
+ * namespace. Since individuality `0fec7071` ("Use product-owned personhood
+ * contexts") the chain derives its own contexts the same way — every context a
+ * runtime accepts is
+ *
+ * ```
+ * context = blake2b-256("product/" ++ product_id ++ "/" ++ suffix_bytes)
+ * ```
+ *
+ * where `suffix_bytes` expands the RFC-0024 `Index`/`Raw` selector:
+ *
+ * ```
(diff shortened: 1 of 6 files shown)


product-sdk/packages/individuality/src/rings.ts
```diff
@@ -0,0 +1,320 @@
...
+ * ```
+ */
+export interface ScoreContextChain {
+    individuality: {
+        constants: {
+            Score: {
+                /** The 32-byte context scores are proven in, as `0x` hex. */
+                score_context(): Promise<string>;
+                /** The network's DotNS TLD, as UTF-8 bytes (`"test"`, `"paseo"`). */
+                Suffix(): Promise<Uint8Array>;
```
Imod7, line 149: Requiring `Score.Suffix` here limits `readScoreContext` to previewnet. Paseo (spec 1000032) and devnet do not publish it, so there the call returns `err`, not `ok(NotProductDerived)`.

product-sdk/packages/individuality/src/rings.ts
```diff
@@ -0,0 +1,320 @@
...
+export async function readScoreContext(
+    chain: ScoreContextChain,
+    options: ReadScoreContextOptions = {},
+): Promise<Result<ScoreContext, ProductIndividualityError>> {
+    try {
+        options.signal?.throwIfAborted();
+        const constants = chain.individuality.constants.Score;
+        const [constant, suffix] = await Promise.all([
+            constants.score_context(),
+            constants.Suffix(),
```
Imod7: `Score.Suffix` is going away. `individuality-community` `ad318ef` (#20, 2026-08-24) dropped `#[pallet::constant]` from it and moved the value into `NetworkSuffix.NetworkSuffix` storage, settable by Root, so it disappears from metadata on the next runtime upgrade and is no longer immutable.

product-sdk/packages/individuality/src/rings.ts
```diff
@@ -0,0 +1,320 @@
...
+        ]);
+        if (!CONTEXT_HEX.test(constant)) {
+            throw new IndividualityDecodeError("score context constant is not 32 bytes of hex");
+        }
+        const context = hexToBytes(constant.slice(2));
+        const tld = new TextDecoder().decode(suffix);
+        const productId = `${PERSONHOOD_PRODUCT_NAME}.${tld}`;
+        const derived = productContext(productId, {
```
Imod7: This rebuilds what `personhoodContext(tld, "score")` already does, so it skips that function's tld check. Please call it instead, and make the `TextDecoder` above it fatal: a `Score.Suffix` that is not valid UTF-8 currently becomes replacement characters in `productId` rather than an error. Same pairing for any later lite auth context read.

product-sdk/packages/individuality/src/contexts.ts
```diff
@@ -0,0 +1,258 @@
...
+ * paseo networks), so the same context name on two networks is two different
+ * values — which is why it is a parameter and never a default.
+ *
+ * @param tld - a single DotNS label, without the leading dot.
+ * @throws ProductIndividualityError on a TLD that is empty or not a single
+ *   label: `"peopl" + ".te.st"` and `"peopl.te" + ".st"` would collide, so a
+ *   composite here is always a caller bug.
+ */
+export function personhoodContext(tld: string, name: PersonhoodContextName): Uint8Array {
+    if (tld.length === 0 || tld.includes(".") || tld.includes("/")) {
```
Imod7: Worth rejecting two more cases here: an uppercase tld, and one over the runtime's 16-byte bound (`MAX_NETWORK_SUFFIX_LENGTH`). Both pass silently today and fail much later as an unexplained proof rejection. `tld !== tld.toLowerCase()` and a length check cover them, plus two rows on the existing `test.each`.

Imod7 (commented):
Please add `ScoreContextChain` to `packages/sdk/src/individuality/contract.test.ts`, beside the existing `IndividualityChain` assertions.

Please add the new surface to `product-sdk/skills/product-sdk-individuality/SKILL.md`.

product-sdk/packages/individuality/src/rings.ts
Imod7: **Fixed** in [`3a3e215`](https://github.com/paritytech/product-sdk/commit/3a3e2151e8ebe3dd4a9667f28a3d08a3bc5a2718). The contract split three ways and the TLD became a typed input, so a chain that can resolve none is a compile error.

product-sdk/packages/individuality/src/rings.ts
Imod7: **Fixed** in [`3a3e215`](https://github.com/paritytech/product-sdk/commit/3a3e2151e8ebe3dd4a9667f28a3d08a3bc5a2718), with a resolution order, plus [`2688274`](https://github.com/paritytech/product-sdk/commit/2688274572a990e6b7420c197ef4dd7caed25ffc) guarding the storage read when no block source is available.

product-sdk/packages/individuality/src/rings.ts
Imod7: **Fixed** in [`3a3e215`](https://github.com/paritytech/product-sdk/commit/3a3e2151e8ebe3dd4a9667f28a3d08a3bc5a2718), alongside the Criticals as expected.

product-sdk/packages/individuality/src/contexts.ts
Imod7: **Fixed** in [`c935304`](https://github.com/paritytech/product-sdk/commit/c9353042fa0de76eba333cfaecb23fa82aac6226). The bound is pinned from both sides: 17 bytes rejected, 16 accepted.

Imod7:
> Please add `ScoreContextChain` to `packages/sdk/src/individuality/contract.test.ts`, beside the existing `IndividualityChain` assertions.

**Fixed** in [`2e5cead`](https://github.com/paritytech/product-sdk/commit/2e5ceadf68882c5052152ef3ed3306b7766a8d0d). All three suffix contracts are asserted per environment, both `RingLocation` directions are pinned, and `ScoreContextChain` joined `FromPapiSatisfiesContracts`.

> Please add the new surface to `product-sdk/skills/product-sdk-individuality/SKILL.md`.

**Fixed** in [`b3e9386`](https://github.com/paritytech/product-sdk/commit/b3e93868364518035e8e841240b26125dc42cc87).

Imod7 (approved):

PR #348, opened 2026-08-31
feat(individuality): withLiteAlias runs a call under a lite-person origin

peetzweg:
Second piece of the lite-personhood sign-up flow (#354): `withLiteAlias(signer, info)`, the lite sibling of `withAsPerson`, filling the `PeopleLiteAuth` extension inside `signTx`.

Why: the two-transaction lite sign-up (unsigned `PeopleLite.set_alias_account` bind, then the signed, fee-free `Game.sign_up_with_account_lite_invite`) needs this extension on both legs, and dim2 hand-encodes it today.

How: three variants. `AliasWithAccount` reads the nonce back out of the `CheckNonce` slot PAPI filled, so the two cannot disagree. `AliasWithProof` forces `VerifyMultiSignature=Disabled`, so the host emits an unsigned general transaction with origin `None`. `AliasWithAccountRevised` builds the pallet's `(implication, "revise", account, nonce)` tuple. Every value round-trips through a codec built from the metadata being signed against, so devnet's shorter field list throws instead of emitting plausible wrong bytes. `AsLitePerson` is deliberately absent: it authenticates the host-custody canonical lite account, which no product-side signer can be.

Both signers now run one ordered `signTx` body through an internal `withOriginExtension`, each keeping its own `buildValue`; that order is the fragile part and was previously implemented and explained twice. `buildImplication` and `implicationMessage` now require the extension identifier instead of defaulting to `AsPerson`, which turned a silent wrong-slice hash into a compile error at 19 call sites. Neither is exported, so nothing changes for `withAsPerson`.

The individuality skill now documents the flow end to end: the ordering trap (`Custom(175)`, `NoAliasBinding`, when the bind leg has not landed), the two contexts the chain allowlists for the bind (`peopleLiteAuth` and `score`, both `peopl.<tld>`), and that the proof therefore comes from the personhood product rather than from dim2.

489 package tests pass; workspace tests, typecheck and biome clean.


product-sdk/packages/individuality/src/as-lite-alias-signer.ts
```diff
@@ -0,0 +1,592 @@
...
+ * `signBytes` and `publicKey` pass through untouched. PAPI stamps `publicKey`
+ * into the extrinsic and uses it to fetch the nonce, so it has to stay the
+ * inner signer's.
+ *
+ * @param signer - the signer to wrap, e.g. from
+ *   `AccountsProvider.getProductAccountSigner`.
+ * @param info - which lite-person origin to use, and where the proof comes from.
+ * @returns a `PolkadotSigner` usable anywhere the original was.
+ */
+export function withLiteAlias(signer: PolkadotSigner, info: LiteAliasInfo): PolkadotSigner {
```
Imod7, line 195: This ends up being the same body as `withAsPerson`. With the test blocks stripped and the four differing names normalised, all that is left between the two files is the identifier handed to `implicationMessage` and `buildImplication`; even the branch is the same literal.

product-sdk/packages/individuality/src/as-lite-alias-signer.ts
```diff
@@ -0,0 +1,592 @@
...
+     * Signed by an account already bound to the lite alias, via
+     * `PeopleLite.set_alias_account`. Needs no proof.
+     *
+     * The chain reads `PeopleLite.AccountToAlias` for the signing account and
+     * requires the stored ring revision to be current. When it is not, the
+     * chain answers `Custom(172)` (stale alias) and `AliasWithAccountRevised`
+     * is the variant that fixes it. That cannot be detected from here without
+     * reading the ring root.
+     */
+    | { tag: "AliasWithAccount" }
```
Imod7, line 89: [optional]: The doc above this names `Custom(172)` for a stale binding, which is right, but not `Custom(175)` for no binding at all.

product-sdk/packages/individuality/src/origin-extension.ts
```diff
@@ -0,0 +1,270 @@
...
+ * requesting a ring VRF proof defensively, caching the decoded pipeline — is
+ * identical, and identical in ways that are easy to get subtly wrong, so it
+ * lives here once. Internal to the package on purpose: these are
+ * implementation details of the two signers, not a public contract.
+ *
+ * The errors are `AsPersonError` throughout. It is the write half's error
+ * class, predates the second extension, and callers already catch it; a
+ * parallel class per extension would split one failure domain in two.
+ */
+import type { PolkadotSigner } from "polkadot-api";
```
Imod7, line 19: This import is not used anywhere in the file.

Imod7 (commented):
Please update the individuality skill doc also.

In `product-sdk/packages/individuality/src/as-person-implication.ts`, lines 72 and 110, both:

```
    identifier: string = AS_PERSON,
```

Now that a second extension calls these, could the identifier become required?

product-sdk/packages/individuality/src/as-lite-alias-signer.ts
Imod7: **Fixed** in [`c7c9382`](https://github.com/paritytech/product-sdk/commit/c7c938238a86dfb57e88d7af386c1e06ee34a218), by the first of the three options: both signers now delegate to an internal `withOriginExtension(signer, spec)` carrying the ordered steps, each keeping its own `buildValue`. 489 tests unchanged, which is the evidence it is a pure move.

product-sdk/packages/individuality/src/as-lite-alias-signer.ts
Imod7: **Fixed** in [`bc48ff6`](https://github.com/paritytech/product-sdk/commit/bc48ff61ba470226b97839fc4e8c51609fdde880), including the generalization: `AliasWithAccountRevised` names `Custom(175)` too, since that arm also opens from `AccountToAlias`. All three codes now carry their pallet symbol.

product-sdk/packages/individuality/src/origin-extension.ts
Imod7: **Fixed** in [`c7c9382`](https://github.com/paritytech/product-sdk/commit/c7c938238a86dfb57e88d7af386c1e06ee34a218): `withOriginExtension` needs `PolkadotSigner`, so the import became genuine rather than being deleted and re-added.

Imod7:
> Please update the individuality skill doc also.

**Fixed** in [`0f1efd1`](https://github.com/paritytech/product-sdk/commit/0f1efd1f305577a4d4e5a83691bfbb5d88a3166f). The skill gained a lite-flow section, the frontmatter and the write-half bullet name `withLiteAlias`, the "ask for them to be exported" sentence is gone, and `index.ts` names both signers. One claim in that new section was itself wrong and was corrected in [`e7fcc22`](https://github.com/paritytech/product-sdk/commit/e7fcc226e98419f1480114c1f0a226926c81bc41): the chain's context allowlist is the gate, not host-side identity.

> In `product-sdk/packages/individuality/src/as-person-implication.ts`, lines 72 and 110, both:
> 
> ```
>     identifier: string = AS_PERSON,
> ```
> 
> Now that a second extension calls these, could the identifier become required?

**Fixed** in [`c7c9382`](https://github.com/paritytech/product-sdk/commit/c7c938238a86dfb57e88d7af386c1e06ee34a218). Making the parameter required flushed out 19 call sites that had been taking the `AsPerson` default silently, including in `as-person-implication.ts`'s own tests.

Imod7 (approved):

PR #349, opened 2026-08-31
feat(individuality): Score.register builders and withScoreParticipant

peetzweg:
Follows https://github.com/paritytech/product-sdk/pull/348, now merged. The registration piece of https://github.com/paritytech/product-sdk/issues/354: what a product needs to turn a scored participant into a full person.

**Why:** dim2 hand-rolls the `Score.register` message and the `ScoreAsParticipant` extension today, and the personhood product needs the same builders for the cross-product handoff.

**How:** `registerMessage(account)` pins the byte-exact `"pop register using" ++ account` contract, a raw 50-byte concatenation, never SCALE. `registerPersonhoodTx` builds the call from a caller-supplied `(memberKey, proofOfOwnership)` pair, width-checked and opaque: only the personhood product's own host session can mint it, so the builder never tries, which lets the same code serve today's two-product handoff and a future single-product path. `readRegistrationEligibility` folds `Score.Participants` and `Score.PersonhoodThreshold` at one pinned block into a `readyToRegister` verdict. `withScoreParticipant(signer)` is the third signer on the shared origin-extension machinery: fee-free dispatch from a 0-balance participant account.

One judgment call: the proposal's caller-supplied nonce was dropped in favour of reading it back out of the `CheckNonce` slot PAPI filled, since the chain rejects any disagreement with nothing local to read.

**From review:** the two new chain contracts are asserted against the paseo, previewnet and devnet descriptors in `packages/sdk`, `registerMessage` rethrows a malformed address as `ProductIndividualityError` instead of PAPI's bare `Invalid checksum`, `readRegistrationEligibility` takes the account arm only, and the individuality skill doc gains a registration section.

Message bytes and flow verified live on previewnet (spec 1000036) on 2026-08-28. 520 package tests pass; workspace tests + typecheck + biome clean.


product-sdk/packages/individuality/src/register.ts
```diff
@@ -0,0 +1,549 @@
...
+/**
+ * The `Score.register` call, typed structurally so the package needs no
+ * descriptor dependency. Matched by hand against the previewnet descriptors on
+ * 2026-08-31:
+ *
+ * ```
+ * Score.register: TxDescriptor<{ key?: [SizedHex<32>, SizedHex<64>] }>
+ * ```
+ */
+export interface RegisterChain<Tx = unknown> {
```
Imod7, line 124: Please assert this contract and `RegistrationEligibilityChain` in `packages/sdk/src/individuality/contract.test.ts`, where every other chain contract here is asserted. Nothing else catches a descriptor regeneration, because the test doubles satisfy the contract by construction.

product-sdk/packages/individuality/src/register.ts
```diff
@@ -0,0 +1,549 @@
...
+ * The pallet builds the same bytes with
+ * `account.using_encoded(|b| [PREFIX, b].concat())`, and `AccountId32` encodes
+ * as its bare 32 bytes, so this is a raw concatenation with no SCALE framing.
+ *
+ * @param account - the registering account (the transaction signer), as an
+ *   SS58 address or its raw 32-byte public key.
+ * @throws ProductIndividualityError when the account is not 32 bytes.
+ */
+export function registerMessage(account: SS58String | Uint8Array): Uint8Array {
+    const bytes = typeof account === "string" ? AccountId().enc(account) : account;
```
Imod7: A malformed SS58 address throws PAPI's bare `Error: Invalid checksum` here, not a `ProductIndividualityError`, so `isSdkError(e)` is false and the `@throws` line above is only half true. Could you wrap this call and rethrow with the original as `cause`.

product-sdk/packages/individuality/src/register.ts
```diff
@@ -0,0 +1,549 @@
...
+                };
+            };
+        };
+    };
+}
+
+/** Options for {@link readRegistrationEligibility}. */
+export interface ReadRegistrationEligibilityOptions {
+    /** Whose standing to read. Must be the identity that would sign `register`. */
+    registrant: AirdropRegistrant;
```
Imod7: [optional]: `Score.register` reads `Participants[Account(signer)]` and consults no other key, so the alias arm of `AirdropRegistrant` reads a record this call never looks at. Worth narrowing to `Extract<AirdropRegistrant, { tag: "Account" }>` in this PR while the type is unreleased and it costs nothing; `playerKey` still accepts it, so the body is unchanged.

product-sdk/packages/individuality/src/as-score-participant-signer.ts
```diff
@@ -0,0 +1,290 @@
...
+            // False is an immediate rejection for a restricted origin, and false
+            // is PAPI's default. Skipped only when the chain has no such extension.
+            if (pipeline.extensions.some((slot) => slot.identifier === RESTRICT_ORIGINS)) {
+                extensions = withSlot(
+                    pipeline,
+                    extensions,
+                    RESTRICT_ORIGINS,
+                    encodeChecked(pipeline.codec(pipeline.slot(RESTRICT_ORIGINS).type), true),
+                );
+            }
```
Imod7: [optional]: This block is byte-identical to the same one at `as-person-signer.ts` line 209 and `as-lite-alias-signer.ts` line 212. Consider a `withRestrictOrigins(pipeline, extensions)` helper in `origin-extension.ts`.

product-sdk/packages/individuality/src/register.ts
```diff
@@ -0,0 +1,549 @@
...
+
+/**
+ * A participant's standing against the registration guards, at one pinned
+ * block.
+ */
+export interface RegistrationEligibility {
+    at: FinalizedSnapshot;
+    /** `null` for an account `Score` has never seen. */
+    participant: PersonhoodParticipant | null;
+    /** `Score.PersonhoodThreshold` at the same block. A `u8` on chain. */
```
Imod7: [nit]: It is a `u8` on paseo and previewnet but a `u32` on devnet.

Imod7 (commented):
Please update the individuality skill doc in this PR. It has no registration section, so anyone building this flow from it hand-rolls the message, and the quoted line is now false since there are three exported signers.

product-sdk/packages/individuality/src/register.ts
Imod7: Fixed in [`d111e1d`](https://github.com/paritytech/product-sdk/commit/d111e1d4ea8bfd4972de3d5d0e18bb50cf7e1b59).

product-sdk/packages/individuality/src/register.ts
Imod7: Fixed in [`3a0e9fb`](https://github.com/paritytech/product-sdk/commit/3a0e9fb93715a2957004b767b2fdc232f779b01a).

product-sdk/packages/individuality/src/register.ts
Imod7: Fixed in [`3a0e9fb`](https://github.com/paritytech/product-sdk/commit/3a0e9fb93715a2957004b767b2fdc232f779b01a).

product-sdk/packages/individuality/src/as-score-participant-signer.ts
Imod7: Fixed in [`54dc146`](https://github.com/paritytech/product-sdk/commit/54dc146c80a35ebffe70782b3bf0faf503a769fa).

product-sdk/packages/individuality/src/register.ts
Imod7: Fixed in [`9cad132`](https://github.com/paritytech/product-sdk/commit/9cad132017c409f7a924ef8286465db0d3579d46).

Imod7:
> Please update the individuality skill doc in this PR. 

Fixed in [`9cad132`](https://github.com/paritytech/product-sdk/commit/9cad132017c409f7a924ef8286465db0d3579d46).

Imod7 (approved):

PR #350, opened 2026-08-31
feat(individuality): the free lite-personhood game sign-up

peetzweg:
The sign-up piece of the lite-personhood flow (https://github.com/paritytech/product-sdk/issues/354): the fee-free `Game.sign_up_with_account_lite_invite` builder, and the read that says whether either leg is worth submitting. dim2 hand-rolls this today; humanity-spa needs the same reads next. https://github.com/paritytech/product-sdk/pull/348 landed the signer it rides on, so this targets `main`.

**How:** `readLiteSignUpRequirement` runs the account requirement read plus the lite gates at one pinned block — `PeopleLite.AccountToAlias` and its context against `Score.score_context`, `Game.LiteInvites`, `PeopleLite.LitePeople`, `Members.Root`, `Game.StmtAccountToAlias`, opt-in `Members.Members`. `signUpWithLiteInviteTx` shares the identifier-key and airdrop-VRF guards with `signUpWithAccountTx`, whose body moved unchanged into a shared internal; its public API is untouched.

**Worth review.** `LiteSignUpBlocker` is a parallel union (`SignUpBlocker | <9 lite arms>`) rather than a widening, so exhaustive consumers of the account read keep compiling. The read takes the same three suffix overloads as `readScoreContext` plus a `tld`, following #347: previewnet resolves from its constant, paseo needs `tld`, and a client that resolves neither is a compile error. Three arms encode chain rules invisible from here — `AlreadyPlaying` (`sign_up_inner` rejects an invited sign-up whenever a `Game.Players` entry exists, before it reads `registered`, and a finished game leaves one behind, so those players use `signUpWithAccountTx`), `StaleAlias` (`Custom(172)` in `validate`, one `Members.Root` read away), and `AccountIsAStatementAccount`, checked before both and reachable only by clients other than this one.

Verified against the `individuality-community` pallet source and a live previewnet run (spec 1000036, 2026-08-28). 553 package and 1,943 workspace tests, build, typecheck and biome clean. The contract test pins the read's full parameter type per environment, each assertion falsified by flipping it, and the `fromPapi` path alongside it. `SKILL.md` gains a lite sign-up section.


product-sdk/packages/individuality/src/signup-lite.ts
```diff
@@ -0,0 +1,805 @@
...
+        } else {
+            const invited = await query.Game.LiteInvites.getValue(binding.ca.alias, at);
+            if (invited !== undefined && invited !== account) {
+                liteBlockers.push({ tag: "AnotherAccountInvited", invited });
+            }
+        }
+
+        return ok({
+            ...game,
+            canSignUp: game.canSignUp && liteBlockers.length === 0,
```
Imod7: This evaluates to `true` for an account that already has a `Game.Players` entry, which the chain rejects with `UseInviteButAlreadyPlaying`. `AlreadyRegistered` needs `registered === true`, and a finished game leaves the entry behind with `registered: false`, so this fires for every lite player from their second game on.

Imod7 (approved):

PR #360, opened 2026-09-03
chore(descriptors): re-pin every drifting chain

Imod7:
Refs #242

### Description
Nine of the eleven bundled descriptors had drifted from their live runtimes and a tenth was never measurable. Five had drifted at genesis level, so the descriptor addressed a chain that no longer exists and `getChainAPI` raised `GenesisMismatchError` before any storage read. All eleven now match live.

### Changes
Three commits split by failure mode: the five re-genesised chains, the four whose codeHash alone was stale, and `polkadot-asset-hub`, which the drift probe calls unreachable but is reachable and had drifted.

Both individuality chains moved to spec 3000000, where `Score.Suffix` is gone and `NetworkSuffix` storage replaces it, so six contract assertions were renamed, and previewnet's TLD change from `test` to `testnet` moved its score context. `paseo-asset-hub` dropped `AsRingAlias`. `cloud-storage` hand-pinned previewnet Bulletin's genesis and was addressing a chain that is gone.

The drift report now names which kind of drift a chain has and prints the probe's exit code, so `unreachable` can be told apart from a timeout.

### Testing
```
LC_ALL=en_US.UTF-8 pnpm -r test
pnpm test:e2e
```

All pass, 1,961 tests.

Before, `previewnet-individuality` pinned `0x34999c29...` against a live `0xf720c28f...`, so a product on previewnet could not connect at all. After, every chain matches, and the six flipped assertions were each falsified by inverting them.


product-sdk/examples/cloud-storage-demo/e2e/fixtures.ts
```diff
@@ -4,9 +4,11 @@ import { test as base } from "@playwright/test";
 import {
     createTestHostFixture,
     PASEO_ASSET_HUB,
-    type ChainConfig,
+    type HexString,
+    type NetworkConfig,
     type TestHost,
```
eugypalu, line 8: Looks like `host-demo, keys-demo, signer-demo, statement-store-demo and storage-demo's e2e/fixtures.ts` still use the old chain: `PASEO_AH / ChainConfig` shape, which test-sdk silently ignores. would it make sense to migrate these too?

Imod7: Good catch! Fixed in [`7897499`](https://github.com/paritytech/product-sdk/commit/7897499b45f600cffe4744b548e7f149d388e010)

eugypalu (approved):

PR #366, opened 2026-09-07
fix(terminal): sign as the RFC-0022 product account

### LINKED ISSUE #168: product-sdk-terminal: sign via `createTransaction` with the product account
## Summary

`@parity/product-sdk-terminal` builds its `PolkadotSigner` on top of `polkadot-api/pjs-signer` and forwards PAPI transaction signing to `host-papp` / mobile as a Polkadot-JS-style `signPayload` request. That PJS bridge has two problems on the live chain:

1. **It can't represent every signed extension.** On Paseo Next v2, `Revive.map_account()` carries the `AsPgas` signed extension, which `@polkadot-api/pjs-signer` doesn't know how to map into a PJS payload — so signing fails locally in the CLI *before* the request reaches mobile. This is the root cause tracked in **#80**.
2. **The mobile app wraps `signRaw` variants in the `<Bytes>…</Bytes>` anti-phishing envelope.** That's correct for raw-message signing but corrupts a transaction payload routed through `signRaw`.

Separately, the signer derives its public key from `session.remoteAccount.accountId` — the wallet's *selected/root* account — rather than the app-scoped **product account**. PAPI stamps that key into the extrinsic and verifies against it, so signing with the wrong key fails whenever the product account isn't the selected account.

The fix is to stop using the PJS bridge for transactions and instead route signing through the host's **`createTransaction`** path, signing as the **product account**.

## Original / related issues

- **#80** — terminal cannot sign Paseo v2 `Revive.map_account` because `AsPgas` is not representable through the PJS signing bridge. *(root issue)*
- **#156** — `signRaw` fails with legacy accounts (signer sent as hex pubkey instead of SS58). *(related — the other PJS-path bug, fixed upstream)*

## The fix (two parts)

**Fix 1 — route `signTx` through `createTransaction`.**
`PolkadotSigner.signTx` calls `session.createTransaction` (the host-papp `CreateTransactionRequest`/`Response` SSO pair) instead of `signRaw({ tag: "Payload" })`. The paired wallet builds **and** signs the extrinsic from the structured `ProductAccountTransaction`, so:
- the wallet can decode/display the transaction instead of blind-signing opaque bytes, and
- every signed extension the chain declares — including `AsPgas` — is forwarded verbatim and survives end-to-end (fixes #80), and
- the `<Bytes>` anti-phishing envelope no longer applies to transaction payloads.

`signBytes` keeps the `signRaw({ tag: "Bytes" })` path for the anti-phishing envelope (correct there).

**Fix 2 — sign with the product account, not the wallet's selected account.**
The signer's public key must be the product account's sr25519 key for `[productId, derivationIndex]`. Two ways, both supported:
- **(a) self-derive** via `@novasamatech/statement-store`'s `deriveProductAccountPublicKey(rootAccountId, productId, derivationIndex)` — soft public derivation over `/product/<productId>/<derivationIndex>` (matches the wallet's own derivation), or
- **(b) accept a pre-derived** product-account public key passed in by the caller.

## PRs

| PR | Repo | Role |
|----|------|------|
| [product-sdk #167](https://github.com/paritytech/product-sdk/pull/167) | product-sdk | **The fix** — terminal signs via `createTransaction` + accepts the product-account key (Fix 1 + Fix 2b). Draft, stacked on #163. |
| [triangle-js-sdks #197](https://github.com/paritytech/triangle-js-sdks/pull/197) | triangle-js-sdks | Upstream helper `deriveProductAccountPublicKey` in `@novasamatech/statement-store` (enables Fix 2a — self-derivation). |
| [product-sdk #163](https://github.com/paritytech/product-sdk/pull/163) | product-sdk | Prereq — catalog bump to `@novasamatech 0.8.x` (base of #167). Approved. |
| [triangle-js-sdks #193](https://github.com/paritytech/triangle-js-sdks/pull/193) | triangle-js-sdks | Related — legacy-signer SS58 fix (the other PJS-path bug, #156). |

## Dependency on an upstream release

Fix 2a needs `deriveProductAccountPublicKey` published in `@novasamatech/statement-store` (#197). Fix 1 and Fix 2b need only `@novasamatech/host-papp 0.8.x`, which is already published. The `@novasamatech` packages are released upstream (triangle-js-sdks); **a release cut is needed there** to consume #197 (and #193). If the upstream release is blocked, the fallback is to move/republish the relevant packages under `@parity`.

## Sequence

1. Merge **#163** (catalog 0.8.x) → retarget **#167** to `main` → review + merge #167. This delivers **Fix 1 + Fix 2b** and lets a consumer drop its hand-rolled PJS signer — **no upstream release required for this step**.
2. Cut the upstream `@novasamatech` release (#193 + #197).
3. Follow-up product-sdk PR: switch Fix 2 to **self-derivation** via `deriveProductAccountPublicKey` (#197) and remove the selected-account fallback.

Imod7:
Closes #168

### Description

Product accounts were derived with three soft junctions from the session's root public key, which predates RFC-0022. RFC-0022 made `//product//{productId}` two hard junctions, and a public key cannot cross those, so the derived address stopped matching the one the hosts use. The subtree key now comes from the paired wallet via `getProductSubtree`, and the account is soft-derived from it.

Product-account addresses change as a result, so anything funded at a previously derived address, or any allowance granted to one, is worth checking before upgrading. Signing failed closed in the meantime: the wallet resolves an account selector rather than a key, so it signed as the correct account and transactions were rejected rather than misapplied.

### Changes

`deriveProductAccountPublicKey(productSubtreePublicKey, derivationIndex)` replaces the three-argument form, and `createChainCode` is gone. `createSessionSigner`, `createSessionSignerForAccount` and `deriveProductPublicKey` return promises, since the first derivation per product costs one consent-free round trip to the phone, cached afterwards. The RFC-0022 index expansion moves to `@parity/product-sdk-utils`, shared with `packages/individuality`.

### Testing

The old fixtures were generated from the implementation they were meant to check, so they froze the wrong answer. The test now derives the subtree from a seed and matches host-rust-core's own vector in `truapi-server/tests/wasm_crypto_vectors.rs` (`0x1c1ae478...21cd5b5c`).

```bash
cd product-sdk && pnpm install --frozen-lockfile
LC_ALL=en_US.UTF-8 pnpm -r --no-bail test
```

1985 tests pass across 19 packages, terminal 196 including both interop suites. Not covered: signing against a physically paired phone.


ReinhardHatko:
We built this branch and ran it end-to-end from a terminal consumer
(playground-cli) against a real Android handset, since the PR notes "Not
covered: signing against a physically paired phone". Pairing is confirmed
working; the RFC-0022 derivation is blocked on the phone.

## Blocker: `getProductSubtree` is never answered

```
Login failed: Could not fetch the product subtree key: getProductSubtree timed out — queue freed
```

Attempted twice — once inside login on a cold cache, once standalone against the
already-established session. Identical ~180 s timeout, no subtree cache ever
written, so this is not a login-time race.

A control on the same session, minutes apart, isolates it to this one message:

| Message | RemoteMessage index | Result |
|---|---|---|
| Pairing handshake (V2) | — | ✓ succeeds, `SsoSessionsV4` persisted |
| `ResourceAllocationRequest` | 5 | ✓ responds in **19.8 s** |
| `ProductSubtreeRequest` | 16 | ✗ times out |

So the channel and the statement-store allowance are healthy, and the build has
the v0.9 crypto — it simply does not answer the new message. Not an index
misalignment either: `host-papp@0.10.0` and Android's `SsoMessageContent.kt`
agree on 16/17, with every preceding variant aligned.

Handset: Android build 1026 (2026-09-14), which is current. The full handler
chain (codec → mapper → dispatch → handler) is present in that tag, so this is a
live bug rather than unshipped work. Filed as
paritytech/polkadot-android-community#123.

Credit where due: it fails closed with an actionable message rather than hanging
forever or deriving a wrong address — the "deliberately no fallback" stance
working as designed.

## The obvious workaround does not work either

Shipping #364 alone is tempting, since pairing is verified and #366 is now
blocked on a phone fix. We tested it: built `main` (#364, no #366) against an
unmigrated consumer, same handset.

```
✓ logged in      5FxYmV9z…
✓ account in use playground.dot/0 — 0x669ab828… · 5CwPWFHf…
✓ allowances     granted (2)
✕ mapping        not ready yet
```

Pairing, login and allowance grants all succeed — and the mapping check can
never pass. `requestResourceAllocation` has the phone mint PGAS to its RFC-0022
address while the consumer checks the address the old derivation produces. The
failure is also indistinguishable from a transient one, so our UI reports "this
usually resolves on its own… run login again", looping forever.

#364 alone therefore converts "cannot pair" into "pairs, then silently has an
unusable account", which is worse because it is not obviously broken. **Holding
the two together looks correct** — this is evidence for that call, not against
it.

## Why the timing matters

Every terminal consumer is currently unable to pair at all. bulletin-deploy
looks healthy only because its normal paths (`--mnemonic` / `--suri` / pool /
injected signer) never pair; `bulletin-deploy login` fails identically.
playground-cli has no such cover — `playground login` *is* the pairing flow, and
the statement-store allowance lapses after ~2–3 days, so users are forced back
through it.

## One changeset request

Consumers that derived product addresses locally now need the phone reachable on
a cold cache. For us `deriveSessionAddresses` feeds the login identity line and
`playground status`, both previously offline-capable. Not an objection — it is
inherent to hard junctions — but it deserves a line in the changeset, alongside
the minimum mobile build number once #123 is resolved.

Happy to re-run any of this; the branch is built and wired locally, and the
handset is available.

ReinhardHatko:
One addendum: the `host-papp@0.10.0` bump in #364 also changed `AllocatableResource.SmartContractAllowance` from `value: number` to the tagged `{ tag: "Index", value }`. That is independent of this PR (it reproduces against `main`), but it is undocumented and fails silently under bundlers that strip types without checking — filed as #379 so it does not get lost in this thread.

ReinhardHatko:
## Signing verified on a physically paired phone — the PR's uncovered gap is closed

Following up on [my earlier comment](https://github.com/paritytech/product-sdk/pull/366#issuecomment-5676876721): we got past the `getProductSubtree` blocker and ran the rest of this PR end to end from a terminal consumer (playground-cli) against Android build 1026. **The derivation and the signing path both check out.**

### Derivation matches what the wallet uses

```
✓ logged in      5FxYmV9z8u1htxCoPUUbqT2dbA3Lv7XPb2f7aLNfrrfQE7GA
✓ account in use <product>/0 — 0x8755a050… · 5C7MNyCVdrRBoHepqf4ztQeKMVd5BCdbDqk5J2KmK5BsUQJ5
✓ allowances     granted (2)
✓ mapping        ready
```

`mapping ready` is the meaningful line: our `checkMapping` reads `Revive.OriginalAccount` for the address **the host derived**, and it passes — so the wallet minted PGAS to exactly the account `getProductSubtree` + the RFC-0022 soft-derive produced. A subsequent drip landed 1 PAS at the same address, confirmed by direct chain read.

For contrast, the same consumer on `main` (#364 without #366) can never satisfy that check — the old derivation yields a different address from the one the phone funds, and our UI reports it as transient, so users retry forever. That asymmetry is a good sanity check that the new derivation is the correct one, not merely a different one.

### The phone signs as that account

The above only proves the address is right and fundable — the account's nonce was still 0, i.e. it had never originated a transaction. So we submitted the smallest self-originated extrinsic (`Balances.transfer_keep_alive`, 0 PAS to itself) through the session signer:

```
nonce before: 0
submitting self-transfer of 0  >>> APPROVE ON PHONE <<<
block: 0x195dc900ba261860…
dispatch ok: true
nonce after:  1
```

One approval prompt, signed, included, dispatched OK, **nonce advanced on-chain**. That is the PR's "Not covered: signing against a physically paired phone" line, now covered.

### Important caveat — this does not mean the PR is shippable yet

We could only get there by changing the consumer's product id to match the phone's TLD. With `playground.dot` against a `paseo`-TLD handset, `getProductSubtree` still times out — paritytech/polkadot-android-community#123, now root-caused by the Android team as a product-id TLD re-validation in the SCALE mapper that throws and is swallowed.

So: **#366's derivation and signing are sound; the blocker is entirely external.** Any consumer whose product id carries a different TLD from the paired phone still cannot complete a login, and for us that is the default (`playground.dot`, TLD-independent by convention).

Two things that would help consumers when this ships:

- a minimum mobile build number in the changeset, once #123 lands;
- a note that the first derivation per product needs the phone reachable, since it moves a previously offline path (our login identity line and `playground status`) onto a round trip. Measured 3 s–30 s when it succeeds, so the timeout guidance matters.

Happy to re-run any of this against a fixed Android build — the branch stays wired locally.

TarikGul (approved):

PR #378, opened 2026-09-15
feat(host): expose the unwatermarked raw-signing calls

tommyldev:
Stacked on #376 (the `@parity/truapi` 0.16.0 bump); retarget to `main` once that merges.

## Summary

`AccountsProvider` gains two deprecated calls that wrap the host methods added in [host-rust-core#731](https://github.com/paritytech/host-rust-core/pull/731) and shipped in `@parity/truapi` 0.16.0:

```ts
signRawUnwatermarkedDeprecated(account: ProductAccount, data: Uint8Array): Promise<Uint8Array>
signRawUnwatermarkedDeprecatedWithLegacyAccount(account: { publicKey: Uint8Array }, data: Uint8Array): Promise<Uint8Array>
```

Both sign `data` with **no** `<Bytes>` watermark and return the raw signature bytes, mirroring the `signBytes` of the two `PolkadotSigner` factories. They exist because People chain's `Resources.register_person` verifies a bare-byte `lite_identity_proof` ([host-rust-core#612](https://github.com/paritytech/host-rust-core/issues/612)), which the ordinary watermarked `signBytes` can never satisfy. Hosts log a deprecation warning and show a stronger confirmation for these calls; both go away once the runtime accepts watermarked proofs.

The four raw-signing paths now share one `toBytesPayload` helper and one `AccountId()` codec instance. Skill docs get a short paragraph on when (not) to use the new calls.

**Breaking for implementors:** both are required members of the exported `AccountsProvider` interface, so alternative implementations and hand-rolled test doubles must add them. Callers are unaffected; `createFakeTruApiClient` already models both (#376).

Consumer: paritytech/humanity-spa#54 (lite → full username claim).

## Validation

- `pnpm install --frozen-lockfile`, `pnpm check`, `pnpm build`, `pnpm typecheck`, `pnpm test` — all green locally (host: 140 tests, incl. two new ones asserting the wire request shape and decoded signature for each call).
- Smoke-tested from humanity-spa against a `truapi-host` 0.16.0 built from the release tag on paseo-next-v2: codec-2 handshake, 48 host calls, 0 failures.

## Changeset

`pending-changesets/host-unwatermarked-signing.md` — `@parity/product-sdk-host: minor`, `@parity/product-sdk: minor`.


TarikGul:
Approving. Checked the two new methods against truapi 0.16.0 and they reuse the exact same request/response types as signRaw and signRawWithLegacyAccount, so the only difference is the method name and the host skipping the watermark. That is what makes the shared toBytesPayload helper safe across all four paths. Refactor is a real cleanup, tests pin the exact wire request and guard against the Bytes wrap sneaking back in, and the changeset plus SKILL docs are spot on (host and umbrella minor is correct since the umbrella re-exports host). Build, typecheck, 140 tests, and biome all pass locally. Good to merge.

tommyldev:
@TarikGul Still requires an approval

decrypto21 (approved):

PR #380, opened 2026-09-15
fix(docs): tag the derivation index, pin AllocatableResource

### LINKED ISSUE #379: #364 changed AllocatableResource.SmartContractAllowance shape without a changeset entry — silent break for consumers that strip types
## Summary

PR #364 bumped `@novasamatech/host-papp` to `0.10.0`, which changed the
`SmartContractAllowance` derivation index from a plain `number` to the tagged
RFC-0022 selector:

```ts
// before
{ tag: "SmartContractAllowance", value: 0 }

// after
{ tag: "SmartContractAllowance", value: { tag: "Index", value: 0 } }
```

`AllocatableResource` is public API — consumers construct these values to pass
to `requestResourceAllocation`. The change is not mentioned in #364's changeset,
which describes the bump as a pairing fix.

## Why it is worth its own issue

It fails **silently** under the bundlers terminal consumers actually use. `bun
build` and `esbuild` strip types without checking them, so an affected consumer
compiles clean and ships a wrong-shaped `AllocatableResource`; only a separate
`tsc --noEmit` catches it. Our own build pipeline is exactly this shape, and the
error surfaced only because we typecheck separately:

```
src/utils/allowances/resources.ts(91,38): error TS2322: Type 'number' is not
  assignable to type '{ tag: "Index"; value: number } | { tag: "Raw"; value: Uint8Array } | undefined'
src/utils/allowances/smartContracts.ts(26,5): error TS2322: (same)
```

## Verified as #364, not #366

Typechecking an unmigrated consumer against `main` (which has #364 and not #366)
reproduces both errors. So this arrives with the pairing bump and is independent
of the RFC-0022 derivation work — a consumer taking #364 alone still needs a
source change.

## Suggested fix

A changeset entry (and a line in any migration note) calling out the shape
change, with the before/after above. The code change on the consumer side is
mechanical once known; the cost is entirely in not knowing.

Worth checking whether any other `AllocatableResource` variant changed shape in
the same bump — we only exercised `SmartContractAllowance` and
`BulletInAllowance`.

Imod7:
Closes #379

### Description

The migration page passed a bare number to `truApi.resourceAllocation.request`, which has never matched any pinned `@parity/truapi`. That payload is a tagged `DerivationIndex`, so the samples now pass `{ tag: "Index", value: n }`. The six allocation and permission types the two packages export but do not define are also pinned to literal shapes inside their existing `import.meta.vitest` blocks, so a dependency bump that reshapes public API fails typecheck by name instead of shipping quietly, as it nearly did in #364. The pins are type-only. A stale host-papp version in the `host.ts` module doc and the old bare-number form in the auth skill reference are fixed in passing.

### Testing

The docs sample, compiled against the installed `@parity/truapi` 0.16.0:

```
before: error TS2322: Type 'number' is not assignable to type 'DerivationIndex | undefined'
after:  compiles clean
```

All six pins were proven red before green, each against a shape it exists to catch: the host-papp 0.8.9 bare `number` payload, the pre-#364 `productDerivationSecret` field, a dropped allowance-policy variant, a dropped outcome variant, a flipped optional payload, and a narrowed nested field. Each raises `TS2344` on its own assertion line.

```bash
cd product-sdk && pnpm build && pnpm typecheck && pnpm -r test && pnpm check
cd ../docs && pnpm install --frozen-lockfile && pnpm docs:generate && pnpm build
```

All pass, 1966 tests.


TarikGul:
Approving. This is the right fix for the #379 silent break. Verified the pins against the installed truapi 0.16.0 and host-papp 0.10.0 and every one matches its source exactly, including the two easy-to-miss bits: truapi uses optional value and host-papp uses required, and the BulletinAllowance vs BulletInAllowance casing is a real difference between the two libraries, not a typo. The SmartContractAllowance pin locks the Index or Raw shape that changed silently before, so a future reshape now fails typecheck naming the type instead of blowing up at runtime. Docs fix is consistent across the table, the sample, and the new note. Terminal patch changeset is the right scope since the only shipped change is the stale host-papp version in the doc comment. Build, typecheck, host and terminal tests, and biome all pass locally. Good to merge.

TarikGul (approved):

